PR #28, opened 2017-09-22
Cleaned up the 3DBall example scene

ASPePeX:
Sorted the "Games" in the hierarchy and named them according to their position in the scene.
Also added a Game prefab.


awjuliani:
Looks good to me! Thanks for making these changes, @ASPePeX.

PR #40, opened 2017-09-27
Added worker-id flag and pass through to environment.

ashwal:
Makes managing multiple tests/async agents much easier than editing the environment.py file directly.


awjuliani:
Looks good! Thanks for making this change.

PR #63, opened 2017-10-06
Fixed `Academy` children `Brain` gathering.

JoshuaLight:
Hi! Really like your project and wanting to help.

So I've found, that `Academy` is using `GetComponentsInChildren` method to gather all child `Brain` components. And, as far as I know (empirically too), this method has a tricky name: actually, it also takes `Brain` component from calling object.

Also `GetComponentsInChildren` looks recursively in all children of object, not the nearest only.

And there also is error (oh, I see the typo here):
image

Which, as I can see, means, that `Academy` will search for brain only it's children objects (not recursively), so for now we can get such error panel even in situations, when `Academy` will actually find some `Brains`.

For example, such configuration will result in one `Brain` in `Academy`:
image

I've fixed this by manually looping over children of `Academy`.

_P.S. It looks like I've accidentally removed some empty lines. If it's not ok, I'll revert. ^^_

### DIFF
--- a/unity-environment/Assets/ML-Agents/Scripts/Academy.cs
+++ b/unity-environment/Assets/ML-Agents/Scripts/Academy.cs
@@ -67,7 +67,7 @@ public abstract class Academy : MonoBehaviour
 
 
     [HideInInspector]
-    private Brain[] brains = new Brain[0];
+    private List<Brain> brains = new List<Brain>();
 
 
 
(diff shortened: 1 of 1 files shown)


awjuliani:
Hi @JoshuaLight,

Thanks for catching this. The fix makes sense, and I will merge it.

PR #64, opened 2017-10-06
Fixed typo in `BrainEditor`.

JoshuaLight:
Replaced `myst` with `must`.


awjuliani:
Thanks for catching this!

PR #88, opened 2017-10-15
BrainEditor: Replaced `new SerializedObject()` with property.

JoshuaLight:
Hi!

Just small adjustment for `BrainEditor`. There is a new `SerializedObject` every `OnInspectorGUI` right now, and this can be optimized away with `serializedObject` property, that do some caching.

image


JoshuaLight:
I've also found right now, that `if` statement, that checks `brainParameters` can be simplified a bit, so I've updated `PR`. Also I've removed some strange empty lines, but if they make sense, I'll revert.

awjuliani:
Thanks for making these changes, @JoshuaLight!

PR #91, opened 2017-10-17
 Add progress report and fixed step bug.

SPINACHCEO:
1. Fixed: "step" in if statement should be "steps".
2. Add time report including percent and etr to max step.


awjuliani:
Hi @SPINACHCEO,

Thanks for adding this. I think people will appreciate it. One thing I would ask that you change is that progress reporting should be together with summary reporting, so as to avoid having too many config flags. 

I also haven't looked closely enough, but if the model training is interrupted and loaded again from halfway through training, will the time remaining estimate still be correct?

SPINACHCEO:
@awjuliani 
Hi, I just made a new pull request with a better progress bar that don't mess up the screen.
I like it very much :)

SPINACHCEO:
Please refer to the my new pull request.

PR #93, opened 2017-10-18
Add Runner sample environment

mplantady:
A classic runner game where the agent needs to jump through a series of obstacles and collect bonuses.

**Modifications on the existing codebase :**
- Switch maxStep as protected, to be able to use it in the game specific academy.
- Add collision type in Unity TagManager

**What I learn from it :**
- It is hard to balance rewards between an optional goal (bonus) and a primary goal (surviving)
- Using curriculum-level design is quickly essential when we add more and more type of obstacles
- Hyperparameters needs to be tweak to follow the desired learning curve 

**What could be done later :**
- Create an endless mode (when maxStep = 0) using an object pool
- Create more complex sequences with different heights


dora-gt:
Though it's a small problem, indent is broken, so CTL+I (format code) in Mono Develop will help you.

BTW, how long did it take to learn?
I just want to know.

mplantady:
Thanks I re-indent them, should be better.

It took approximately 3 hours. Actually 3 consecutive trainings of 1 hour, to let me reorganize the level sequences depending on agent weakness.

I've reduced hidden-units at 32, because I felt that 64 was too much for so few input data. But it doesn't make a real difference...

dora-gt:
Only 3 hours... omg... your design seems to be so clever.
My environment never works, maybe something is wrong.
Anyway, thank you.

awjuliani:
Hey @mplantady!

Apologies for taking so long to finally get back to this. There was a lot of internal back and forth about the best way to share these kinds of environments created by the community. While we initially thought we'd like to fold them into the repo itself, we ended up deciding that it would be best to host them elsewhere. As such, we are going to be creating a dedicated section of the Asset Store for people to share environments in the near future. We've also created a page to highlight created environments on the new Unity Connect page (https://connect.unity.com/challenges/ml-agents-1). If you are up for it, it would be great to see the infinite runner demo you made up there, possibly with a link to your project fork containing the game.

PR #99, opened 2017-10-19
Added Blocks example

dracolytch:
Another example environment. It trains and works! All new assets for this environment are isolated in a subdirectory under samples.


dracolytch:
Added a robot arm example where the robot arm has two degrees of freedom.

awjuliani:
Hey @dracolytch!

Apologies for taking so long to finally get back to this. There was a lot of internal back and forth about the best way to share these kinds of environments created by the community. While we initially thought we'd like to fold them into the repo itself, we ended up deciding that it would be best to host them elsewhere. As such, we are going to be creating a dedicated section of the Asset Store for people to share environments in the near future. We've also created a page to highlight created environments on the new Unity Connect page (https://connect.unity.com/challenges/ml-agents-1). If you are up for it, it would be great to see the demo projects you made up there, possibly with a link to your project fork containing the game.

MarcoMeter:
I think the Asset Store is a suitable platform for sharing environments. It would be great to have a guideline for sharing these kind of projects. Like informing the developers to upload the environment plus the ml-agents files. This would be required for people to not run into issues due to outdated versions of the ml-agents. However, an intuitive way would be necessary to provide the required python files, which are not inside of the Unity project.

I had a look at most of the projects posted to Unity Connect. It would be awesome if people could follow a little structure to make it easier for others to identify inputs, outputs, environmental constraints and rewards. Sometimes some of that relevant information is left out or hard to find. But maybe I'm too much of a scientist and have to high expectations. I guess there are lots of people doing that for fun (and that's great!).

PR #160, opened 2017-12-01
Corrected state space of tennis environment

atamahjoubfar:
State space of the tennis brain is corrected to 8 variables according to the `CollectState()` method of the `TennisAgent.cs` script.


awjuliani:
Thanks for catching this!

PR #179, opened 2017-12-15
Fixed to give lesson index parameter when start up

dora-gt:
Fixed to give lesson index parameter when start up.
With this option, user can use `--load` and `--curriculum=<file>` at the same time, and start lesson from one that user exactly wants.


awjuliani:
Hi @dora-gt,

This looks great! Would it be possible to make the same changes to our `PPO.ipynb` notebook as well for those who prefer to use that for training? Once that change is made I am happy to merge it!

dora-gt:
@awjuliani Hi,

I applied the same modification to `PPO.ipynb`, and confirmed it worked!

awjuliani:
Perfect. Will merge into development 👍

PR #180, opened 2017-12-16
Improve memory management

awjuliani:
* More efficient memory usage within core ML-Agents scripts, with special focus on `ExternalCommunication.cs`. 

* Reduces GC memory each frame from ~25 to ~5 when using external brain on `Push Area` environment, and `Time.Scale` of `25`.

* Introduces small breaking change where `state` var in `CollectState()` no longer needs to be created every time. This also means attempts to create it throw an error, as it is now created when the Agent is initialized. The fix is to delete the line, but this will need to be done by developers who have implemented their own environments.

### DIFF
--- a/unity-environment/Assets/ML-Agents/Scripts/Brain.cs
+++ b/unity-environment/Assets/ML-Agents/Scripts/Brain.cs
@@ -36,6 +36,9 @@ public enum StateType
     continuous}
 ;
 
+
+
+
 /** Only need to be modified in the brain's inpector.
  * Defines what is the resolution of the camera
 */
@@ -86,6 +89,15 @@ public class BrainParameters
  */
 public class Brain : MonoBehaviour
 {
+    // Current agent info
+    public Dictionary<int, List<float>> currentStates = new Dictionary<int, List<float>>();
+    public Dictionary<int, List<Camera>> currentCameras = new Dictionary<int, List<Camera>>();
+    public Dictionary<int, float> currentRewards = new Dictionary<int, float>();
+    public Dictionary<int, bool> currentDones = new Dictionary<int, bool>();
+    public Dictionary<int, float[]> currentActions = new Dictionary<int, float[]>();
+    public Dictionary<int, float[]> currentMemories = new Dictionary<int, float[]>();
+
+
     public BrainParameters brainParameters = new BrainParameters();
     /**< \brief Defines brain specific parameters such as the state size*/
     public BrainType brainType;
@@ -188,11 +200,15 @@ public class Brain : MonoBehaviour
 
     }
 
-    /// Collects the states of all the agents which subscribe to this brain 
-    /// and returns a dictionary {id -> state}
-    public Dictionary<int, List<float>> CollectStates()
-    {
-        Dictionary<int, List<float>> result = new Dictionary<int, List<float>>();
+
+    public void CollectEverything() {
+        currentStates.Clear();
+        currentCameras.Clear();
+        currentRewards.Clear();
+        currentDones.Clear();
+        currentActions.Clear();
+        currentMemories.Clear();
+
         foreach (KeyValuePair<int, Agent> idAgent in agents)
         {
             idAgent.Value.SetCumulativeReward();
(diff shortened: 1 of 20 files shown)


unity-environment/Assets/ML-Agents/Scripts/Brain.cs
```diff
@@ -86,6 +89,15 @@ public class BrainParameters
...
 {
+    // Current agent info
+    public Dictionary<int, List<float>> currentStates = new Dictionary<int, List<float>>();
+    public Dictionary<int, List<Camera>> currentCameras = new Dictionary<int, List<Camera>>();
+    public Dictionary<int, float> currentRewards = new Dictionary<int, float>();
+    public Dictionary<int, bool> currentDones = new Dictionary<int, bool>();
+    public Dictionary<int, float[]> currentActions = new Dictionary<int, float[]>();
+    public Dictionary<int, float[]> currentMemories = new Dictionary<int, float[]>();
+
+
```
flamz3d: I'm going to make some comments regarding the coding convention for C# at Unity. You don't need to refactor all your code to the standard, but it's a good idea to progressively adhere to it. Alot of the coding convention is annoying and I find arbitrary, but some of it is useful because it enforces good patterns for APIs

Member fields (like currentStates, etc) should:

- always start with m_ /w camelcase (m_CurrentStates)

- never be public.  This is important because the API is not clear. For example, if I have an instance of a brain, can I call currentStates.Add(...)? Usually, you make your fields private or protected, use [SerializedField] if they need to persist, and make a matching property with relevant access rights.
Example:

```
protected Dictionary<int,int> m_CurrentStates = new ...
public IEnumerable<KeyValuePair<int,int>> states { get { return m_CurrentStates;} }

```

unity-environment/Assets/ML-Agents/Scripts/Brain.cs
```diff
@@ -86,6 +89,15 @@ public class BrainParameters
...
     public BrainParameters brainParameters = new BrainParameters();
     /**< \brief Defines brain specific parameters such as the state size*/
```
flamz3d: in C# comments should come before the code being commented. This allows tools like Visual Studio or Ryder to automatically parse the doc and show it alongside intelisense results, etc.

unity-environment/Assets/ML-Agents/Scripts/ExternalCommunicator.cs
```diff
@@ -25,6 +25,20 @@ public class ExternalCommunicator : Communicator
...
+    List<float> concatenatedMemories = new List<float>(1024);
+    List<bool> concatenatedDones = new List<bool>(32);
+    List<float> concatenatedActions = new List<float>(1024);
+
+    //Dictionary<int, List<Camera>> collectedObservations;
+    //Dictionary<int, List<float>> collectedStates;
+    //Dictionary<int, float> collectedRewards;
+    //Dictionary<int, float[]> collectedMemories;
+    //Dictionary<int, bool> collectedDones;
+    //Dictionary<int, float[]> collectedActions;
```
flamz3d: remove commented code

unity-environment/Assets/ML-Agents/Scripts/ExternalCommunicator.cs
```diff
@@ -240,25 +256,21 @@ public void giveBrainInfo(Brain brain)
     {
         string brainName = brain.gameObject.name;
```
flamz3d: we try to use C#'s type inference as much as we can. For instance:

`string brainName = "blah";`
becomes
`var brainName = "blah";`

it might seem trivial, but it helps *alot* during code refactoring. and makes you write C# code much faster :)
Also, when the returning value is a complicated generic, it's much nicer to do:

`var theStates = brain.states;`
than
`IEnumerable<KeyValuePair<int,int>> brain.state;`

unity-environment/Assets/ML-Agents/Scripts/ExternalCommunicator.cs
```diff
@@ -315,6 +327,9 @@ public void UpdateActions()
                 string brainName = brain.gameObject.name;
 
                 Dictionary<int, float[]> actionDict = new Dictionary<int, float[]>();
+                Dictionary<int, float[]> memoryDict = new Dictionary<int, float[]>();
+                Dictionary<int, float> valueDict = new Dictionary<int, float>();
+
```
flamz3d: use type inference here.
`var actionDict = new Dictionary<int, float[]>();`

PR #218, opened 2017-12-28
Update Ball3DDecision.cs

LionH:
I think it is important to add a simple heuristic example to demonstrate how to implement a Decision


LionH:
Please review the modification. Very simple implementation of how to code a Decision.

awjuliani:
Hi @LionH,

Thanks for making this addition. I agree that having an example of a heuristic function is definitely valuable for the community. Happy to incorporate this change.

PR #220, opened 2017-12-29
Added information about input normalization to the best practices doc

MarcoMeter:
I'm not sure if that's mentioned at all in the docs. I've observed that many people, who posted issues, did not normalize their input space. That's quiet an important step in my opinion.


awjuliani:
Hi @MarcoMeter,

Thanks for submitting this PR! I think that this is likely worth including. As a note, we actually included an auto-normalization function as part of our PPO implementation, however it applies moreso to continuous control tasks rather than discrete control, where we've observed that it actually impedes performance.

Issue #229, opened 2018-01-05
Unity was unable to read the arguments passed at launch.


tananuka:
Please Help.
I am using Unity 2017.3.0f3 Personal and macOS High Sierra 10.13.2
I installed Python 2.7 from package ver. 9.0.1, that downloaded from official [site](https://www.python.org/downloads/mac-osx/) 
Then i performed in terminal: `pip install tensorflow`
that's all what i installed, maybe i missed something?
Because when i am trying to launch Balance Ball Example i get exception:
`UnityAgentsException: The brain Ball3DBrain was set to External mode but Unity was unable to read the arguments passed at launch.
CoreBrainExternal.InitializeCoreBrain () (at Assets/ML-Agents/Scripts/CoreBrainExternal.cs:27)
Brain.InitializeBrain () (at Assets/ML-Agents/Scripts/Brain.cs:187)
Academy.Awake () (at Assets/ML-Agents/Scripts/Academy.cs:124)`

MDReptile:
That error means the brain was set to external mode, and the game wasn't able to connect to the Anaconda/PPO for some reason when it launched. If you want to test an environment you can't have brain set to external, that is only for when the PPO stuff drives the agents, I test with it set to player.

tananuka:
thanks @MDReptile ,
i installed anaconda from official site and setup tensorflow using this [guide](https://www.tensorflow.org/versions/r0.12/get_started/os_setup#anaconda_installation)
but i have new [problem](https://github.com/Unity-Technologies/ml-agents/issues/200))

MDReptile:
You need to edit the PPO when you open it in your browser, where it has "env_name" you put the name of your built exe. Like "MyTestBuild.Exe" would be "MyTestBuild" (without .exe)

jeremyPLR:
You get the `unable to read the arguments passed at launch` error if you try to play the scene inside of the Unity editor with your `brain type` set to `external`.  These are the steps that I follow:

1. Open the scene in the Unity editor and set the `brain type` to `external`.
2. Save and build the scene (e.g. 3dball.app).
3. Train your brain from python on your command line (or jupyter notebook).
4. Follow the instructions to embed the trained model into Unity.
5. Now, back in Unity, set the `brain type` to `internal` and set the `graph model` to the bytes file you exported from python.
6. Play the scene in Unity and you should be able to observe the results of your training.

tananuka:
@MDReptile  @jeremyPLR 
[...]

Issue #200, opened 2017-12-22
UnityEnvironmentException: Couldn't launch the 3DBall environment. Provided filename does not match any environments.


shengtianzhu:
Hi
I tried the 3DBall Tutorial. But when I try to load the environment I get the error:

![image](https://user-images.githubusercontent.com/7135752/34294463-a6fa6756-e743-11e7-9465-06a771918ee2.png)

I use Unity to Build the exe file named"3DBall".And Copy this flie to python folder.But it's show error.
on the Tutorial step:
![image](https://user-images.githubusercontent.com/7135752/34294778-0ad21f98-e745-11e7-818b-f686b42b05ba.png)
I think the "environment" is "3DBall.exe"  flie and  "3DBall_Data" folder. Copy this two flie to python folder.
What is error on my step?

shengtianzhu:
I am using Windows 10. other environment like the Tutorial say

Izhmael:
Hello,
I tried only on MAC. On it i have to save build in unity-environment folder, and next in the Jupyter notebook change env_name = "name_of_the_file" in Basics.ipynb. 

<img width="1130" alt="zrzut ekranu 2017-12-22 11 33 14" src="https://user-images.githubusercontent.com/3924185/34295155-09c36cf8-e70c-11e7-8aff-aa09cbaaae29.png">

shengtianzhu:
Hi @Izhmael  
Thanks! I try on your step. But still error
![image](https://user-images.githubusercontent.com/7135752/34316546-8f6959d6-e7d3-11e7-9ea9-fc81d44c3ea4.png)
I don't think I know how to use the jupyter. I'll learn how to use the jupyter, first

awjuliani:
Hi @shengtianzhu,

It seems that you are running one ipython notebook from another. Is there a reason for this?

Fangh:
you have to put your build name 3DBall.exe in python directory. (mine is called build) :

![image](https://user-images.githubusercontent.com/6020679/34326279-20f2ee1e-e8a9-11e7-836d-29c2e4f12df1.png)

shengtianzhu:
Hi @Fangh 
this is my python directory screenshot
![image](https://user-images.githubusercontent.com/7135752/34330621-e20a9f1a-e959-11e7-9083-d3bd3e178cad.png)
and my Basics.ipynb screenshot is
[...]

PR #236, opened 2018-01-07
Fixed issue with unity environment not being found on MacOS

wouterhardeman:
Fixes the problem for MacOS mentioned in #200 and #229.

When the internal executable can't be found, it will look for any file in the 'Contents/MacOS/' folder and run it. If another solution is preferred (like optionally specifying the correct file name), let me know. But this should work for now. 😄


awjuliani:
Hi @wouterhardeman,

This is great! Thanks for making this fix. I know a couple people (myself included) have accidentally renamed the `.app` before without thinking of the consequences. Glad we won't have to worry about it anymore.

PR #241, opened 2018-01-11
Added documentation for users of Conda Envs.

tall-josh:
- Creating a new Conda Env
- Installing requirements
- Linking Conda Env with Jupyter Kernel


awjuliani:
Hi @tall-josh,

Thanks for writing this documentation. We are actually reworking our docs for the upcoming v0.3 release, and unfortunately this won't fit in there, as we are actually writing a more in-depth guide.

PR #252, opened 2018-01-15
Tiny TensorFlowSharp docs tweak

60days:
Namespace capitalization change for people copy/pasting - otherwise namespace is not found


awjuliani:
Thanks for catching this, @60days!

PR #259, opened 2018-01-18
 Fixed indent error in initialization

ZephyrRaine:
Hello, just a small typo found in curriculum.py when trying to train without curriculum_file.

set_lesson_number was called even if curriculum_file was equal to None, which was causing an error in the function since self.max_lesson_number wasn't declared.


ZephyrRaine:
Turns out it leads to further bugs, because it still tries to access curriculum later. Sorry for messy pull requests haha.

Issue #277, opened 2018-01-23
Sometimes tensorflow fails at creation with windows


floAr:
When running under windows sometimes the creation of the device fails, this is a known behaviour under windows for tensorflow

Log:
```
2018-01-23 08:58:04.434680: I C:\tf_jenkins\home\workspace\rel-win\M\windows-gpu\PY\36\tensorflow\core\common_runtime\gpu\gpu_device.cc:1030] Found device 0 with properties:
name: Quadro K2100M major: 3 minor: 0 memoryClockRate(GHz): 0.6665
pciBusID: 0000:01:00.0
totalMemory: 2.00GiB freeMemory: 1.66GiB
2018-01-23 08:58:04.439581: I C:\tf_jenkins\home\workspace\rel-win\M\windows-gpu\PY\36\tensorflow\core\common_runtime\gpu\gpu_device.cc:1120] Creating TensorFlow device (/device:GPU:0) -> (device: 0, name: Quadro K2100M, pci bus id: 0000:01:00.0, compute capability: 3.0)
2018-01-23 08:58:05.803329: E C:\tf_jenkins\home\workspace\rel-win\M\windows-gpu\PY\36\tensorflow\stream_executor\cuda\cuda_blas.cc:366] failed to create cublas handle: CUBLAS_STATUS_ALLOC_FAILED
2018-01-23 08:58:05.807146: E C:\tf_jenkins\home\workspace\rel-win\M\windows-gpu\PY\36\tensorflow\stream_executor\cuda\cuda_blas.cc:366] failed to create cublas handle: CUBLAS_STATUS_ALLOC_FAILED
2018-01-23 08:58:05.810992: E C:\tf_jenkins\home\workspace\rel-win\M\windows-gpu\PY\36\tensorflow\stream_executor\cuda\cuda_blas.cc:366] failed to create cublas handle: CUBLAS_STATUS_ALLOC_FAILED
2018-01-23 08:58:05.815134: E C:\tf_jenkins\home\workspace\rel-win\M\windows-gpu\PY\36\tensorflow\stream_executor\cuda\cuda_blas.cc:366] failed to create cublas handle: CUBLAS_STATUS_ALLOC_FAILED
[...]

PR #278, opened 2018-01-23
Added growth parameter to stop failing with allocation under windows …

floAr:
…for #277

https://stackoverflow.com/questions/41117740/tensorflow-crashes-with-cublas-status-alloc-failed


awjuliani:
Thanks @floAr for making this change!

PR #289, opened 2018-01-26
fix export_graph bug when env_name contains a sub folder

zhenghongzhi:
Fix export_graph bug when env_name contains a sub folder
For example:
When env_name is "build/game.exe", the export_graph won't work properly and this pull request fixes this.


zhenghongzhi:
sorry for the wrong branch

PR #324, opened 2018-02-09
Fix typo "leaning_rate"

Coac:


awjuliani:
Thanks for catching this typo, @Coac!

PR #361, opened 2018-02-23
Dev docs

jo3w4rd:
Merging doc work into the development branch.


mmattar:
@awjuliani - Can you review and approve? Since @jo3w4rd and I worked on this, a (brief) fresh review is due. Thanks.

awjuliani:
Overall it looks great. A few comments based on a light read-through:
* There are still some sections unfinished, such as the TensorFlow and Deep Learning articles. Is the plan to add those later?
* In "Getting Started" there is a note to insert the Scene Hierarchy, but the image isn't there.
* The TensorBoard image on "Getting Started" doesn't correspond to 3DBall. This could confuse new users expecting their screens to look like the image.
* The rollerball example includes code suggesting that the user use a reward whenever the ball gets closer to the goal. This kind of reward shaping is something we explicitly don't want to encourage, as it can lead to undesirable behavior and slower learning in many cases.

mmattar:
Thanks @awjuliani. Regarding point 1 - yes, the plan is to expand those, but we want to integrate back into dev-0.3 earlier and start working off that. Regarding points, 2-4, I'll defer to @jo3w4rd.

jo3w4rd:
2. Will fix.
3. Good point, I'll replace it with one from 3D Ball.
4. I tried that first, the Agent didn't learn very well at all. It worked fine if I didn't use physics (AddForce) to move the agent, but otherwise, the agent wouldn't be able to find the target unless the target happened to be closer to the front of the platform (even after an extended period of training). Open to ideas to improve things.

awjuliani:
Hi @jo3w4rd,

I just checked this out with Yuan. If you remove the reward shaping, and also decrease the "existential penalty" by 10x, then it looks to learn fine. I think it would be better to encourage simpler, sparser rewards like these, even if it means slightly longer training times, or the need for larger batches/buffers.

PR #376, opened 2018-03-01
Add config for crawler, and change crawler scene

awjuliani:
* Added Crawler hyperparameters to `trainer_config.yaml`
* Changed number of crawlers in scene to 12
* Changed Max-steps for crawlers to 5000


awjuliani:
Crawler now achieves higher reward, and new `.bytes` model doesn't use epsilon.

PR #379, opened 2018-03-02
Fix for when not using teacher helper

awjuliani:
* We assumed that the teacher helper would also be used when doing behavioral cloning. This assumption can be violated, and needs to be handled.


eshvk:
When does it make sense to allow the violation of this assumption as expected behavior?

awjuliani:
It is an optional script which can be attached to agents, but not necessary for imitation learning to work. As such, we shouldn't expect users to attach it. The value of it means they most likely will though. We don't want to build this functionality into the agent class itself, since the agent should be agnostic to how it is trained.

PR #417, opened 2018-03-09
[Important] Fixes the logic of the academy reset

vincentpierre:
The External Communicator now sends the AcademyDone properly


vincentpierre:
Tested on PPO, BC and Curriculum

PR #437, opened 2018-03-12
Update API version in anticipation of v0.3 release

awjuliani:


python/unityagents/environment.py
```diff
@@ -37,7 +37,7 @@ def __init__(self, file_name, worker_id=0,
         atexit.register(self.close)
         self.port = base_port + worker_id
         self._buffer_size = 12000
-        self._python_api = "API-2"
+        self._python_api = "API-3"
```
eshvk: Can we call the variable name `_version_`  in both C# and Python, i.e. 

`self._version_ = "API-3"` or even `self._version_ = "0.3"` and `const string _version_ = "API-3"`

awjuliani: Done.

python/unityagents/environment.py
```diff
@@ -115,14 +115,14 @@ def __init__(self, file_name, worker_id=0,
                     .format(str(file_name)))
 
             if "apiNumber" not in p:
-                self._unity_api = "API-1"
+                self._unity_version = "API-1"
```
mmattar: Not a fan of this, might be easily confused with Unity Version, as in 2018.1, etc.

awjuliani: I agree this could be problematic. Any suggestion for a better name?

mmattar: mlagents_python_version and mlagents_unity_version (?)

mmattar: Actually, why do we need two local vars .... we just need to create a const and check that it is different from p["apiNumber"] (?)

unity-environment/Assets/ML-Agents/Scripts/ExternalCommunicator.cs
```diff
@@ -36,7 +36,7 @@ public class ExternalCommunicator : Communicator
     StreamWriter logWriter;
     string logPath;
 
-    const string api = "API-2";
+    const string _version_ = "API-3";
```
mmattar: why the "_"?

python/unityagents/environment.py
awjuliani: We didn't have API version in v0.1, so we need to be able to catch Unity not providing one.

unity-environment/Assets/ML-Agents/Scripts/ExternalCommunicator.cs
awjuliani: @eshvk recommended this.

PR #460, opened 2018-03-14
updated the comments

xiaomaogy:


python/Basics.ipynb
```diff
@@ -109,13 +117,17 @@
    "metadata": {},
    "source": [
     "### 5. Take random actions in the environment\n",
-    "Once we restart an environment, we can step the environment forward and provide actions to all of the agents within the environment. Here we simply choose random actions based on the `action_space_type` of the default brain."
+    "Once we restart an environment, we can step the environment forward and provide actions to all of the agents within the environment. Here we simply choose random actions based on the `action_space_type` of the default brain. \n",
+    "\n",
+    "You are going to see 10 messages that tells you how much rewards you are going to see in the next 10 episodes, then the Unity environment is going to be stuck, waiting for further signals telling it what to do next. So not seeing any moving animation is expected when running this notebook. "
```
mmattar: "You are going to see 10 messages that tells you how much rewards you are going to see in the next 10 episodes, then the Unity environment is going to be stuck, waiting for further signals telling it what to do next. So not seeing any moving animation is expected when running this notebook. " --> "Once this cell is executed, 10 messages will be printed that detail how much reward will be accumulated for the next 10 steps. The Unity environment will then pause, waiting for further signals telling it what to do next. Thus, not seeing any animation is expected when running this cell."

PR #501, opened 2018-03-16
Corrected Wall Jump description

awjuliani:
Wall jump description now correctly states that there are two brains used in this environment.


docs/Learning-Environment-Examples.md
```diff
@@ -109,7 +109,7 @@ If you would like to contribute environments, please see our
     * -0.0005 for every step.
     * +1.0 if the agent touches the goal.
     * -1.0 if the agent falls off the platform.
-* Brains: One brain with the following observation/action space.
+* Brains: Two brain, each with the following observation/action space.
```
mmattar: Two brains (?)

mmattar: As in, should it be Two brains (plural)?

PR #535, opened 2018-03-22
Update Installation.md

ukarroum:
By default Unity use .Net 3.5 which can cause trouble because the ML-Agents is using the function Clear on StringBuilder which is supported by .Net 4.0 or above


eshvk:
@ukarroum Please change the remote branch for your PR to the `develop` branch and we can consider it. This contains current code that we are working on. Currently master will contain only released code.

jo3w4rd:
The dependency on .NET 4 when you are not using an Internal brain is under discussion. @eshvk since this is only a documentation change, we can make a "hotfix" exception to the normal git flow process. I don't have a problem including a note until we decide what to do about the dependency issue.

xiaomaogy:
We might move this into a FAQ page.

xiaomaogy:
I've made this change into the FAQ page in the develop branch, so I am closing this PR. Thanks for your suggestion which triggered our change.

PR #578, opened 2018-04-04
fixed all the line ending

xiaomaogy:


xiaomaogy:
This is a fix that changes all the line ending from CRLF to LF.

PR #580, opened 2018-04-04
Report means instead of totals for losses

awjuliani:
Change how model losses get computed for display in Tensorboard. 

* Replace totals over epochs and batches with per-sample means of magnitude, which are more interpretable. 
* Adjust documentation where necessary.


mmattar:
@awjuliani Would this PR require any corresponding changes to our TensorBoard screenshot images?

awjuliani:
@mmattar Yeah, they should ideally be changed. I will create new ones.

PR #581, opened 2018-04-04
Use Beta Distribution for Continuous Control Policy

awjuliani:
## About

Replaces Gaussian distribution with Beta distribution for continuous control model policy output. 

* Implementation of: http://proceedings.mlr.press/v70/chou17a/chou17a.pdf .
* Improves overall cumulative reward and training stability on Crawler and Reacher (see graph below). Other CC environments provide equal or better performance as well.
* Provides bounded output of CC models (between 0 and 1). This has the nice property of giving environment creators an assurance about the range of values to expect from the model.
* Added `ScaleContinuousAction()` function to Unity to scale beta output from [0, 1] to any arbitrary range.
* Retrained all CC models using beta distribution.

While this new approach has nice theoretical and practical benefits, it may be the case that it causes regressions in performance under some circumstances. As such, I'd really appreciate if others could stress test it, especially @Hunter-Unity on some of his new high-dimensionality locomotion tasks to ensure performance increases there too.

### Crawler performance (Beta: blue; Gaussian: grey)

screen shot 2018-04-04 at 4 39 29 pm


python/unitytrainers/ppo/trainer.py
```diff
@@ -360,8 +359,8 @@ def update_model(self):
                              self.model.all_old_probs: np.array(
                                  _buffer['action_probs'][start:end]).reshape([-1, self.brain.vector_action_space_size])}
                 if self.is_continuous_action:
-                    feed_dict[self.model.epsilon] = np.array(
-                        _buffer['epsilons'][start:end]).reshape([-1, self.brain.vector_action_space_size])
+                    feed_dict[self.model.output] = np.array(
+                        _buffer['actions'][start:end]).reshape([-1, self.model.a_size])
```
vincentpierre: You are setting the output of the model ? This does not seem right, can you explain what you did ?

unity-environment/Assets/ML-Agents/Examples/3DBall/3DBall.unity
```diff
@@ -13,7 +13,7 @@ OcclusionCullingSettings:
 --- !u!104 &2
 RenderSettings:
```
vincentpierre: Was it necessary to push this file ?

awjuliani: Yes. We are using a new `.bytes` file for 3DBall.

unity-environment/Assets/ML-Agents/Scripts/CoreBrainPlayer.cs
```diff
@@ -77,7 +78,7 @@ public void DecideAction(Dictionary<Agent, AgentInfo> agentInfo)
         {
             foreach (Agent agent in agentInfo.Keys)
             {
-                var action = new float[brain.brainParameters.vectorActionSize];
+                float[] action = Enumerable.Repeat(0.5f, brain.brainParameters.vectorActionSize).ToArray();
```
vincentpierre: I think there is a strong assumption here that the developer will ALWAYS have continuous actions rescaled with opposite min and max values. I do not think we should make this assumption. Instead, the NN should output between -1 and 1 and the rescale should be a multiplication instead of the `ScaleContinuousAction` method.

python/unitytrainers/ppo/trainer.py
awjuliani: You set the output, and the backpropagation process calculates what was required to achieve it. This avoids having to cache the random values used in the sampling.

mmattar:
@awjuliani To address regression concerns, would it make sense to have the choice of the distribution a trainer configuration? Would facilitate experimentation in the future as our environments increase in complexity.

python/unitytrainers/ppo/trainer.py
mmattar: @awjuliani Your response above would be a great code comment :)

Sohojoe:
I adapted my [MujocoUnity](https://github.com/Sohojoe/MujocoUnity) project to work with ml-agents and tested this PR with Walker2d and Hopper - In both cases it performed about the same. 

**Walker2d old (grey) vs new (red) 300k steps**
walker old grey vs new red

**Hopper old (green) vs new (blue) 300k steps**
hopper old green vs new blue

**Notes:**
* Initially Hopper totally failed - this was due to the agressive termination function (whereby it would terminate if the head rotated more than a few degress - this was based on OpenAI's implementation). Once I removed that, it worked. hopper with openai termination
* I tried 1m steps and got the same result
* I tried 3m steps and things exploded (when to 0 as if the training was starting from scratch and never improved) at about 2.75m steps
* I used the hyperparamters stated [here](https://github.com/Sohojoe/MujocoUnity/blob/master/Docs/ml_agents.md) (the top set I used for the 3m step test. All other tests where with the 2nd set)
* Entropy was always negative 
 entropy hopper old green vs new blue

awjuliani:
@Sohojoe,

Thanks for doing this testing! It is super helpful to have additional data points. I also love the work you are doing to get mujoco equivalant environment within Unity. It would be great to get those more integrated into ml-agents as additional default environments, if you are interested. 

One of the reasons for the gap in performance I was seeing might be because we have been clipping the Gaussian output to [-1, 1], which was exacerbating the assignment problem that using the Beta distribution is meant to fix. I'll test Crawler again with clipping at [-3, 3]. 

The dropping to of reward to 0 is unfortunately due to a numerical instability problem in TensorFlow when using `tf.Distributions.Beta`. I've got some ideas for workarounds, but if we can't fix it easily, then it may not make sense to switch just yet.

@mmattar @vincentpierre 

I will change things so that beta and gaussian are both options which can be chosen as a hyperparameter, and that both provide the same default scale of [-x, x] (whichever is most performant). For now we can default to Gaussian, with optional Beta, and once I can prevent the numerical issue, then we can consider moving it to default.

Sohojoe:
@awjuliani 

It would be great to have the mujoco environments as default ml-agents environments and I am starting to work towards that. First, I am working to get mujoco.humanoid back running as it may require some refactoring in the shared code. Once I have it back running I’ll create a PR to get things going.

How did it go with beta on crawler with clipping at [-3, 3]?

Quick question: does a python training step (i.e. max_steps: 3e5) increment for each brain or does it include all brains? I had assumed the former (in which case PPO is performing well vs OpenAI.Baselines.DPPG). However, It seams that the wall clock training time decreases when adding more brains with the same max_step.

One thought for your humanoid branches - I was able to get down to two float observations per motor (local rotation & velocity) + 6 global floats (x,z position y rotation of head + x,y,z velocity) - I believe I got this from OpenAI.Roboschool

awjuliani:
@Sohojoe It turns out the [-3, 3] clipped Gaussian outperforms even Beta on the Crawler, getting a new top score of 3808. I'm going to close this PR for now, since it looks like the wider gaussian gives similar performance, and critically doesn't have the numerical stability issue.

screen shot 2018-04-16 at 11 53 05 am

As for the step count, it gets incremented every `env.step()`, but is kep separately per brain. The likely reason for the increased wall clock time is that we currently don't do any parallelization of the training process, so brain trainers run sequentially every step. We hope to be changing that in the near future. 

Thanks for looking into our humanoids. I'll let @Hunter-Unity know, as he is currently the one working on it.

PR #583, opened 2018-04-06
zh-localization docs

xiaomaogy:


xiaomaogy:
Still in the process of resolving conflicts..

PR #589, opened 2018-04-07
CrawlerLegContact incorrectly refers to Ground

sterlingcrispin:
No game object is named "Platform" therefore the legs never make contact with anything, this fixes that


awjuliani:
Thanks for making this fix, @sterlingcrispin !

PR #649, opened 2018-04-18
Use Clipped Gaussian

awjuliani:
This PR makes the following changes:

* Moves clipping of continuous control model into model itself. Output is now always [-1, 1].
* Internal model values are now clipped between [-3, 3] before being rescaled to [-1, 1] for output. This improves training performance by providing a wider range of values within which the pdf of the gaussian can fall. Output of [-1, 1] is used to be more environment-creator friendly.
* Fixes issue where `epsilon` was erroneously being used to reconstruct old probabilities during PPO update, leading to reduced learning performance.
* Introduce `ScaleAction()` function within python to easily rescale values from [-1, 1] to arbitrary range.
* Re-train all CC models using improved algorithm. All performance levels are equal or improved. In the case of Crawler, improvement is drastic (see graph below).
* Update documentation appropriately. 
* Made miscellaneous minor code style and optimization improvements within environments.

Crawler Performance (Green new; Grey old):

screen shot 2018-04-17 at 4 43 56 pm


unity-environment/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -756,6 +756,20 @@ public void UpdateTextAction(string textActions)
...
+    /// Scales continous action from [-1, 1] to arbitrary range.
+    /// </summary>
+    /// <param name="rawAction"></param>
+    /// <param name="min"></param>
+    /// <param name="max"></param>
+    /// <returns></returns>
+    protected float ScaleAction(float rawAction, float min, float max)
+    {
+        var middle = (min + max) / 2;
+        var range = max - min;
```
vincentpierre: I think this is `var range = (max - min) / 2`
Also, I don't think this is needed, or it should go into a utils class ?

vincentpierre:
@xiaomaogy Can we have some tests of TensorFlowSharp with Android and iOS for this branch since it uses new operators?

xiaomaogy:
@vincentpierre I am still working on the iOS part, but once I finish that I will test the plugin against this branch.

sterlingcrispin:
Hey this is awesome! My agent is performing so much better I need to redesign my environment to keep the agent from falling off the ends, this is the same training with the same entropy/reward except on this new branch, the steep upward light blue trend being this branch VS the main branch

gausppo

MarcoMeter:
Hey @awjuliani 
did you base your clipping approach/implementation on a publication (just wondering)?
This is [one paper](https://arxiv.org/abs/1802.07564) which I came across while looking for clipped gaussian.

Issue #148, opened 2017-11-20
Need help understanding the training result!


matonglidewazi:
Hi,

Thanks for this platform, it is awesome, compatible with almost everything! 

I'm  a conducting experiment with OpenBCI (Brain wave capture device),

I just connected the devices and solved communication between it and Unity, now we can use the brainwave as the input state, the program runs smoothly, 

and our action is the Color, HSV stimulating the eyes of people,

Currently I've done small demos, lasting for around 4 minutes, 8 minutes and 15 minutes,

the results are as follow:

![ask](https://user-images.githubusercontent.com/22267548/33012662-9106ca1a-ce1c-11e7-9b63-2b894317c45e.JPG)

The problem is, although the reward is climbing (the alpha wave stable, which is our target), the entropy 
decline very little. **I'm having trouble understanding the entropy here**. In a classification task, the cross entropy could be regard as a benchmark of the model precision, but what does entropy means in PPO?

And since it's a human experiment, our device is not so light and the tester has to face the screen. In a word, the time span for one experiment is limited to no more than 20 mins, or the tester would behave abnormaly. Is there any method to accelerate this process?

Thanks for your time,

awjuliani:
Hi @matonglidewazi,

This is very cool work that you are doing, and I am happy to provide support on this!

In this case, entropy refers to the "randomness" in the policy. If you are using discrete actions, then you can expect this to decrease over time toward 0, which would correspond to the probability distribution over actions all focusing on a single action. When using discrete control, the `beta` parameter controls the strength of the entropy regularization, which encourages exploration. However, if `beta` is too large, then the policy will remain random and never learn specific actions. The opposite can happen where the policy quickly becomes deterministic in a "local minima." If you are using continuous actions however, then the entropy is learned, and there is no `beta` to adjust.

Using reinforcement learning with human data is a unique challenge. As you mention, with participants which can only provide 20 mins of data at a time, there is only so much which can be collected. If the data you have now isn't sufficient, I would recommend exploring off-policy RL methods (Such as DQN for discrete control, or DDPG for continuous control), which could be used to gather data from many different participants, and then collected together to learn a single policy.

matonglidewazi:
Hi @awjuliani 

Yeah thanks for your help!

In this case we are using continous actions, but we don't know when to stop. 

[...]

PR #664, opened 2018-04-21
Patch 4

sterlingcrispin:
clarified functionality of Beta hyperparamter based on https://github.com/Unity-Technologies/ml-agents/issues/148#issuecomment-345790241 from @awjuliani


awjuliani:
Hi @sterlingcrispin This is actually no longer the case. We now use entropy regularization for both continuous and discrete control models. As such, `beta` has the same interpretation in both cases.

PR #684, opened 2018-04-25
Update environment.py

VETURISRIRAM:
Have made some changes related to the laggy connections establishments.

### DIFF
--- a/python/unityagents/environment.py
+++ b/python/unityagents/environment.py
@@ -258,9 +258,12 @@ class UnityEnvironment(object):
         :param bw:
         :return:
         """
-        s = self._recv_bytes()
-        s = self._process_pixels(image_bytes=s, bw=bw)
-        self._conn.send(b"RECEIVED")
+        try:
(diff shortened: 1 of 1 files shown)


eshvk:
Hi @VETURISRIRAM:
 Couple of points:
1. In general, a PR is supposed to be one semantically contained unit. Your two PR(s) would need to be squashed and sent together as a single PR of one commit for review. 
2. I think adding try-catch in general is good, I am less convinced that there is a good rationale for:
  a. The new Exception class which inherits from `UnityException`. Seems very specific and the use of it can be replaced by a suitable log message. Also, it `UnityOverTimeException` and `UnityTimeOutException` are similar enough as variable names that it makes it confusing. 
  b. Not really understanding why we have a nested try catch pattern as opposed to try with multiple catch statements

awjuliani:
Hi @VETURISRIRAM,

We will be replacing the current communicator with a new system in the next ML-Agents release. As such, I will be closing this PR.

PR #685, opened 2018-04-25
Added an exception class

VETURISRIRAM:
Changes to avoid laggy connection establishments.

### DIFF
--- a/python/unityagents/exception.py
+++ b/python/unityagents/exception.py
@@ -13,7 +13,12 @@ class UnityEnvironmentException(UnityException):
     """
     pass
 
-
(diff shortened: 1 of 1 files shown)


eshvk:
Can you please explain why this class was added? The functionality is not being used anywhere.

VETURISRIRAM:
The class was added to handle the laggy  socket connections. It is always a good practice in socket programming to have such a class to look for the connections that take long time and try to revert the requests.

awjuliani:
Hi @VETURISRIRAM,

As mentioned in the other thread, we are moving to a new system for communication, and as such will be closing this.

PR #730, opened 2018-05-08
Add ignore for plugins folder

awjuliani:
The Rider IDE creates additional files in the Unity project which should be ignored in version control. Since all ML-Agents specific plugins are located within the `ML-Agents` sub-folder, I propose this repo ignore the high-level Plugins folder.


xiaomaogy:
Are you switching from Visual Studio to Rider?

awjuliani:
@xiaomaogy Rider is what I typically use.

xiaomaogy:
@awjuliani Oh I thought Visual Studio is the recommend IDE we use and that's why I use it. But I've always preferred the JetBrains series, will switch to Rider in the future.

PR #739, opened 2018-05-10
Curiosity Driven Exploration & Pyramids Environments

awjuliani:
* Adds implementation of *Curiosity-driven Exploration by Self-supervised Prediction* (https://arxiv.org/abs/1705.05363) to PPO trainer.
* To enable, set `use_curiosity` flag to `true` in hyperparameter file.
* Includes refactor of `unitytrainers` model code to accommodate new feature.
* Adds new `Pyramids` environment (w/ documentation). Environment contains sparse reward, and can only be solved using PPO+Curiosity.

In particular looking for feedback on python-side refactoring, as there may be nicer way of doing things.


mmattar:
@awjuliani How do the results change by enabling curiosity for our current environments?

python/unitytrainers/models.py
```diff
@@ -44,109 +56,137 @@ def create_visual_input(o_size_h, o_size_w, bw, name):
         visual_in = tf.placeholder(shape=[None, o_size_h, o_size_w, c_channels], dtype=tf.float32, name=name)
         return visual_in
 
-    def create_vector_input(self, s_size):
+    def create_vector_input(self):
```
vincentpierre: Should `create_vector_input` take in a name as argument in the same way `create_visual_input` does ?

python/unitytrainers/models.py
```diff
@@ -159,7 +199,8 @@ def create_new_obs(self, num_streams, h_size, num_layers):
             final_hiddens.append(final_hidden)
         return final_hiddens
 
-    def create_recurrent_encoder(self, input_state, memory_in, name='lstm'):
+    @staticmethod
+    def create_recurrent_encoder(input_state, memory_in, sequence_length, name='lstm'):
         """
```
vincentpierre: For uniformity, maybe the name in all those methods should be an optional argument (same with reuse) ?

python/unitytrainers/ppo/models.py
```diff
@@ -44,6 +49,92 @@ def create_reward_encoder():
         update_reward = tf.assign(last_reward, new_reward)
         return last_reward, new_reward, update_reward
 
+    def create_curiosity_encoders(self):
```
vincentpierre: Maybe for latter : Each encoder could be in its own script ?

python/unitytrainers/ppo/models.py
```diff
@@ -44,6 +49,92 @@ def create_reward_encoder():
...
+                next_visual_encoders.append(encoded_next_visual)
+
+            hidden_visual = tf.concat(visual_encoders, axis=1)
+            hidden_next_visual = tf.concat(next_visual_encoders, axis=1)
+            inverse_input_list.append(hidden_visual)
+            inverse_input_list.append(hidden_next_visual)
+            encoded_state_list.append(hidden_visual)
+            encoded_next_state_list.append(hidden_next_visual)
+
+        if self.o_size > 0:
```
vincentpierre: I think we have a better way to check for continuous vector observation than than. It might cause problems in scenarios where the vector observation is a continuous single float.

python/unitytrainers/ppo/models.py
```diff
@@ -44,6 +49,92 @@ def create_reward_encoder():
...
         return last_reward, new_reward, update_reward
 
+    def create_curiosity_encoders(self):
+        """
+        Creates state encoders for current and future observations.
+        Used for implementation of ﻿Curiosity-driven Exploration by Self-supervised Prediction
+        See https://arxiv.org/abs/1705.05363 for more details.
+        :return: current and future state encoder tensors.
+        """
+        inverse_input_list = []
```
vincentpierre: I cannot see where this is used. Where is this list consumed ?

python/unitytrainers/ppo/models.py
```diff
@@ -44,6 +49,92 @@ def create_reward_encoder():
...
+            encoded_state_list.append(encoded_vector_obs)
+            encoded_next_state_list.append(encoded_next_vector_obs)
+
+        encoded_state = tf.concat(encoded_state_list, axis=1)
+        encoded_next_state = tf.concat(encoded_next_state_list, axis=1)
+        return encoded_state, encoded_next_state
+
+    def create_inverse_model(self, encoded_state, encoded_next_state):
+        """
+        Creates inverse model TensorFlow ops for Curiosity module.
```
vincentpierre: Maybe add something like : Predicts the action taken between state and next state.

python/unitytrainers/ppo/models.py
```diff
@@ -44,6 +49,92 @@ def create_reward_encoder():
...
+        :param encoded_state: Tensor corresponding to encoded current state.
+        :param encoded_next_state: Tensor corresponding to encoded next state.
+        """
+        combined_input = tf.concat([encoded_state, encoded_next_state], axis=1)
+        if self.brain.vector_action_space_type == "continuous":
+            pred_action = tf.layers.dense(combined_input, self.a_size, activation=None)
+            squared_difference = tf.reduce_sum(tf.squared_difference(pred_action, self.selected_actions), axis=1)
```
vincentpierre: Where is self.selected_actions created in continuous control. In discrete, it is in create_dc_actor_critic. I think self.selected_actions should be in create_dc_actor_critic then. Did you try curiosity with a continuous control environment to make sure self.selected_action is used ?

python/unitytrainers/ppo/trainer.py
```diff
@@ -231,31 +260,41 @@ def add_experiences(self, curr_all_info: AllBrainInfo, next_all_info: AllBrainIn
...
+                        self.training_buffer[agent_id]['actions_pre'].append(actions_pre[idx])
                     a_dist = stored_take_action_outputs[self.model.all_probs]
                     value = stored_take_action_outputs[self.model.value]
                     self.training_buffer[agent_id]['actions'].append(actions[idx])
-                    if self.is_continuous_action:
-                        self.training_buffer[agent_id]['actions_pre'].append(actions_pre[idx])
                     self.training_buffer[agent_id]['prev_action'].append(stored_info.previous_vector_actions[idx])
                     self.training_buffer[agent_id]['masks'].append(1.0)
-                    self.training_buffer[agent_id]['rewards'].append(next_info.rewards[next_idx])
+                    if self.use_curiosity:
```
vincentpierre: Would it make more sense to always have an intrinsic reward but only have it non-zero when curiosity is used ?

python/unitytrainers/ppo/trainer.py
```diff
@@ -345,7 +390,7 @@ def update_model(self):
         """
         num_epoch = self.trainer_parameters['num_epoch']
         n_sequences = max(int(self.trainer_parameters['batch_size'] / self.sequence_length), 1)
-        total_v, total_p = [], []
+        total_v, total_p, total_f, total_i = [], [], [], []
```
vincentpierre: Maybe it is time to find better names for these variables or is this outside of the scope of this PR ?

unity-environment/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -675,6 +675,11 @@ protected void AddVectorObs(Quaternion observation)
         info.vectorObservation.Add(observation.z);
         info.vectorObservation.Add(observation.w);
     }
+    
+    protected void AddVectorObs(bool observation)
```
vincentpierre: Add a `<summary>`

unity-environment/Assets/ML-Agents/Examples/Pyramids/Scripts/PyramidSwitch.cs
```diff
@@ -0,0 +1,48 @@
...
+    private bool state;
+    private GameObject area;
+    private PyramidArea areaComponent;
+
+    public bool GetState()
+    {
+        return state;
+    }
+
+    private void Start()
```
vincentpierre: InitializeAgent?

unity-environment/Assets/ML-Agents/Examples/Pyramids/Scripts/PyramidSwitch.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        area = gameObject.transform.parent.gameObject;
+        areaComponent = area.GetComponent<PyramidArea>();
+        Reset();
```
vincentpierre: Can we name Reset ResetSwitch ?

python/unitytrainers/ppo/models.py
```diff
@@ -44,6 +49,92 @@ def create_reward_encoder():
...
+            inverse_input_list.append(encoded_vector_obs)
+            inverse_input_list.append(encoded_next_vector_obs)
+            encoded_state_list.append(encoded_vector_obs)
+            encoded_next_state_list.append(encoded_next_vector_obs)
+
+        encoded_state = tf.concat(encoded_state_list, axis=1)
+        encoded_next_state = tf.concat(encoded_next_state_list, axis=1)
+        return encoded_state, encoded_next_state
+
+    def create_inverse_model(self, encoded_state, encoded_next_state):
```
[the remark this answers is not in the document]
awjuliani: Good point.

python/unitytrainers/ppo/models.py
awjuliani: Can you elaborate? Are you thinking of a `curiosity.py`?

python/unitytrainers/ppo/models.py
awjuliani: What would you suggest?

python/unitytrainers/ppo/trainer.py
awjuliani: If we did that we'd need some code to generate zeroes and keep track of them, which doesn't seem to be as elegant.

python/unitytrainers/ppo/trainer.py
awjuliani: I will make them less terrible here. It is justified since we are adding new ones.

unity-environment/Assets/ML-Agents/Examples/Pyramids/Scripts/PyramidSwitch.cs
awjuliani: Yes. We. Can.

python/unitytrainers/ppo/models.py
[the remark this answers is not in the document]
awjuliani: Actually, this may require a larger refactoring. I think it would be best to wait as part of a our larger changes we will be making to trainers in the future.

python/unitytrainers/ppo/models.py
```diff
@@ -35,6 +36,8 @@ def __init__(self, brain, lr=1e-4, h_size=128, epsilon=0.2, beta=1e-3, max_step=
         else:
             self.create_dc_actor_critic(h_size, num_layers)
         if self.use_curiosity:
+            self.encoding_size = encoding_size
```
vincentpierre: encoding_size could have a more explicit name.

python/unitytrainers/ppo/models.py
```diff
@@ -35,6 +36,8 @@ def __init__(self, brain, lr=1e-4, h_size=128, epsilon=0.2, beta=1e-3, max_step=
...
+            self.curiosity_strength = curiosity_strenght
```
vincentpierre: typo

python/trainer_config.yaml
```diff
@@ -17,13 +17,22 @@ default:
     sequence_length: 64
     summary_freq: 1000
     use_recurrent: false
+    use_curiosity: false
```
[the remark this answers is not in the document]
awjuliani: There is no model learned when `use_curiosity` is false. Are you suggesting we don't learn the model when strength == 0?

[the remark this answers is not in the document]
vincentpierre: Yes, and 0 is the default.

python/unitytrainers/ppo/models.py
vincentpierre: `curiosity_encoding_size` or `forward_model_h_size` or`curiosity_h_size`

python/unitytrainers/ppo/trainer.py
```diff
@@ -43,6 +43,7 @@ def __init__(self, sess, env, brain_name, trainer_parameters, training, seed):
         self.use_recurrent = trainer_parameters["use_recurrent"]
         self.use_curiosity = bool(trainer_parameters['use_curiosity'])
         self.sequence_length = 1
+        self.has_updated = False
```
vincentpierre: Can you put a comment describing what it does ?
I think it could also benefit a better name. From the code, I think it is something like `had_first_update`

unity-environment/ProjectSettings/TagManager.asset
```diff
@@ -25,6 +25,7 @@ TagManager:
   - switchOff
   - pyramid
   - switchOn
+  - stone
```
vincentpierre: So many tags... Can we have themn more generic like `object0`, `object1`, `enemy0`...

python/unitytrainers/ppo/models.py
```diff
@@ -8,7 +8,8 @@
 
 class PPOModel(LearningModel):
     def __init__(self, brain, lr=1e-4, h_size=128, epsilon=0.2, beta=1e-3, max_step=5e6,
-                 normalize=False, use_recurrent=False, num_layers=2, m_size=None):
+                 normalize=False, use_recurrent=False, num_layers=2, m_size=None, use_curiosity=False,
+                 curiosity_strength=0.01, encoding_size=128):
```
vincentpierre: Please rename encoding_size to something containing curiosity

PR #770, opened 2018-05-18

animaleja32:
Shouldn't Done(); be placed after the rewards are given?


awjuliani:
Hi @animaleja32,

Technically the order does not matter, since `Done` only takes effect at the end of the engine step, rather than immediately within the code. Still though, it probably makes more conceptual sense to put it afterwards in tutorial material. Thanks for submitting this.

PR #782, opened 2018-05-20
Update Installation-Windows.md

emaines:
Would fail with current version if ran 'pip install' : "ERROR: You must give at least one requirement to install (see "pip help install")". 'pip install -r requirements.txt' works.


emaines:
Ok, thanks,That'll teach me for installing AI frameworks in the wee bit hours of the morning :P Cancelling the pull request.

PR #786, opened 2018-05-21
Fixed path variables for Anacondo3

LeighS:
The proposed path variables were incorrect. The paths after installing Anaconda3 were actually in the %UserProfile% location.


xiaomaogy:
Thanks for your contribution, I've tested this myself and the issue you pointed out is correct.

PR #863, opened 2018-06-15
fix all the places that used v0.3 and should be fixed, also updated t…

xiaomaogy:
…he migration doc


docs/Migrating.md
```diff
@@ -1,4 +1,9 @@
-# Migrating to ML-Agents v0.3
+# Migrating from ML-Agents v0.3 to ML-Agents v0.4
+
+## Unity API
+ * `using MLAgents;` needs to be added in all of the C# scripts that uses ML-Agents.
```
awjuliani: `uses` should be `use`.

PR #896, opened 2018-06-20
Update sky color for walker, crawler, and pyramids

awjuliani:


awjuliani:
@xiaomaogy The walker looks strange with only one, and the crawler looks strange with more than one 😜

PR #901, opened 2018-06-21
Replace CC `tanh` activation with `swish`

awjuliani:
Improves performance by ~25% on Walker and Crawler. Slight adjustments to 3DBall to ensure no performance loss. 

Crawler and Walker models on master were actually trained using `swish`, so making the PR into hotfix branch for consistency.

### DIFF
--- a/python/trainer_config.yaml
+++ b/python/trainer_config.yaml
@@ -118,6 +118,7 @@ Ball3DBrain:
     buffer_size: 12000
     summary_freq: 1000
     time_horizon: 1000
+    lambd: 0.99
     gamma: 0.995
     beta: 0.001
 
--- a/python/trainer_config.yaml
+++ b/python/trainer_config.yaml
@@ -114,7 +114,7 @@ VisualPyramidBrain:
 
 Ball3DBrain:
     normalize: true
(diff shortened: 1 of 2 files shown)


xiaomaogy:
@awjuliani Why does changing the batch size from 1200 to 64 for 3dball improves the performance?

xiaomaogy:
@awjuliani Also shall we merge the hotfix-0 branch after this merge?

awjuliani:
@xiaomaogy This was just what empirically seemed to work well. For some reason the value function estimate for 3DBall is relatively poor. My hypothesis is that decreasing the batch size allowed for better gradients for learning the value function. Large batch sizes are actually bad in most problems. They are only really used in training continuous control models because it is "good" that the gradients all wash out, since it doesn't push the policy too much in any direction. I think this is likely bad for the value estimate, since it means it may not be able to keep up with the true value function of the environment. The reason for changing the `lambda` parameter is similar. By increasing it, the algorithm relies less on the value estimate when producing the advantage, as as such isn't as negatively effected by the estimate being poor. 

Of course the question of why this wasn't an issue with the `tanh` is a bit of a mystery to me still.

PR #906, opened 2018-06-22
Remove Discrete Observations

awjuliani:
* Remove discrete observations as separate category of vector observations. As such, there are now no `discrete` or `continuous` vector observations, only vector observations.
* Environments which use discrete observations can now use `AddVectorObs(index, range)` in order to add one-hot vectors to the observations.
* Modified documentation appropriately.
* Changed `Basic` environment appropriately, and re-trained model.
* Rebuilt protobuf definitions to reflect this change.
* Upgrade protobuf to `3.6.0`.


awjuliani:
Thanks for pointing this out @xiaomaogy. I have made the changes.

PR #923, opened 2018-06-26
Fix doc which refers to nonexistent agents.

dericp:
The documentation refers to an "agent on the left" and one on the right. Neither of these agents seem to exist since the image is a line graph.


awjuliani:
hey @dericp,

Can you make this PR into `hotfix-0`?

Issue #911, opened 2018-06-24
Experimental (.NET 4.6 Equivalent) ?


ivenzor:
According to the documentation (https://github.com/Unity-Technologies/ml-agents/blob/master/docs/Basic-Guide.md) when editing Player Configuration settings you should have .NET **4.6** in both Scripting Runtime Version and Api Compatibility Level.

However in my installation I can only choose .Net 3.X Equivalent or .Net _4.X_ Equivalent.
Was my environment installed properly or something is missing regarding that specific  .NET 4.6 option I don't have?

I'm using Unity 2018.1.5f1 on Win10.

xiaomaogy:
You should choose the .NET 4.X Equivalent. .NET 4.6 was from the Unity 2017.x version. I will update the documentation accordingly. Thanks for the notes.

ivenzor:
Many thanks for your reply.

Issue #909, opened 2018-06-22
Please, update the documentation: buffer_size is not used in BC python code


icaro56:
In the documentation below, is writen that buffer_size is used in both PPO and BC:

https://github.com/Unity-Technologies/ml-agents/blob/master/docs/Training-ML-Agents.md

I am study the Behavioral Cloning Trainer and Model Python code and I did not find buffer_size being used.

xiaomaogy:
@icaro56 It is used in trainer_controller.py: 262, is_ready_update() function.

icaro56:
@xiaomaogy , only in PPO buffer_size is used inside is_ready_update.

Another hyperparameter that is not used by BC is the beta.

icaro56:
Please, mark this issue as bug.

I spent a long time changing values of hyperparameters that were not even used. Correcting this information in the documentation will help others.

xiaomaogy:
@icaro56 Yes you are correct, this is a documentation error on our side. Basically in ppo we will use the buffer to collect a certain number of experiences and train on it, then clear the buffer and recollect the experiences, and in this process the buffer size is a constant number. In behavioral cloning we store all of the training data in the buffer, and the buffer size keeps increasing, and thus we don't have a constant buffer size. 

I will update the documentation accordingly. Thanks for your help in pointing this out.

icaro56:
**Another hyperparameters that are not used by BC are the beta**

Beta are not used by BC.

xiaomaogy:
Hi @icaro56, thanks for raising this issue! There is the new PR that fixes your issue which will be reflected in release-v0.5. https://github.com/Unity-Technologies/ml-agents/pull/1193

xiaomaogy:
Thanks for reaching out to us. Hopefully you were able to resolve your issue.  We are closing this due to inactivity, but if you need additional assistance, feel free to reopen the issue.

PR #924, opened 2018-06-27
fixed the doc according to the issue

xiaomaogy:


xiaomaogy:
This PR is a fix for these two issues. 

https://github.com/Unity-Technologies/ml-agents/issues/911
https://github.com/Unity-Technologies/ml-agents/issues/909

Issue #927, opened 2018-06-27
BatchVisualObservations() is not optimized.


tcmxx:
BatchVisualObservations() in CoreBrainInternal is very slow. 

I think There are couple of reasons why:
1. Based on profiling data, textures[b].height is very slow for some reason.
2. pixel colors are not access in a continuous memory pattern.

The code I am using is like this, not sure if it is what you want but it is much faster

```    public static float[,,,] BatchVisualObservations(
        this List<Texture2D> textures, bool blackAndWhite)
    {
        int batchSize = textures.Count;
        int width = textures[0].width;
        int height = textures[0].height;
        int pixels = 0;
        if (blackAndWhite)
            pixels = 1;
        else
            pixels = 3;
        float[,,,] result = new float[batchSize, height, width, pixels];
        float[] resultTemp = new float[batchSize * height * width * pixels];
        int hwp = height * width * pixels;
        int wp = width * pixels;
        for (int b = 0; b < batchSize; b++)
        {
            Color32[] cc = textures[b].GetPixels32();
            for (int h = height - 1; h >= 0; h--)
            {
                for (int w = 0; w < width; w++)
                {
                    Color32 currentPixel = cc[(height - h - 1) * width + w];
                    if (!blackAndWhite)
                    {
                        // For Color32, the r, g and b values are between
                        // 0 and 255.
                        /*result[b, height - h - 1, w, 0] =
                            currentPixel.r / 255.0f;
                        result[b, height - h - 1, w, 1] =
                            currentPixel.g / 255.0f;
                        result[b, height - h - 1, w, 2] =
                            currentPixel.b / 255.0f;*/

                        resultTemp[b * hwp + h * wp + w * pixels] = currentPixel.r / 255.0f;
                        resultTemp[b * hwp + h * wp + w * pixels + 1] = currentPixel.g / 255.0f;
                        resultTemp[b * hwp + h * wp + w * pixels + 2] = currentPixel.b / 255.0f;

                        /*result[b,h,w,0] = currentPixel.r / 255.0f;
                        result[b, h, w, 1] = currentPixel.g / 255.0f;
                        result[b, h, w, 2] = currentPixel.b / 255.0f;*/
                    }
                    else
                    {
                        /*result[b, height - h - 1, w, 0] =
                            (currentPixel.r + currentPixel.g + currentPixel.b)
                            / 3;*/
                         resultTemp[b * hwp + h * wp + w * pixels] =
[...]

PR #940, opened 2018-06-29
Optimize BatchVisualObservations

tcmxx:
for larger input resolution and faster fixedupdate rate the improvement should be obvious. #927


awjuliani:
Thanks for making this PR, @tcmxx !

PR #967, opened 2018-07-11
Fixed Python tests.

dericp:
Discrete observations were removed but the tests were not updated to reflect this.

- Removed all references to vector_observation_space_type and vectorObservationSpaceType in Python tests.
- Tests are all now passing.

Every Unity environment in the repo references seems to still reference vectorObservationSpaceType when I do a search for "vectorObservationSpaceType". Is this okay?


xiaomaogy:
@dericp I've tried your branch and it looks good. Regarding why the vectorObservationSpaceType is still in all of the .unity files, I am not sure. The vectorObservationSpaceType is removed from the Brain script menu, but it is still in the file. Maybe it is just some sort of cache that we can ignore for now.

PR #968, opened 2018-07-11
Curriculum learning moved from environment to trainer.

dericp:
- Fixed some typos in documentation.
- Moved `curriculum.py` into `python/unitytrainers`.
- Updated test accordingly.

Questions:
- I'm not sure if I'm importing the `Curriculum` class into `trainer_controller.py` the preferred way.
- Should reset parameters become a property of the environment?


python/unitytrainers/curriculum.py
```diff
@@ -1,6 +1,6 @@
 import json
 
-from .exception import UnityEnvironmentException
+from unityagents.exception import UnityEnvironmentException
```
awjuliani: I am not sure we want to use UnityEnvironmentException here.

python/unitytrainers/trainer_controller.py
```diff
@@ -12,8 +12,11 @@
 from tensorflow.python.tools import freeze_graph
 from unitytrainers.ppo.trainer import PPOTrainer
 from unitytrainers.bc.trainer import BehavioralCloningTrainer
+from unitytrainers import Curriculum
 from unityagents import UnityEnvironment, UnityEnvironmentException
 
+from communicator_objects import UnityRLInitializationInput
```
awjuliani: Why is this necessary?

python/unitytrainers/curriculum.py
dericp: Good point. Should I make an exception module under unitytrainers and add a `CurriculumException`?

python/unitytrainers/trainer_controller.py
dericp: Woops, accidentally left that in from some experiments. Will remove.

dericp:
@awjuliani Addressed your comments. If it looks good I'll merge it!

PR #1002, opened 2018-07-24
Hotfix 0.4.0b

awjuliani:
* Corrects observation space description for `PushBlock` environment.
* Fixes bug preventing using environments with python multi-processing.
* Fixes bug preventing agents to be initialized without a brain.

### DIFF
--- a/python/unityagents/rpc_communicator.py
+++ b/python/unityagents/rpc_communicator.py
@@ -15,7 +15,8 @@ logger = logging.getLogger("unityagents")
 
 
 class UnityToExternalServicerImplementation(UnityToExternalServicer):
-    parent_conn, child_conn = Pipe()
+    def __init__(self):
+        self.parent_conn, self.child_conn = Pipe()
 
     def Initialize(self, request, context):
         self.child_conn.send(request)
--- a/unity-environment/Assets/ML-Agents/Scripts/Agent.cs
+++ b/unity-environment/Assets/ML-Agents/Scripts/Agent.cs
@@ -257,10 +257,9 @@ namespace MLAgents
         void OnEnable()
         {
             textureArray = new Texture2D[agentParameters.agentCameras.Count];
-            for (int i = 0; i < brain.brainParameters.cameraResolutions.Length; i++)
+            for (int i = 0; i < agentParameters.agentCameras.Count; i++)
             {
-                textureArray[i] = new Texture2D(brain.brainParameters.cameraResolutions[i].width, 
-                    brain.brainParameters.cameraResolutions[i].height, TextureFormat.RGB24, false);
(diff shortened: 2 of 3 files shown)


unity-environment/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -257,10 +257,9 @@ public abstract class Agent : MonoBehaviour
         void OnEnable()
         {
             textureArray = new Texture2D[agentParameters.agentCameras.Count];
-            for (int i = 0; i < brain.brainParameters.cameraResolutions.Length; i++)
+            for (int i = 0; i < agentParameters.agentCameras.Count; i++)
             {
-                textureArray[i] = new Texture2D(brain.brainParameters.cameraResolutions[i].width, 
-                    brain.brainParameters.cameraResolutions[i].height, TextureFormat.RGB24, false);
+                textureArray[i] = new Texture2D(1, 1, TextureFormat.RGB24, false);
```
xiaomaogy: What is the reason behind this change?

python/unityagents/rpc_communicator.py
```diff
@@ -15,7 +15,8 @@
 
 
 class UnityToExternalServicerImplementation(UnityToExternalServicer):
-    parent_conn, child_conn = Pipe()
+    def __init__(self):
+        self.parent_conn, self.child_conn = Pipe()
```
xiaomaogy: Is this the fix for the bug preventing using environments with python multi-processing.

unity-environment/Assets/ML-Agents/Scripts/Agent.cs
awjuliani: The agent can be initialized without a brain attached, in which case the old code would throw an error. In the new code we use the agent's camera list to initialize the texture2d array. But, since we don't know how big the textures should be, we initialize them to 1x1 to start with.

python/unityagents/rpc_communicator.py
awjuliani: Yep. In the old code they were global variables, which prevented multiple instances of them from being created correctly.

unity-environment/Assets/ML-Agents/Scripts/Agent.cs
xiaomaogy: Why do we want to allow the agent be initialized without a brain attached?

awjuliani: We may want to programmatically attach brains during runtime, as is the case with `WallJump`.

xiaomaogy: Oh I see.

PR #1035, opened 2018-07-30
Fixing bug when no curriculum folder is passed.

dericp:
- The old Curriculum object would accept None
as a location for the curriculum. If the
location was None, it would return default
values as its config and lesson number.

- The new MetaCurriculum does not accept
None as a location for the curriculum
folder. This was done to remove unnecessary
edge case functionality from curriculums.

- None checks have been added into
trainer_controller. In the future,
it should be possible to better refactor
trainer_controller so that these None
checks can be removed. This is preferable
to hard-coding default behavior into
MetaCurriculum objects when a metacurriculum
would not even be in place.

### DIFF
--- a/python/unitytrainers/trainer_controller.py
+++ b/python/unitytrainers/trainer_controller.py
@@ -97,7 +97,7 @@ class TrainerController(object):
         else:
             self.meta_curriculum = MetaCurriculum(self.curriculum_folder, self.env._resetParameters)
 
-        if self.meta_curriculum is not None and self.curriculum_folder is not None:
+        if self.meta_curriculum is not None:
             for brain_name in self.meta_curriculum.brains_to_curriculums.keys():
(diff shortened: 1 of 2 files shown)


python/unitytrainers/trainer_controller.py
```diff
@@ -257,16 +258,22 @@ def start_learning(self):
             else:
                 sess.run(init)
             global_step = 0  # This is only for saving the model
-            self.meta_curriculum.increment_lessons(self._get_progresses())
-            curr_info = self.env.reset(config=self.meta_curriculum.get_config(), train_mode=self.fast_simulation)
+            if self.meta_curriculum is not None:
```
awjuliani: Since these five lines are used twice here, does it make sense to break them out into their own "check curriculum and reset env" function?

dericp:
Got approval to merge offline. Merging.

PR #1037, opened 2018-07-30
Corrected python command

yinghuayang:
Replaced "python3" with "python" and added that people need to replace the run-identifier with their desired name.


jo3w4rd:
Same comment as the related PR: Whether you need to use `python3` or just `python` depends on your platform and Python installation. We have chosen to use `python3` explicitly here to make it clear that Python 3 (and not Python 2) is required.

Issue #972, opened 2018-07-13
Bug with --lesson command line argument


scotthovestadt:
When trying to resume from a previous lesson, providing the `--lesson` command line argument currently does not jump you straight into that lesson. Instead, the first lesson is started. When completed, the lesson counter is incremented from the value provided by `--lesson` and you jump to that lesson.

For example, I provide `--lesson=3`. I expect to start at the 4th lesson (0, 1, 2, 3). Instead, I start at the 1st lesson (0). When I complete that, the lesson counter is incremented (3+1) and I proceed to the 5th lesson (0, 1, 2, 3, 4).

I think I see the area in the code it can be fixed and I would be willing to explore that and file a PR unless you already have a fix coming down the pipeline.

eshvk:
Hey @scotthovestadt!  we appreciate the bug notif and thank you for offering to submit a PR. @dericp is looking into Curriculum Learning and would be able to better answer reg. the PR.

dericp:
Hi @scotthovestadt, I've been reworking this section of the code. I'll make sure to unit test this and fix the bug.

dericp:
@scotthovestadt I believe https://github.com/Unity-Technologies/ml-agents/pull/1043 fixes this bug. It could be great if you could try it on your end!

PR #1043, opened 2018-07-31
Fixing curriculum reset parameter bug.

dericp:
The environment did not respond to reset parameter values defined by a curriculum until the second reset. This was because on the first reset, the reset parameters were not updated by the Academy.

This bug is has been brought up in a couple issues:
- https://github.com/Unity-Technologies/ml-agents/issues/1041
- https://github.com/Unity-Technologies/ml-agents/issues/972

All C# tests are still passing. I have tested this on environments like 3DBall and WallJump. It would be great if I could get more suggestions for how to test these changes.


awjuliani:
Looks good to me. I will defer to @vincentpierre though, since he was responsible for the "first reset" code.

PR #1128, opened 2018-08-26
Printing Unity logo exception is now caught.

dericp:
If an error is raised while attempting to print the big Unity logo, the `learn.py` script will instead print `Unity Technologies`.

### DIFF
--- a/ml-agents/mlagents/learn.py
+++ b/ml-agents/mlagents/learn.py
@@ -69,9 +69,8 @@ def main():
                         ¬`▀▀▀█▓
 
         ''')
-    # TODO: figure this out
     except:
-        print('UNITY!!!')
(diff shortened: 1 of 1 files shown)


ml-agents/mlagents/learn.py
```diff
@@ -69,9 +69,8 @@ def main():
                         ¬`▀▀▀█▓
 
         ''')
-    # TODO: figure this out
     except:
-        print('UNITY!!!')
+        print('Unity Technologies')
```
vincentpierre: Can we do something more like
```python
print('\n\n\tUnity Technologies\n')
```

dericp: I like it!

PR #1142, opened 2018-08-28
changed the version number to 2017.4

xiaomaogy:


xiaomaogy:
Still testing if it is working with 2017.1 after it is switched to 2017.4.

PR #1146, opened 2018-08-29
Added checkbox to turn action masking on/off

mmattar:


mmattar:
@vincentpierre - added fixed and re-ran with (orange line) and without (blue line) masking.

results

PR #1157, opened 2018-08-30
Adding specific mlagents-learn vs. learn.py example.

dericp:


dericp:
@unityjeffrey Good point, I'll make it more clear in the doc.

PR #1162, opened 2018-08-30
Release v0.5 debug core internal

xiaomaogy:


xiaomaogy:
Don't review, for compare only

PR #1173, opened 2018-09-04
Project Version 2017.4

awjuliani:
Due to a bug in 2017.1, we are setting the default version of the UnitySDK project to 2017.4.


awjuliani:
@xiaomaogy Can you update the migrating document?

docs/Installation.md
```diff
@@ -4,7 +4,7 @@ To install and use ML-Agents, you need install Unity, clone this repository and
 install Python with additional dependencies. Each of the subsections below
 overviews each step, in addition to a Docker set-up.
 
-## Install **Unity 2017.1** or Later
+## Install **Unity 2017.4.10f1** or Later
```
dericp: Does it have to be 2017.4._10_ or later? Could it be any version of 2017.4?

awjuliani: I think we should just say `2017.4`

xiaomaogy: But we haven't tested on 2017.4's early versions, have we?

docs/Migrating.md
```diff
@@ -9,6 +9,12 @@
   packages, `mlagents.env` and `mlagents.trainers`. `mlagents.env` can be used
   to interact directly with a Unity environment, while `mlagents.trainers`
   contains the classes for training agents.
+* The supported Unity version has changed from `2017.1 or later` to `2017.4.10f1 or later`. The lower versions might still work, but you may encounter error that looks like this:
```
dericp: Do we want 2017.4 here as well?

docs/Migrating.md
```diff
@@ -9,6 +9,12 @@
...
+```
```
dericp: Can specify highlighting with ```console

docs/Migrating.md
```diff
@@ -9,6 +9,12 @@
...
+Instance of CoreBrainInternal couldn't be created. The the script class needs to derive from ScriptableObject.
+UnityEngine.ScriptableObject:CreateInstance(String)
+```
+You can fix the error by removing `CoreBrain` from the CoreBrainInternal.cs:16, click on your Brain Gameobject to let the scene recompile all the changed C# scripts, then add the `CoreBrain` back. (Make sure your brain is in Internal mode, and your TensorFlowSharp plugin is imported and ENABLE_TENSORFLOW flag is set), but this fix is only valid locally and unstable.
```
dericp: Cap line length to 80 characters (minus code samples and things that need to be verbatim).

docs/Migrating.md
dericp: Cap line length to 80 characters (minus code samples and things that need to be verbatim).

docs/Migrating.md
unityjeffrey: i would consider a few changes.  In migrating.md - indicate that we are now following Unity's LTS for support going forward in the future starting with 2017.4.. .  Indicate this helps us maintain good quality and support, etc..  and that future version of ML-agents using Unity's LTS versions will be supported but only in those future versions

 I would move the 2017.1 issue into either FAQ and provide a link to it (i.e. if you are still using older versions, see here for known issues in 0.5).  We should not necessarily call it out here

Lastly, I would add a note in the main read about our moving over to the LTS supported versions of Unity.

xiaomaogy: @unityjeffrey 
1. What does "and that future version of ML-agents using Unity's LTS versions will be supported but only in those future versions" mean?
2. What is the "main read"？

PR #1183, opened 2018-09-06
Three Re-trained Models

awjuliani:
Updated models for the following example environments:
* SoccerTwos
* PushBlock
* Basic


awjuliani:
@xiaomaogy Thanks for catching those changes.

PR #1184, opened 2018-09-06
updated the Pyramids model

xiaomaogy:


awjuliani:
@xiaomaogy Just to confirm, this new model solves the task well?

xiaomaogy:
screen shot 2018-09-05 at 6 28 50 pm

The reward looks like this. The model solves the task, but sometimes a little bit slow. I can let the training last a little bit longer.

PR #1190, opened 2018-09-06
Adding MAPPO algorithm

miguelsuau:
The code still needs to be adjusted to the structure of the new ml-agents version


miguelsuau:
There are two main differences with respect to the original PPO algorithm that affect the trainer_controller class:

1. The function `add_experiences` in MAPPO takes the actions of the other agents as an extra argument. The code snippet below shows how the function call looks like in `trainer_controller.py` :

`if self.trainer_parameters_dict[brain_name]['trainer'] == "mappo":
       trainer.add_experiences(curr_info, new_info, take_action_outputs[brain_name], take_action_vector)
else:
       trainer.add_experiences(curr_info, new_info, take_action_outputs[brain_name])`

2. We need to simulate the actions that the other agents might take when the time horizon is reached. The way I do this is by calling the function `simulate_action` before I process the experiences.

I believe we should find a way to do both things inside the MAPPOTrainer class so that `trainer_controller.py` doesn't need to be modified.

@awjuliani any thoughts?

awjuliani:
@miguelsuau I think it would be fine to extend the `add_experience` and `process_experience` functions to optionally take actions from other agents. 

As for the simulate function, it looks to be the same as take action, except no action is actually taken in the environment, correct? I imagine we can work around that as well. Have you had a chance to look at the new Policy abstraction in v0.5? In this case, we could use a single policy and call `evaluate` on it once for the real action and another time for the simulated actions.

miguelsuau:
I just learned that there is a new way of specifying discrete actions using what you call branches. Before, I would just take the index of each action, convert it into a one hot encoded vector and feed it into the neural network. The problem now (please correct me if I am wrong) is that you can take multiple subactions at the same time. I can see two ways of tackling this.

- Create a huge binary vector that encodes all possible combinations of the subactions.
- Encode each of the branches separately for each agent and feed those to the NN.

Any other ideas?

Sohojoe:
@miguelsuau how does MAPPO improve over PPO?

miguelsuau:
@Sohojoe MAPPO uses a centralized critic. The critic neural network is fed with information about the other agents (actions and observations). The method is similar to the one proposed in this paper https://arxiv.org/pdf/1706.02275.pdf, although the algorithm they use is DDPG instead of PPO.

ScriptBono:
Is there any update on the implementation of the MAPPO algorithm? Would love to see this feature since it creates a whole universe of new possibilities.

unityjeffrey:
hi @ScriptBono - we are revising some of our current approach toward multi-agents (including potentially implementing MAPPO).  we should have an update in the coming weeks or so

chriselion:
Sorry, this got automatically closed when we deleted the `develop` branch, and it looks like I can't reopen or change the target branch

chriselion:

miguelalonsojr:
Closing PR as stale.

PR #1193, opened 2018-09-06
fix the training doc

xiaomaogy:


xiaomaogy:
This is fixing the issue raised by @icaro56 in https://github.com/Unity-Technologies/ml-agents/issues/909

PR #1194, opened 2018-09-06
Update readmes to be more informative

awjuliani:


xiaomaogy:
I know this is not part of this PR, but on ml-agents/README.md: 28, does it make sense to link to 
docs/Python-API.md
docs/Training-ML-Agents.md 

instead of linking it to the main doc page?

Issue #1216, opened 2018-09-12
Imitation Learning has error now in 0.5


icaro56:
Unity Academy name: Academy
        Number of Brains: 2
        Number of External Brains : 1
        Reset Parameters :
		
Unity brain name: EspecialistaOnlyOneAuto
        Number of Visual Observations (per agent): 0
        Vector Observation space size (per agent): 56
        Number of stacked Vector Observation: 1
        Vector Action space type: discrete
        Vector Action space size (per agent): []
        Vector Action descriptions: , , , , , , , , 
Unity brain name: Aprendiz
        Number of Visual Observations (per agent): 0
        Vector Observation space size (per agent): 56
        Number of stacked Vector Observation: 1
        Vector Action space type: discrete
        Vector Action space size (per agent): [9]
        Vector Action descriptions: actions
2018-09-12 12:30:52.561068: I tensorflow/core/platform/cpu_feature_guard.cc:140] Your CPU supports instructions that this TensorFlow binary was not compiled to use: AVX
INFO:mlagents.envs:Hyperparameters for the Imitation Trainer of brain Aprendiz: 
	brain_to_imitate:	EspecialistaOnlyOneAuto
	batch_size:	128
	time_horizon:	128
	graph_scope:	
	summary_freq:	1000
	max_steps:	10000
	batches_per_epoch:	5
	use_recurrent:	False
	hidden_units:	55
	learning_rate:	0.0003
	num_layers:	2
	sequence_length:	64
	memory_size:	256
Process Process-1:
Traceback (most recent call last):
  File "c:\programdata\anaconda3\lib\multiprocessing\process.py", line 258, in _bootstrap
    self.run()
  File "c:\programdata\anaconda3\lib\multiprocessing\process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "c:\programdata\anaconda3\lib\site-packages\mlagents\trainers\learn.py", line 47, in run_training
    tc.start_learning()
  File "c:\programdata\anaconda3\lib\site-packages\mlagents\trainers\trainer_controller.py", line 420, in start_learning
    trainer.update_policy()
  File "c:\programdata\anaconda3\lib\site-packages\mlagents\trainers\bc\trainer.py", line 245, in update_policy
    run_out = self.policy.update(mini_batch, self.n_sequences)
  File "c:\programdata\anaconda3\lib\site-packages\mlagents\trainers\bc\policy.py", line 86, in update
[...]

PR #1220, opened 2018-09-12
Fix branch size discrete actions il samples

icaro56:
#1216 fixed

In HallwayIL and PushBlock scenes, I configured branch_size in student brain.


awjuliani:
Hi @icaro56,

Thanks for catching these fixes. I just realized that unfortunately we won't be able to accept the PR as-is, since you changed the Project version. Please re-commit using version 2017.4. If that is too much work, we can also fix this bug on our end directly. I appreciate you bringing it to our attention in the first place.

icaro56:
@awjuliani , For my research, can I to use Unity Version 2018.2.7f1 or is there some problem?

awjuliani:
@icaro56 Using newer versions for your own use is completely fine. We just want to keep the files on our repository at 2017.4 right now.

PR #1262, opened 2018-09-25
Update unity_env.py

xiaodi-faith:
using multiple observations


awjuliani:
Thanks for making this change!

PR #1265, opened 2018-09-26
Update unity_env.py

xiaodi-faith:
single agent with multiple observations


awjuliani:
Hi @xiaodi-faith,

Thanks for making this PR. Unfortunately however, this changes the behavior of the package in an undesirable way. We want to ensure that the gym wrapper functions as similarly to the atari and mujoco wrappers as possible for maximum compatibility. As such, returning a list of visual observations by default is somewhat I don't believe we would like to do.

PR #1278, opened 2018-10-01
Fix typos

eltronix:


awjuliani:
Thanks for finding these typos and fixing them, @eltronix.

PR #1281, opened 2018-10-01
More semantically correct description

eltronix:
I feel the previous description was misleading.


awjuliani:
Thanks for this one too 😄

PR #1282, opened 2018-10-02
Fix bug in observation vector calculation

bjmolitor:
The calculation of observation vectors is faulty. The old calculation does not reflect distances to the edges and it does not only yield results between -1 and 1. Since distance calculation would have been difficult in one line, I just replaced it by the relative position of the ball (only using two vectors instead of four). I've conducted 500K-step reinforcing trainings before and after the change and got enormously improved results. Contact me for screenshots of the tensorboard or just use the debugger and do the math.

### DIFF
--- a/docs/Learning-Environment-Create-New.md
+++ b/docs/Learning-Environment-Create-New.md
@@ -280,23 +280,20 @@ In our case, the information our Agent collects includes:
   position never changes.
 
 ```csharp
-// Calculate relative position
+// Calculate position relative to the target
 Vector3 relativePosition = Target.position - this.transform.position;
 
-// Relative position
+// Position relative to the target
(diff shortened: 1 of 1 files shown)


bjmolitor:
Red is my best training run with simple reinforcing training out of 8 runs before the fix. Blue is a training run after the fix of the observation function.

2018-10-02 16_10_54-tensorboard

bjmolitor:
Observing the trained internal brain for several minutes, I have not seen the ball fall of platform anymore. If you want to see for yourself, here's a x64 Build of my project with the trained internal brain: [Download](http://bmolitor.de/games/TrainedAI/First_Learning_Enviroment_x64.7z)

awjuliani:
Hi @bjmolitor,

Thanks for making these changes.

PR #1292, opened 2018-10-05
Improve Python-API.md with an example

ucacaxm:
Hi,

First, thank you for you great work in ml-agents !

Since it was not so obvious to create a simple random agent 
with the python API I would like to improve the doc/Python-API.md with an example.
And also fix the import: 
from mlagents.envs.environment import UnityEnvironment

## A simple example of random agent

```python
import sys
import numpy as np

from mlagents.envs.environment import UnityEnvironment


"""The world's simplest agent!"""
class RandomAgent(object):
    def __init__(self, action_space):
        print("Action size="+str(action_space))
        self.action = np.empty((action_space), dtype=float)

    def act(self, observation, reward, done):
        for j in range(self.action.shape[0]):
            self.action[j] = -1.0 + 2.0 * np.random.random_sample()
        return self.action




if __name__ == '__main__':
    env = UnityEnvironment(file_name="3DBall", worker_id=0, seed=1)  # 3DBall executable should exist

    episode_count = 100
    reward = 0
    done = False

    brain = env.brains['BrainRoller']       # See brain name in the Unity Editor
    print("brain="+str(brain))
    print("number of agents=" + str(len(brain.vector_action_space_size)))
    print("action space type="+str(brain.vector_action_space_type))
    print("action space size="+str(brain.vector_action_space_size))
    print("observation space size=" + str(brain.vector_observation_space_size))
    print("=================================")

    info = env.reset()
    agent = RandomAgent(brain.vector_action_space_size)

    agentId = 0            # assume onmy one agent
    for i in range(episode_count):
        info = env.reset()
        gameover = False
        totalReward = 0
        while not gameover:
            brainInfo = info['BrainRoller']
            action = agent.act( brainInfo.vector_observations[agentId], brainInfo.rewards[agentId], brainInfo.local_done[agentId])
            totalReward += brainInfo.rewards[agentId]
            info = env.step(action)
            brainInfo = info['BrainRoller']
            gameover = brainInfo.local_done[agentId]

        print("Game over, total reward=" + str(totalReward))
    env.close()

```

### DIFF
--- a/docs/Python-API.md
+++ b/docs/Python-API.md
@@ -50,7 +50,7 @@ file, put the file in the same directory as `envs`. For example, if the filename
 of your Unity environment is 3DBall.app, in python, run:
 
 ```python
-from mlagents.envs import UnityEnvironment
+from mlagents.envs.environment import UnityEnvironment
 env = UnityEnvironment(file_name="3DBall", worker_id=0, seed=1)
 ```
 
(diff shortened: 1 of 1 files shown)


awjuliani:
Hi @ucacaxm,

Thanks for making this PR. I am somewhat hesitant to add this, since we already have an interactive jupyter notebook dedicated to describing the same process: https://github.com/Unity-Technologies/ml-agents/blob/master/notebooks/getting-started.ipynb. I see that we don't link to it from the `Python-API` page. I will make a point of adding a link to it, which will hopefully make things easier for first-time users.

awjuliani:
I've made a PR with the change here: https://github.com/Unity-Technologies/ml-agents/pull/1311. I think keeping the notebook as the main example is the best way to go, so I will close this PR for now.

ucacaxm:
I did not found the notebook before: it is perfect for this purpose  Yes you are right just linking to the notebook would be better.
Thanks pointing me to the notebook,
Best
Alex
mercredi, 10 octobre 2018, 02:03AM +02:00 de Arthur Juliani  notifications@github.com :

PR #1294, opened 2018-10-06
Update Installation.md

has-taiar:
added a reference to the windows setup docs


xiaomaogy:
Hi @has-taiar , I agree with your change, but there is some minor tweaks on top of what you plan to make. Since I cannot make that change on your branch, I will issue a PR and add you as the contributor on that PR.

xiaomaogy:
@has-taiar I will close this PR because I'm making updates in another PR. Thank you very much for your contribution!

PR #1297, opened 2018-10-07
Update Getting-Started-with-Balance-Ball.md

has-taiar:
Added some instruction on ensuring the solution is loaded in Unity and Play-ing before running the learn python command

### DIFF
--- a/docs/Getting-Started-with-Balance-Ball.md
+++ b/docs/Getting-Started-with-Balance-Ball.md
@@ -213,10 +213,13 @@ When the message _"Start training by pressing the Play button in the Unity
 Editor"_ is displayed on the screen, you can press the :arrow_forward: button in
 Unity to start training in the Editor.
 
+The `--train` flag tells the ML-Agents toolkit to run in training mode.
+
 **Note**: If you're using Anaconda, don't forget to activate the ml-agents
 environment first.
 
-The `--train` flag tells the ML-Agents toolkit to run in training mode.
+**Note**: Before you start training the brain, make sure that you have opened Unity on your computer and laoded the environment from the SDK folder. Running the command above will expect the solution to be laoded and in `Play` mode in Unity.
+
 
 **Note**: You can train using an executable rather than the Editor. To do so,
 follow the intructions in


docs/Getting-Started-with-Balance-Ball.md
```diff
@@ -213,10 +213,13 @@ When the message _"Start training by pressing the Play button in the Unity
...
 Unity to start training in the Editor.
 
+The `--train` flag tells the ML-Agents toolkit to run in training mode.
+
 **Note**: If you're using Anaconda, don't forget to activate the ml-agents
 environment first.
 
-The `--train` flag tells the ML-Agents toolkit to run in training mode.
+**Note**: Before you start training the brain, make sure that you have opened Unity on your computer and laoded the environment from the SDK folder. Running the command above will expect the solution to be laoded and in `Play` mode in Unity.
+
```
mmattar: Delete empty line

docs/Getting-Started-with-Balance-Ball.md
```diff
@@ -213,10 +213,13 @@ When the message _"Start training by pressing the Play button in the Unity
 Editor"_ is displayed on the screen, you can press the :arrow_forward: button in
 Unity to start training in the Editor.
 
+The `--train` flag tells the ML-Agents toolkit to run in training mode.
+
 **Note**: If you're using Anaconda, don't forget to activate the ml-agents
 environment first.
 
-The `--train` flag tells the ML-Agents toolkit to run in training mode.
+**Note**: Before you start training the brain, make sure that you have opened Unity on your computer and laoded the environment from the SDK folder. Running the command above will expect the solution to be laoded and in `Play` mode in Unity.
```
mmattar: Suggested re-wording (including typo fix):

**Note**: Before you start the training session, make sure that the Unity Editor is open and the scene you wish to train is loaded. Running the command above will expect the scene to be active in the Editor and in `Play` mode.

0lento: Am I missing something? if you hit play without training started from mlagents-learn, you'll get an error in Unity when you hit play. Afaik correct order is you launch the training from toolkit first, and only then hit the play button when the command prompt tells you do so. This updated description tells to hit play before starting the training (which shouldn't work at all?) and could cause confusion.

xiaomaogy:
Since this PR is not getting responded by the author, we are closing it. If you are willing to continue to work on this, feel free to reopen it.

PR #1319, opened 2018-10-11
Adds missing instruction in new environment setup

eltronix:


awjuliani:
Hi @eltronix. Thanks for this PR. I don't think we will be accepting it though, since we provide a link to installation instructions in the section right above the one you edited.

eltronix:
Hey @awjuliani, I wish you hadn't closed the PR so quickly without room for discussion.

Nevertheless, I'll go ahead and explain why I suggested the above edit. I've been meticulously reading & implementing instructions in the docs so far and when I came to this page, I was left stranded. It appears the installation instructions you are referring to i.e [this](https://github.com/Unity-Technologies/ml-agents/blob/master/docs/Installation.md) are cognitively inconsistent with the one I suggested [here](https://github.com/Unity-Technologies/ml-agents/blob/master/docs/Basic-Guide.md#setting-up-the-ml-agents-toolkit-within-unity). What I mean is that the later are more complete and will not leave a new user stranded.

Furthermore the first 3 instructions in this section are quite step-wise, in fact they are pretty much reinforcing the said linked instructions above. Therefore I think it would be more consistent to continue with a step-wise approach that does not assume the user will find the final steps elsewhere.

awjuliani:
Hi @eltronix,

I re-opened this, and am assigning @xiaomaogy who has been maintaining the getting started/installation docs. There may indeed be some improvements we can make to ensure the process is as straightforward as possible.

PR #1332, opened 2018-10-14
Fixes agent infinite resetting bug

eltronix:
The check for wether an agent has fallen off the platform was using a wrong value of 1 instead of 0. 
This meant that the agent immediately started in a falling state and entered a thrashing cycle of resetting itself.

### DIFF
--- a/docs/Learning-Environment-Create-New.md
+++ b/docs/Learning-Environment-Create-New.md
@@ -242,10 +242,10 @@ public class RollerAgent : Agent
     public Transform Target;
     public override void AgentReset()
     {
-        if (this.transform.position.y < -1.0)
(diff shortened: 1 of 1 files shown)


awjuliani:
Hi @eltronix,

It looks like the value is actually ` < -1`, which should be correct.

eltronix:
Hey @awjuliani, -1 will definitely work but I've tested it myself and 0 works perfectly too.
The reason is that the sphere representing the agent always has a 0.5 offset from the floor.

awjuliani:
Gotcha. I'll merge this in then.

eltronix:
Hey, just realised I must have been referring to my own flawed local code but thanks for indulging me 😅.

Issue #1403, opened 2018-11-06
Contribution: Recording videos during training


LeSphax:
Hello,

I have recently started to record videos of my game automatically during training using this code https://github.com/unity3d-jp/FrameCapturer.

I find it quite practical for debugging and I could take the time to create a pull request if you think this is something that could be part of ml-agents.

xiaomaogy:
Hi @LeSphax, thank you very much! It would definitely help if you could create a PR.

xiaomaogy:
I'm closing this since we already have merged a PR.

PR #1414, opened 2018-11-11
Add a script to record videos every X seconds or academy steps

LeSphax:
So this is the contribution I briefly mentioned in issue  #1403 
This code uses the frame capturer repository https://github.com/unity3d-jp/FrameCapturer.

So to record videos:

1. Create a camera
2. Add the RecordVideos script to it
3. Choose the recording options on the RecordVideos script.

The frame capturer code was put in Scripts/FrameCapturer. Some parts of the code were removed/changed:

- Removed the examples
- Removed the AudioCapturer and GBufferRecorder
- Changed MovieRecorderEditor to hide all modifiable variables from the user.
- Removed BeginRecording and EndRecording logging from Movie Recorder
- Made it possible to specify the video name in the BeginRecording function

I tried to change as little as possible while removing unused code and making the interface simpler.

The code I actually created:

1. The RecordVideos and RecordVideosEditor scripts
2. A new function in the Academy to get the TotalStepCount

I can also create documentation for it but first I would like some feedback to make sure the current interface and the options make sense.

Testing:
- I tested this on Windows using the 3DBall tutorial, in the editor and with an executable running for 15 minutes.
Mostly made sure that the videos were indeed recording and that the agents were still learning.
- I also used the framework on my own project for 8-15 hours sessions.

### DIFF
--- a/UnitySDK/Assets/ML-Agents/Scripts/Academy.cs
+++ b/UnitySDK/Assets/ML-Agents/Scripts/Academy.cs
@@ -469,14 +469,25 @@ namespace MLAgents
         }
 
         /// <summary>
-        /// Returns the current step counter (within the current epside).
+        /// Returns the current step counter (within the current episode).
         /// </summary>
         /// <returns>
-        /// Current episode number.
+        /// Current step count.
         /// </returns>
         public int GetStepCount()
         {
             return stepCount;
+        }
+
+        /// <summary>
+        /// Returns the total step counter.
+        /// </summary>
+        /// <returns>
+        /// Total step count.
+        /// </returns>
+        public int GetTotalStepCount()
+        {
+            return GetStepCount() + GetEpisodeCount() * maxSteps;
         }
 
         /// <summary>
--- a/UnitySDK/Assets/ML-Agents/Scripts/Academy.cs
+++ b/UnitySDK/Assets/ML-Agents/Scripts/Academy.cs
@@ -169,6 +169,10 @@ namespace MLAgents
         /// <see cref="AcademyReset"/>.
         int stepCount;
 
+        /// The number of total number of steps completed during the whole simulation. Incremented
+        /// each time a step is taken in the environment.
+        int totalStepCount;
+
         /// Flag that indicates whether the inference/training mode of the
         /// environment was switched by the external Brain. This impacts the
         /// engine settings at the next environment step.
@@ -487,7 +491,7 @@ namespace MLAgents
         /// </returns>
         public int GetTotalStepCount()
         {
-            return GetStepCount() + GetEpisodeCount() * maxSteps;
+            return totalStepCount;
         }
 
         /// <summary>
@@ -600,6 +604,7 @@ namespace MLAgents
             AgentAct();
 
             stepCount += 1;
+            totalStepCount += 1;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


LeSphax:
Hello @xiaomaogy,

I have created the pull request, should I also tag someone to review it?

I also noticed that the automatic Quality Review fails, two questions about that:

1. Most of the errors come from the code I imported from  https://github.com/unity3d-jp/FrameCapturer, I was planning to change the code as little as possible in case the repo gets updated and we want to update it on our side as well. Do that seem like a good idea? Or should I remove all the code we don't use and change it to follow the code standards?

2. Getting the code review only after pushing to git is not very practical. Any way I can get the same warnings on visual studio/ visual studio code?

xiaomaogy:
@vincentpierre @awjuliani

LeSphax:
Oki doki, thanks for the update :)

LeSphax:
After discussing with @unityjeffrey all the video recording code was moved to the [videoRecorder repository](https://github.com/Unity-Technologies/video-recorder).
To test the feature, refer to the ReadMe of the videoRecorder repository.

The pull request now only contains the function I added to the academy and a change to the .gitignore file to ignore the plugin.

UnitySDK/Assets/ML-Agents/Scripts/Academy.cs
```diff
@@ -469,16 +469,27 @@ public int GetEpisodeCount()
...
 
+        /// <summary>
+        /// Returns the total step counter.
+        /// </summary>
+        /// <returns>
+        /// Total step count.
+        /// </returns>
+        public int GetTotalStepCount()
+        {
+            return GetStepCount() + GetEpisodeCount() * maxSteps;
```
harperj: Looks like this makes the assumption that every episode reaches `maxSteps`.  In reality we don't have any current account of how many steps each historical episode lasted.  We'll have to store total step count for this method to work correctly.

LeSphax: I am not sure that this is correct, the variable names are actually confusing here. I think the maxSteps here is the number of steps between academy resets and the episode count is the number of academy episodes not agent episodes.

The academy should only reset when it reaches its max step right?

But after checking it seems that there is another problem anyway. It looks like the academy is resetting twice every time and the episode count is also incremented twice.

So in the end, I added a variable for totalStepCount, it makes it easier to understand the code anyway.

It should be fixed in b0d1921 👍

harperj: I think that's mostly correct except that you might reset the Academy early (for example by explicitly calling the Gym wrapper's `reset()`).  Anyway, this change looks good to me :)

harperj:
So these changes specifically look good, though I do also have some commentary about the larger repo (https://github.com/Unity-Technologies/video-recorder).

The project seems to work as expected for our example environments at this time, but the FrameCapturer library it's based on currently isn't officially supported and doesn't seem to work with newer Unity features like the Lightweight Render Pipeline (https://docs.unity3d.com/Packages/com.unity.render-pipelines.lightweight@4.0/).

Before we update the documentation to suggest others use the project I think we'll want to further discuss as a team.  Either way, I think that this is a reasonable atomic change to the C# API.

harperj:
Hi @LeSphax,

I've merged the PR.  As I mentioned, the video-recorder package has limited support for the latest Unity features.  We provide the [Unity Recorder](https://assetstore.unity.com/packages/essentials/unity-recorder-94079) for recording videos as well (which has better support) but it only allows Editor recording.

I'd like to share the video recorder with the community in a way that clarifies it won't work with the Scriptable Render Pipeline.

Thanks -- H

PR #1459, opened 2018-12-04
Fixes to container workflow for GCP

eshvk:
Fix to handle running our containers on VMs (or bare machines) that have file systems with `noexec` flag set, e.g. [GCP Container OS](https://stackoverflow.com/questions/49037720/cannot-run-executable-shell-script-on-google-container-optimized-os). This copies the Unity executable inside the container before executing.

### DIFF
--- a/ml-agents/mlagents/trainers/trainer_controller.py
+++ b/ml-agents/mlagents/trainers/trainer_controller.py
@@ -3,7 +3,9 @@
 """Launches trainers for each External Brains in a Unity Environment."""
 
 import os
+import glob
 import logging
+import shutil
 
 import yaml
 import re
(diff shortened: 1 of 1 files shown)


ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -71,8 +73,31 @@ def __init__(self, env_path, run_id, save_freq, curriculum_folder,
...
-                    docker_target_name=docker_target_name, env_name=env_path)
+                """
+                Comments for future maintenance:
+                    Some OS/VM instances (e.g. COS GCP Image) mount filesystems 
+                    with COS flag which prevents execution of the Unity scene, 
+                    to get around this, we will copy the executable into the 
+                    container.
+                """
+                # Navigate in docker path and find env_path and copy it.
+                for f in glob.glob('/{docker_target_name}/*'.format(
```
xiaomaogy: Also do you always need to do this copying? Or should you first check if you are on a instance that has "noexec" flag, and if true you do this copying.

eshvk: It  is not harmful to do the copying. I think checking within the code would make the code a bit uglier.

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -71,8 +73,31 @@ def __init__(self, env_path, run_id, save_freq, curriculum_folder,
...
+                        docker_target_name=docker_target_name)):
+                    if env_path in f:
+                        try:
+                            b = os.path.basename(f)
+                            if os.path.isdir(f):
+                                shutil.copytree(f,
+                                                '/ml-agents/{b}'.format(b=b))
```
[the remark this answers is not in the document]
eshvk: Correct, b/c this is the write access within the container not the underlying system.

ml-agents/mlagents/trainers/trainer_controller.py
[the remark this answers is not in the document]
eshvk: Fixed

xiaomaogy: Make sense.

Issue #1448, opened 2018-11-27
gym-unity: Int64 isInstance(int) not successful on Windows


theDummbatz:
I am currently trying to run the gym-unity wrapper on Windows.
As mentioned before, the Readme is not up to date, and DQN Baseline example is not working, but this can be easily fixed by using the current method defintion for deepq.learn(). 

While running the GridWorld example, i encountered a different error:
```
Traceback (most recent call last):
  File "C:/Users/ex/Documents/ml-agents/train_agent.py", line 30, in <module>
    main()
  File "C:/Users/ex/Documents/ml-agents/train_agent.py", line 23, in main
    gamma=0.99
  File "C:\Users\ex\AppData\Local\Programs\Python\Python36\lib\site-packages\baselines\deepq\deepq.py", line 282, in learn
    new_obs, rew, done, _ = env.step(env_action)
  File "C:\Users\ex\AppData\Local\Programs\Python\Python36\lib\site-packages\gym_unity\envs\unity_env.py", line 131, in step
    info = self._env.step(action)[self.brain_name]
  File "C:\Users\ex\AppData\Local\Programs\Python\Python36\lib\site-packages\mlagents\envs\environment.py", line 327, in step
    for brain_name in list(vector_action.keys()) + list(memory.keys()) + list(text_action.keys()):
AttributeError: 'numpy.int64' object has no attribute 'keys' 
```

While debugging, i found the reason to be a failing type check in gym_unity\envs\environment.py, line 281: 

` if isinstance(vector_action, (int, np.int_, float, np.float_, list, np.ndarray)):`

`vector_action` is an int64 datatype and on Windows np.int_ evaluates to a standard int32 datytype. 
Similar typechecks are also at different spots in the code.
A hotfix for me was to cast the value in \baselines\deepq\deepq.py, line 280:
```
            if isinstance(action, np.int64):
                env_action = np.int32(action)
            else:
                env_action = action
```
but this obviously is not a real solution. Does the action value have to be an `int64` integer? Has anyone else encountered a similar problem?
I might add that normal baseline examples (e.g. the CartPole training) work fine.

System Specs: 
- MS Windows 7 Enterprise 64-Bit
- Python 3.6.7
- Numpy 1.14
- Tensorflow 1.7.1 CPU-Version

PR #1471, opened 2018-12-10
Support both 32-bit and 64-bit types in UnityEnvironment

harperj:
We check for the single brain case in UnityEnvironment by checking
for applicable non-dict types in the step arguments. However for ints
and floats we just use `np.int_` and `np.float_` for the check, which
are the defaults for your system.

This means if you are using an application (like baselines in #1448)
which uses the wrong int/float size an error will be thrown.  This
change explicitly allows both 32 and 64-bit numbers.

### DIFF
--- a/ml-agents/mlagents/envs/environment.py
+++ b/ml-agents/mlagents/envs/environment.py
@@ -20,6 +20,11 @@ logger = logging.getLogger("mlagents.envs")
 
 
 class UnityEnvironment(object):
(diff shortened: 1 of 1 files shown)


harperj:
I rebased this over the `release-v0.6` branch and fixed a downstream issue caused by int64 / float64 values in `_flatten`.  I've confirmed that baselines DQN now works on Windows.

xiaomaogy:
Hi @harperj If you've tested it feel free to merge.

PR #1520, opened 2018-12-20
Fix for divide-by-zero error with Discrete Actions 

ervteng:
Zero-padding for action masks in the AgentBufferField can cause a divide-by-zero when used by the model to mask actions. 

Fixed by enabling padding to be set to something other than 0, and setting padding for just action_masks to 1.

### DIFF
--- a/ml-agents/mlagents/trainers/buffer.py
+++ b/ml-agents/mlagents/trainers/buffer.py
@@ -28,12 +28,27 @@ class Buffer(dict):
             AgentBufferField with the append method.
             """
 
+            def __init__(self):
+                self.padding_value = 0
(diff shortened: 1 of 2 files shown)


ervteng:
It's still running, but the reward has climbed to ~0.5.

awjuliani:
@ervteng Okay. Once it gets >0.7, then it is good to go.

ervteng:
@awjuliani It's at 0.9 now after 220000 steps.

awjuliani:
@ervteng Sounds good then! 🚢 🇮🇹 !

ervteng:
Tested continuous and multidiscrete environments (with LSTM) as well with this fix. Looks good.

PR #1595, opened 2019-01-12
API for sending custom protobuf messages to and from Unity.

malmaud:
Implements https://github.com/Unity-Technologies/ml-agents/issues/1549.

This allows users to customize two protobuffer messages:

*  They can add fields to `CustomParameters` and pass an instance of their custom parameters to `UnityEnvironment.reset` with the new `custom_config` keyword argument. This is useful if the scene configuration is complex and hard to encode as a few scalar parameters, as the academy is currently set up to support.

* They can also add fields to `CustomOutput` message and set the value of those fields in `AgentAction`, which will in turn be available in Python as part of the agent info returned from `UnityEnvironment.step`. They call `GetCustomOutput` in `AgentAction` or `CollectObservations` to get an instance of the `CustomOutput` object and set the value of its fields.

This enables Unity to return more than a single scalar reward to Python, which is useful for eg sending back the complex effects of an action and having the reward calculation logic happen in Python, or for debugging or computing diagnostics in Python.

I'm happy to write up documentation once (if) an API is agreed upon.


UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -541,6 +548,7 @@ void ResetData()
                           * param.numStackedVectorObservations]);
 
             info.visualObservations = new List<Texture2D>();
+            info.customOutput = new CustomOutput();
```
vincentpierre: It is not clear to me how the user is supposed to pass a `customOutput` to the agentInfo. Here you create a new one and the  `GetCustomOutput()` gives you the customOutput in the AgentInfo. I would rather have at this line `info.customOutput = this.customOutput` and have getter and setter on `Agent.customOuptut`.

ml-agents/mlagents/envs/brain.py
```diff
@@ -26,6 +26,7 @@ def __init__(self, visual_observation, vector_observation, text_observations, me
         self.previous_vector_actions = vector_action
         self.previous_text_actions = text_action
         self.action_masks = action_mask
+        self.custom_output =  custom_output
```
vincentpierre: I think this should be `self.custom_outputs`

protobuf-definitions/proto/mlagents/envs/communicator_objects/agent_info_proto.proto
```diff
@@ -3,6 +3,9 @@ syntax = "proto3";
 option csharp_namespace = "MLAgents.CommunicatorObjects";
 package communicator_objects;
 
+message CustomOutput {
```
vincentpierre: I think both CustomOutput and CustomParameters should live in separate files. Since these are supposed to be modified by the user, they should be separated from the rest of the proto definitions.

ml-agents/mlagents/envs/brain.py
malmaud: I definitely see the logic of that, but I wonder if it will be confusing since `custom_outputs` is a scalar and rather than a repeated field.

protobuf-definitions/proto/mlagents/envs/communicator_objects/agent_info_proto.proto
malmaud: 👍

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
malmaud: My idea was the user does something like

```cs
void AgentAction(...) {
  var myOutput = GetCustomOutput();
  myOutput.x = 5;
}
```

This was to get around the user having to create their own instance of `CustomOutput` , since then the example becomes

```cs
void AgentAction(...) {
  var myOutput = new CommunicatorObjects.CustomOutput();
  myOutput.x = 5;
  customOutput = myOutput;
```

That seems rather ugly, although it become betters if `CustomOutput` is imported into `MLAgents` namespace.

Are you proposing simply

```cs
void AgentAction(...) {
  customOutput.x = 5;
}
```

with `customOutput` a getter on an internal  attribute that is still created automatically by `Agent` instead of by the user?

ml-agents/mlagents/envs/brain.py
vincentpierre: I do not understand this. Shouldn't `custom_output` be a list of custom_output ? If there are `N` agents in the environment each sending a custom_output at the same step, the BrainInfo should contain all of them.

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
vincentpierre: I would like to avoid creating a new instance of CustomOutput at every step when it is not needed. I am not sure what the API should look like here. To be consistent with AddObservation, we could have a AddCustomOutput method. 
Now that I think about it, having an Agent.customOutput might be confusing. The user might try to modify this value from another script.

ml-agents/mlagents/envs/brain.py
malmaud: Ah you're right; I misinterpreted your comment since it is scalar protobuf message attached to agent but is indeed list of those message on the `BrainInfo` instance and so the name change makes sense.

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -80,6 +80,12 @@ public struct AgentInfo
         /// </summary>
         public int id;
 
+        /// <summary>
+        /// User-customizable object for sending structured output from Unity to Python in response
+        /// to an action in addition to a scalar reward.
+        /// </summary>
+        public CustomOutput customOutput;
```
vincentpierre: I would need to investigate but I am not sure this is ever used.

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -1083,5 +1091,23 @@ public static void ObservationToTexture(Camera obsCamera, int width, int height,
...
+        public void SetCustomOutput(CustomOutput customOutput)
+        {
+            info.customOutput = customOutput;
+        }
+
+        /// <summary>
+        /// Creates a new instance of a custom output object.
+        /// </summary>
+        /// <returns>A reference to a new CustomOutput object.</returns>
+        static public CustomOutput CreateCustomOutput()
```
vincentpierre: I think it is fine to let the developer instantiate his own CustomOutput. This method feels redundant

malmaud:
I've very grateful for the team's support. Thank you!

I'll make all those requests changes in the next few days. I do have a use case since this was motivated by one of my PhD projects. I can boil that down to the minimum to show off the API and post a link for discussion.

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
malmaud: My thinking was that could be problematic since the user has to then know about the `CommunicatorObjects` namespace, where the `CustomOutput` class lives. If the C# namespace of the custom_output.proto file ever changes due to some kind of internal refactoring, users would have to adopt their code.

But sure, I'll make this change.

malmaud:
WRT to renaming `CustomOutput` to `CustomObservation`, I realize there are two places you may want custom objects outputted from an agent and I had been conflating them:

* In `CollectObservations`, so the external brain can choose an action based on a custom structure that summarizes the relevant observations. Here the name `CustomObservation` makes total sense.

* In `AgentAction`, so an external brain can use a custom summary of the result of an action to compute rewards. This is nice if you want the reward computation to be split between Unity and Python - eg, you use Unity's physics to compute the positions of objects in response to an action, but you prefer the reward calculation based on those positions to happen in your Python script.

I was thinking of adding a new custom message to `AgentInfoProto` for this case and calling it something like `CustomActionResult`.

vincentpierre:
Unfortunately, there is no way to send data from Unity to Python at the AgentAction level. The second scenario you are proposing would require information to be sent from Unity to Python twice per step (If I understood correctly). I do not think the second scenario you mentioned would fit in the ML-Agents toolkit.

vincentpierre:
I am unclear regarding what `CustomActionResult` is and what it does. It seems to me to be a duplicate of `CustomObservation` (The two are always next to each other) so I wonder what it adds.

malmaud:
Here was my thinking. I'm totally happy to just get rid of `CustomActionResult` though if this doesn't make sense to you.

The basic flow I imagined was:

1) Python sends Unity an action for an agent with `step`.
2) `AgentAction` is called on the user's agent. It creates a `CustomActionResult` that includes metadata about the reward or sufficient information about the environment to calculate reward.
3) `CollectObservations` is called. It creates a `CustomObservation`. 
4) The custom action result and custom observation are sent back to Python as part of the return value of `step` . Python calculates a reward from the custom action result and then chooses a new action based on the custom observation object.

In principle, the Unity agent could just store whatever information it was going to put in the custom action result during `AgentAction` into the custom observation instance instead. I just thought that it might be a simpler interface to separate them since their role is often different - one contains information relevant to calculating a reward in response to an action and the other is relevant to choosing the next action based on properties of the environment.

vincentpierre:
I see what you mean, `CustomObservation` is not only an observation, it is an output (hence why it was named `CustomOutput`). It could be `CustomObservationAndReward` but I think this will be too confusing. I think as a first step we should only include `CustomObservations`. Users will still be able to have custom results for their actions if they integrate them into the `CustomObservation`. The Capabilities would be the same, it might just feel a little hacky to put a reward (or similar) inside the observations. 
Thank you.

malmaud:
Alright. How about I add a getter for the `CustomObservation` too, to make it easier for the user to set fields on it from both `CollectObservations` and `AgentAction`?

vincentpierre:
I think the PR looks good. I think we would need to make sure pytest still work, maybe even make new unit tests. The API looks good, I think you can start making documentation update for it. Let us know if you need help.

malmaud:
Ok, great. It will be a week or so until I can get back to this.

vincentpierre:
@malmaud Did you have time to revisit this PR ?

malmaud:
I can work on it later this week.

malmaud:
OK, ready to work on this again. So it sounds like what's still needed is docs and tests. 

Looks like for docs, I'd add sections to the `docs` folder. 

Can you guide me a bit on how to get started with adding tests? It looks like `ml-agents/tests` is the directory I'd want to be checking out.

vincentpierre:
Yes, all the documentation is present in the `docs` folder. Since this is not a change to the trainers, I would say the tests will go into the `envs` tests. Not sure what tests we should do because the custom proto objects are empty, so hard to assign values to them. I would say what replaces the reset parameters can be tested at least.

malmaud:
OK, I added some documentation. It's primarily in a new `Custom-Protos.md` file in `docs`. Feel free to directly push edits to it on my branch if there are small tweaks you'd like to see.

I implemented all the examples I mention in that file to double-check that they work.

Still have to work on tests.

docs/Custom-Protos.md
```diff
@@ -0,0 +1,167 @@
+# Creating custom protobuffer messages
+
+Unity and Python communicate by sending protobuffer messages to and from each other. You can create custom protobuffer messages if you want to exchange structured data beyond is included by default.
```
vincentpierre: beyond what is included by default

docs/Custom-Protos.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Custom message types
+
+There are three custom message types currently supported, described below. In each case, `env` is an instance of a `UnityEnvironment` in Python. `CustomAction` is described most thoroughly; usage of the  other custom messages follows a similar template. 
+
+### Custom actions
+
+By default, the Python API sends actions to Unity in the form of a floating-point list per agent and an optional string-valued text action. 
+
+You can define a custom action type to replace or augment this by adding fields to the `CustomAction` message, which you can do by editing the file "protobuf-definitions/proto/mlagents/envs/communicator_objects/custom_action.proto".
```
vincentpierre: Quote the file path with ` <-- this

docs/Custom-Protos.md
```diff
@@ -0,0 +1,167 @@
...
+env.reset(custom_reset_parameters=params)
+```
+
+### Custom observations
+
+By default, Unity returns observations to Python in the form of a floating-point vector. 
+
+You can define a custom observation message to supplement that. To do so, add fields to the `CustomObservation` protobuffer message in `custom_observation.proto`. 
+
+Then in your agent, create an instance of a custom observation via `new CommunicatorObjects.CustomObservation`. Then in `CollectObservations`, call `SetCustomObservation` wit the custom observation instance as the parameter.
```
vincentpierre: wit --> with

docs/Custom-Protos.md
```diff
@@ -0,0 +1,167 @@
...
+
+In Python, the custom observation can be accessed in the return of `env.step` as ` custom_observations` property (one entry per agent).
```
vincentpierre: > can be accessed in the return of `env.step`
Can you say something like is accessed by calling `env.step` or `env.reset`

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -1083,5 +1120,23 @@ public static void ObservationToTexture(Camera obsCamera, int width, int height,
...
+        public void SetCustomObservation(CustomObservation customObservation)
+        {
+            info.customObservation = customObservation;
+        }
+
+        /// <summary>
+        /// Get the custom observation that has previously been set for this agent during this episode.
+        /// </summary>
+        /// <returns> An instance of a CustomObservation message.</returns>
+        public CustomObservation GetCustomObservation()
```
vincentpierre: I am not sure this method is useful at all. I would remove it.

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -80,6 +80,12 @@ public struct AgentInfo
         /// </summary>
         public int id;
 
+        /// <summary>
+        /// User-customizable object for sending structured output from Unity to Python in response
+        /// to an action in addition to a scalar reward.
+        /// </summary>
+        public CustomObservation customObservation;
```
vincentpierre: Why does this field have to be public ?

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -826,6 +835,25 @@ public virtual void AgentAction(float[] vectorAction, string textAction)
...
+        /// <param name="vectorAction">
+        /// Vector action. Note that for discrete actions, the provided array
+        /// will be of length 1.
+        /// </param>
+        /// <param name="textAction">Text action.</param>
+        /// <param name="customAction">
+        /// A custom action, defined by the user as custom protobuffer message. Useful if the action is hard to encode
+        /// as either a flat vector or a single string.
+        /// </param>
+        public virtual void AgentAction(float[] vectorAction, string textAction, CommunicatorObjects.CustomAction customAction)
```
vincentpierre: This will change our demo environments code. We will need to update them before merging

UnitySDK/Assets/ML-Agents/Scripts/CommunicatorObjects/UnityToExternalGrpc.cs
```diff
@@ -2,7 +2,7 @@
 //     Generated by the protocol buffer compiler.  DO NOT EDIT!
 //     source: mlagents/envs/communicator_objects/unity_to_external.proto
 // </auto-generated>
-#pragma warning disable 1591
```
vincentpierre: Something very fishy is happening right now. On the `develop` branch, there is a line here saying : 

```csharp
# if UNITY_EDITOR || UNITY_STANDALONE_WIN || UNITY_STANDALONE_OSX || UNITY_STANDALONE_LINUX
```

This line is very important and must be included. But I am most worried by the fact that github does not pick it up. There might be other places like this in the PR where code is reverted without github noticing and this is **very** problematic.

vincentpierre: Update on this, it seems to be because the whole file is conflicting. Maybe this is the reason github does not show the diff here.

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
malmaud: I made it public since all the other properties in `AgentInfo`, where this declaration resides, are public. `Agent` directly accesses the field in `SetCustomObservation` and `ResetData`. The `AgentInfo` instance attached to the agent is still private though to the agent, though.

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
malmaud: Done

docs/Custom-Protos.md
malmaud: done

docs/Custom-Protos.md
malmaud: done

docs/Custom-Protos.md
malmaud: done

docs/Custom-Protos.md
malmaud: done

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
vincentpierre: Actually no because of L854

docs/Custom-Protos.md
```diff
@@ -0,0 +1,167 @@
+# Creating custom protobuffer messages
```
awjuliani: These are `protobuf` messages, not `protobuffer` messages.

docs/Custom-Protos.md
```diff
@@ -0,0 +1,167 @@
...
+
+Unity and Python communicate by sending protobuffer messages to and from each other. You can create custom protobuffer messages if you want to exchange structured data beyond what is included by default.
```
awjuliani: `protobuf`

docs/Custom-Protos.md
```diff
@@ -0,0 +1,167 @@
...
+```
+
+In your Python file, create an instance of a custom action:
+
+```python
+from mlagents.envs.communicator_objects import CustomAction
+env = mlagents.UnityEnvironment(...)
+...
+action = CustomAction(direction=CustomAction.NORTH, walkAmount=2.0)
+env.step(custom_action=action)
```
awjuliani: Don't we need to also provide a regular action along with the custom one?

docs/Readme.md
```diff
@@ -56,3 +56,4 @@
 * [API Reference](API-Reference.md)
 * [How to use the Python API](Python-API.md)
 * [Wrapping Learning Environment as a Gym](../gym-unity/README.md)
+* [Creating custom protobuffer messages](Custom-Protos.md)
```
awjuliani: `protobuf`

docs/Custom-Protos.md
```diff
@@ -0,0 +1,167 @@
+# Creating custom protobuffer messages
+
+Unity and Python communicate by sending protobuffer messages to and from each other. You can create custom protobuffer messages if you want to exchange structured data beyond what is included by default. 
+
+Whenever you change the fields of a custom message, you must run `protobuf-definitions/make.bat` to create C# and Python files corresponding to the new message. Follow the directions in that file for guidance. After running it, reinstall the Python package by running `pip install ml-agents` and make sure your Unity project is using the newly-generated version of `UnitySDK`.
```
awjuliani: Won't `pip install ml-agents` install the pypi version, not the version with the modified protobuf definitions?

docs/Custom-Protos.md
vincentpierre: You don't, the default is a zero array.

docs/Custom-Protos.md
```diff
@@ -0,0 +1,167 @@
...
+
+```protobuf
+syntax = "proto3";
+
+option csharp_namespace = "MLAgents.CommunicatorObjects";
+package communicator_objects;
+
+message CustomAction {
+    enum Direction {
+        NORTH=0;
```
ervteng: These values should be different, otherwise this proto won't compile.

docs/Custom-Protos.md
ervteng: `make.bat` is missing two key pieces of information to get this to work:

- You have to use the right version of Grpc.Tools (1.14.1) when installing with nuget. Otherwise the .cs script will be generated but it will not compile in Unity. 
- You have to have `protobuf` installed (e.g. via brew or apt) before running the script. This one is pretty obvious. 
-

docs/Custom-Protos.md
```diff
@@ -0,0 +1,167 @@
...
+    float walkAmount = 1;    
+    Direction direction = 2;
+}
+```
+
+In your Python file, create an instance of a custom action:
+
+```python
+from mlagents.envs.communicator_objects import CustomAction
+env = mlagents.UnityEnvironment(...)
```
ervteng: Should be `mlagents.envs.UnityEnvironment(...)` or `mlagents_envs.UnityEnvironment`.

docs/Custom-Protos.md
malmaud: OK, I changed `protobuffer->protobuf` in several bunches.

docs/Custom-Protos.md
malmaud: done

docs/Readme.md
malmaud: done

docs/Custom-Protos.md
malmaud: I've tried to clarify that you are meant to give the path to the local `ml-agents` folder, since if `pip` sees it's been given a valid local path it will try to install that before searching pypi.

docs/Custom-Protos.md
malmaud: Good catch, copy/paste error from a working example.

docs/Custom-Protos.md
malmaud: Done

docs/Custom-Protos.md
malmaud: That is true. Since that's just as true with or without this PR though, should adding that information be a separate PR?

ervteng: I think we should add it to this PR since normally there is no reason to use make.bat - but now make.bat is a user-facing interface. @vincentpierre, what do you think?

malmaud:
Thanks everyone!

PR #1669, opened 2019-02-02
Move 'take_action' into Policy class

harperj:
This refactor is part of Actor-Trainer separation. Since policies
will be distributed across actors in separate processes which share
a single trainer, taking an action should be the responsibility of
the policy.

This change makes a few smaller changes:
* Combines `take_action` logic between trainers, making it more
  generic
* Adds an `ActionInfo` data class to be more explicit about the
  data returned by the policy, only used by TrainerController and
  policy for now.
* Moves trainer stats logic out of `take_action` and into
  `add_experiences`
* Fixes a minor issue where `reward_buffer` was being referred to
  on generic Trainers during curriculum learning even though this
  property doesn't exist on the base Trainer class.

### DIFF
--- a/ml-agents/mlagents/trainers/action_info.py
+++ b/ml-agents/mlagents/trainers/action_info.py
@@ -1,0 +1,9 @@
+from typing import NamedTuple, Any, Dict, Optional
+
+
+class ActionInfo(NamedTuple):
+    action: Any
+    memory: Any
+    text: Any
+    value: Any
+    outputs: Optional[Dict[str, Any]]
--- a/ml-agents/mlagents/trainers/policy.py
+++ b/ml-agents/mlagents/trainers/policy.py
@@ -2,9 +2,10 @@ import logging
 import numpy as np
 import tensorflow as tf
 
-from mlagents.trainers import UnityException
+from mlagents.trainers import ActionInfo, UnityException
 from tensorflow.python.tools import freeze_graph
 from mlagents.trainers import tensorflow_to_barracuda as tf2bc
+from mlagents.envs import BrainInfo
 
 logger = logging.getLogger("mlagents.trainers")
 
@@ -79,13 +80,32 @@ class Policy(object):
                             .format(self.model_path))
             self.saver.restore(self.sess, ckpt.model_checkpoint_path)
 
-    def evaluate(self, brain_info):
+    def evaluate(self, brain_info: BrainInfo):
         """
         Evaluates policy for the agent experiences provided.
         :param brain_info: BrainInfo input to network.
         :return: Output from policy based on self.inference_dict.
         """
         raise UnityPolicyException("The evaluate function was not implemented.")
+
+    def get_action(self, brain_info: BrainInfo) -> ActionInfo:
+        """
+        Decides actions given observations information, and takes them in environment.
+        :param brain_info: A dictionary of brain names and BrainInfo from environment.
+        :return: an ActionInfo containing action, memories, values and an object
+        to be passed to add experiences
+        """
+        if len(brain_info.agents) == 0:
+            return ActionInfo([], [], [], None, None)
+
+        run_out = self.evaluate(brain_info)
+        return ActionInfo(
+            action=run_out.get('action'),
+            memory=run_out.get('memory_out'),
+            text=None,
+            value=run_out.get('value'),
+            outputs=run_out
+        )
 
     def update(self, mini_batch, num_sequences):
         """
--- a/ml-agents/mlagents/trainers/ppo/trainer.py
+++ b/ml-agents/mlagents/trainers/ppo/trainer.py
@@ -109,27 +109,6 @@ class PPOTrainer(Trainer):
         self.policy.increment_step()
         self.step = self.policy.get_current_step()
 
-    def take_action(self, all_brain_info: AllBrainInfo):
-        """
-        Decides actions given observations information, and takes them in environment.
-        :param all_brain_info: A dictionary of brain names and BrainInfo from environment.
-        :return: a tuple containing action, memories, values and an object
-        to be passed to add experiences
-        """
-        curr_brain_info = all_brain_info[self.brain_name]
-        if len(curr_brain_info.agents) == 0:
-            return [], [], [], None, None
-
-        run_out = self.policy.evaluate(curr_brain_info)
-        self.stats['Policy/Value Estimate'].append(run_out['value'].mean())
-        self.stats['Policy/Entropy'].append(run_out['entropy'].mean())
-        self.stats['Policy/Learning Rate'].append(run_out['learning_rate'])
-        if self.policy.use_recurrent:
-            return run_out['action'], run_out['memory_out'], None, \
-                   run_out['value'], run_out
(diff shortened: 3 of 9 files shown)


ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -193,10 +194,10 @@ def start_learning(self, env, trainer_config):
 
         # Prevent a single session from taking all GPU memory.
         self.initialize_trainers(trainer_config)
+        policies = {brain_name: trainer.policy for brain_name, trainer in self.trainers.items()}
```
harperj: @vincentpierre you mentioned some concerns about this, could you elaborate?  @eshvk is reviewer on this PR but wanted to make sure to get your input

ml-agents/mlagents/trainers/policy.py
```diff
@@ -79,14 +80,33 @@ def _load_graph(self):
...
-    def evaluate(self, brain_info):
+    def evaluate(self, brain_info: BrainInfo):
         """
         Evaluates policy for the agent experiences provided.
         :param brain_info: BrainInfo input to network.
         :return: Output from policy based on self.inference_dict.
         """
         raise UnityPolicyException("The evaluate function was not implemented.")
 
+    def take_action(self, brain_info: BrainInfo) -> ActionInfo:
```
eshvk: Coming from the perspective of the `Policy` class as implementing a `policy` mathematical function, I think it makes sense to `take_action` here. It reminds me of say a key-value store where you feed in observations and get back actions.  Going from that perspective, what are your thoughts on renaming `take_action` to something like `get_action`? This way it is clear that this is very distinct in behavior from say `take_step` which exists in the same code.

ml-agents/mlagents/trainers/ppo/trainer.py
```diff
@@ -181,6 +160,12 @@ def add_experiences(self, curr_all_info: AllBrainInfo, next_all_info: AllBrainIn
         :param next_all_info: Dictionary of all current brains and corresponding BrainInfo.
         :param take_action_outputs: The outputs of the take action method.
```
eshvk: Fix comment to make it explicit that we are referencing a method of a different class now that `take_action` is not part of `Trainer` type.

ml-agents/mlagents/trainers/policy.py
harperj: @eshvk I think that's a very good suggestion.  No action is being taken here, we are just looking up the correct action info for a given set of obs/reward :)

ml-agents/tests/trainers/test_policy.py
```diff
@@ -0,0 +1,51 @@
...
+    brain_info_with_agents = BrainInfo([], [], [], agents=['an-agent-id'])
+    result = policy.take_action(brain_info_with_agents)
+    expected = ActionInfo(
+        policy_eval_out['action'],
+        policy_eval_out['memory_out'],
+        None,
+        policy_eval_out['value'],
+        policy_eval_out
+    )
+    assert (result == expected)
```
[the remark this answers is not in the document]
harperj: @mmattar Do you know how we can configure this so we can avoid these unnecessary issues?  Asserts in tests shouldn't be an issue.

ml-agents/mlagents/trainers/action_info.py
```diff
@@ -0,0 +1,9 @@
+from typing import NamedTuple, Any, Dict, Optional
+
+
+class ActionInfo(NamedTuple):
```
eshvk: 😍

ml-agents/mlagents/trainers/action_info.py
```diff
@@ -0,0 +1,9 @@
...
+    action: Any
+    memory: Any
+    text: Any
```
eshvk: @vincentpierre @awjuliani `take_action` for both `BCTrainer` and `PPOTrainer` returns None for this entry of the tuple. Is there a point in even having this as a return from `take_action`?

ml-agents/mlagents/trainers/policy.py
eshvk: Yeah, exactly! Please feel free to incorporate it now in this PR or we can table it for a future PR.

harperj:
Review comments addressed, PTAL

ml-agents/mlagents/trainers/action_info.py
harperj: Per our discussion, we'll deprecate this later.

ml-agents/mlagents/trainers/trainer_controller.py
eshvk: I'd be curious to know if this change still makes sense. We are now exposing to the `TrainerController` the `Policy` object(s). Initially, I think it made sense because the `Policy`(s) will be distributed to different `ActorManager` processes. How about now that we might not be doing that?

eshvk: I'd be also curious to hear your thoughts on how we could make the following a bit more explicit: Now, we have a local variable in `start_learning`,  `policies` that is mutated when `update_policy` is called in another method `take_step` for each iteration of the loop.

ml-agents/mlagents/trainers/trainer.py
```diff
@@ -139,6 +130,12 @@ def update_policy(self):
         """
         raise UnityTrainerException("The update_model method was not implemented.")
 
+    def curriculum_lesson_changed(self) -> None:
```
eshvk: Clarifying question: Does the idea of a curriculum make sense for all types of `Trainer`? E.g. `BCTrainer`. I am wondering if this method should be specific to `PPOTrainer` (s)?

cc @vincentpierre

harperj: Yeah, I was rethinking this one recently as well.  I think curricula don't make much sense for `BCTrainer`, though in theory there's no reason why the idea of changing the reset parameters based on certain criteria would be a problem.

Maybe the idea of the `reward_buffer`, which is really the issue here, should be moved up into the TrainerController / Curriculum class since it's only used for that purpose at this time.

ml-agents/mlagents/trainers/trainer_controller.py
harperj: I'd propose for now we just get rid of `TrainerController` knowing anything about policies until the time we need it.  I do think that the Policy is the right place for `get_action` so we can just additionally expose that via the Trainer.  This will make sense if we want a trainer to potentially know about multiple policies in the future as well.

ml-agents/mlagents/trainers/trainer.py
harperj: After some thought, I am going to remove this method and revert the code around the `reward_buffer` to how it was before.  I think we should deal with that issue separately from this PR.

ml-agents/mlagents/trainers/trainer_controller.py
harperj: I made this change, PTAL.

eshvk: 👍

Issue #1693, opened 2019-02-09
Use RenderTexture instead of Camera for Visual Observation


pyjamads:
Suggestion:
I would like to render straight to a RenderTexture manually and use that as visual observation instead of using a camera.
I have found a few ways to resolve it for now, but I think ML-agents would benefit from adding RenderTexture as an alternative to the Camera.

For others looking, the solution suggested here, by rendering to a texture in a canvas is a way around it: https://github.com/Unity-Technologies/ml-agents/issues/611

Another alternative would be to do something like this Frame stacking example, that supports textures as an alternative to cameras:
https://github.com/Unity-Technologies/ml-agents/issues/952

ervteng:
That could be pretty useful for a number of cases! This feature isn't currently on our roadmap, but If a community member wants to contribute an implementation of the feature to ml-agents, it would be very much appreciated.

pyjamads:
I created a pull request with the functionality #1698

ervteng:
Huge thanks, we will take a look!

pyjamads:
ups, looks like I forgot a check for the RenderTexture size.

It should be added it to the pull request.

ervteng:
We've merged this feature into our `develop` branch, and it will show up in our next release. Thanks again for the contribution!

pyjamads:
Thanks, that sounds great!

PR #1698, opened 2019-02-11
Use RenderTexture instead of Camera for Visual Observation

pyjamads:
Created in relation to #1693 

Added support for Visual Observations made from RenderTextures in addition to cameras.


ervteng:
Hey @pyjamads, we took a look at the PR and we'd like to merge it in as it will be useful to a lot of people! I think the implementation looks great. Thanks again. 

With all of our features we like to include an example environment that demonstrates its usage, and some documentation, so we'd need these two items before merging. Do you have a simple example env of using RenderTextures? If not I recommend taking the GridWorld example and making it render to a RenderTexture rather than a Camera. 

Also, if you could add some documentation (under "Multiple Visual Observations") for using your feature [here](https://github.com/Unity-Technologies/ml-agents/blob/master/docs/Learning-Environment-Design-Agents.md) that would be super helpful!

pyjamads:
I used the BananaVisual setup for testing, but seeing as that environment is hard to learn, it's probably not the best example. I'll make a copy of the GridWorld example instead, and look at where to extend the documentation.

pyjamads:
Added documentation, GridWorld Example, and a tweak to the agent inspector, so the help box only shows up when you're actually using both cameras and render textures together.

ervteng:
Hey @pyjamads, apologize for the delay. I was able to clone your repo and test out your environment. I'm getting a weird visual artifact and a missing prefab on the GridWorldRenderTexture - is this expected?

Thanks!
rendertexture-gridworld

pyjamads:
Hi @ervteng 

The visual artifact, is the canvas with the RawImage render texture on it, but it seems the RawImage wasn't referencing the RenderTexture, so it was just showing a white square, this is what I'd expect in the scene:
image
And it should look like this in the Game View, once the renderTexture is attached correctly:
image

The Prefab is just the Agent prefab, which it seems wasn't "serialized" after I added the extra agentParameter, so I've fixed the missing reference in RawImage, and removed/added the agent again, as well as forcing the agent prefab to be serialized and pushed those changes (scene + agent prefab).

ervteng:
Hey @pyjamads, thank you for the fix! The RenderTexture now works for me. But I'm unable to instantiate your agent Prefab - Unity is throwing the "Unable to instantiate prefab. Prefab may be broken" error. Not sure if the serialization messed with the prefab at all. 

Maybe as a debug step you could send me the .prefab file directly and I can try to import it? Perhaps Git is messing with the file.

pyjamads:
Hi @ervteng 
This is the prefab file, rename the extension to .prefab
[agent.txt](https://github.com/Unity-Technologies/ml-agents/files/2909164/agent.txt)

ervteng:
Hey @pyjamads, apologize for the delay, I just got around to testing your prefab. Thanks for sending it over. 

I am still unable to instantiate it, however. If I drag it into the scene it's giving the "Prefab may be broken" error, for both the one you committed and the one you sent me. But otherwise I can edit the prefab by itself. I think I'll try to re-create the prefab from the old version. Is the only change that we have the RenderTexture box?

ervteng:
Hey @pyjamads, I was able to get your branch of the code to run by re-importing the old prefab. We'll be able to merge the change now. Thanks for your help again!

I also ran into an issue where the grid world's observations were not being updated as quickly when using RenderTexture vs. Camera. This is because we call `Camera.Render()` in Agent.cs before making a decision, regardless of how fast the actual simulation is running, but the same is not done with RenderTextures. I was able to fix this by adding a Camera.Render() call in the GridAgent.cs. Just wanted to let you know in case you run into this issue in the future.

pyjamads:
Right it makes sense that you have to manually update your RenderTexture when running CollectObservation happens, for my own usecase, I am rendering a 2D sprite scene directly to a RenderTexture when the call happens. It might be worth adding that little note to the manual, in order to avoid one or more frames of lag, caused by using RenderTextures.

PR #1702, opened 2019-02-12
convenient isntall of tensorflow gpu via conda

JulesDoe:
Its mutch more convenient to have conda install everything for you.
And im not shoure about that but in my (not very complete) testing, it seemd that its not necessary to have tensorflow 1.7.1. but works fine with 1.12.0.

### DIFF
--- a/docs/Installation-Windows.md
+++ b/docs/Installation-Windows.md
@@ -163,10 +163,137 @@ check Nvidia's page [here](https://developer.nvidia.com/cuda-gpus).
 
 Currently for the ML-Agents toolkit, only CUDA v9.0 and cuDNN v7.0.5 is supported.
 
+>If yo prefere to know wich parts are needed continue reading. If you just want 
+to get installed as quick as posible see the Anadonca way of installing it [here](instal-tensorflow-gpu-via-anaconda).
+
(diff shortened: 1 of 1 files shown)


eshvk:
@JulesDoe thank you for the PR. One thought:  I'd prefer if we keep the install instructions as is and add additional instructions for Anaconda. This is because our current installation of `mlagents-learn` is designed to not exclusively depend on just Conda.

JulesDoe:
@xiaomaogy since it seems to me that you did the last tensorflow version bump. do you happen to know if ther are parts that break when running ml-agents with tensorflow greater than 1.7 ... or are the docs just outdated?

xiaomaogy:
@JulesDoe We didn't want to go above 1.7 because TensorFlowSharp doesn't have a version above 1.7.1, now we replace that with Barracuda it is actually possible. However we don't have the bandwidth to do that right now.

xiaomaogy:
Also I'm going to close this PR, because:

1. The links doesn't work. 
2. There is grammar problems in the sentences. 
3. We are actually planning to release this week, and can't wait for you to do the change. 

Sorry about that= =

Issue #1706, opened 2019-02-12
typo in Installation-Windows.md


dmalpica:
**With your approval, I'd like to submit a fix for the windows installation guide:**

**The instructions are a little unclear as there is no mention of invoking the 'python' command to launch the REPL from the Anaconda Prompt.**

> Lastly, you should test to see if everything installed properly and that TensorFlow can identify your GPU. In the same Anaconda Prompt, type in the following command:
> 
> import tensorflow as tf
> 
> sess = tf.Session(config=tf.ConfigProto(log_device_placement=True))

**I can see the problem in the markdown, the word 'python' is on the same line as the opening triple single quotes.**

**Thanks for the guide!**

dmalpica:
I would also like to add a warning to this guide: Installing the CUDA toolkit as is may overwrite your Nvidia display drivers. This can cause problems with other applications such as photoshop, Maya and 3D Studio Max.

awjuliani:
Hi @dmalpica 

Thanks for your interest in the ML-Agents toolkit. Please feel free to make a PR with these changes. We can review them from there.

awjuliani:
Closing this issue for now. Please feel free to re-open if you deem it necessary.

PR #1760, opened 2019-02-25
Develop windows install instructions update

dmalpica:
As discussed here: https://github.com/Unity-Technologies/ml-agents/issues/1706

### DIFF
--- a/docs/Installation-Windows.md
+++ b/docs/Installation-Windows.md
@@ -58,8 +58,7 @@ From here, click the __Environment Variables__ button. Double click "Path" under
 __System variable__ to edit the "Path" variable, click __New__ to add the
 following new paths.
 
-```
-console
+```console
 %UserProfile%\Anaconda3\Scripts
 %UserProfile%\Anaconda3\Scripts\conda.exe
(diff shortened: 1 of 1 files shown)


dmalpica:
@vincentpierre Thanks for the feedback. I looked around but couldn't find a solid source documenting the issues I experienced, so I took out the warning.

That said, I found a more recent post that might be helpful:
https://www.pugetsystems.com/labs/hpc/The-Best-Way-to-Install-TensorFlow-with-GPU-Support-on-Windows-10-Without-Installing-CUDA-1187/

I addressed the other issue differentiating the anaconda prompt command from the python commands.

> I am not sure the Warning is necessary given that it is not a ML-Agents issue. Maybe adding a link to a CUDA documentation page that discusses this issue would be good if the Warning stays.
> I am not a fan of having
> 
> ```shell
> python
> 
> <python code>
> ```
> 
> Because the two are in different languages. I would prefer : Open Python in the Prompt by calling
> 
> ```shell
> python
> ```
> 
> and then type the following commands
> 
> ```python
> <python code>
> ```
> 
> to test if everything was installed properly

PR #1813, opened 2019-03-11
Soccer Twos - Fixes missing tag change, plus code cleanup

awjuliani:
* Tag on agent was not changed when assigned to a new team.

(We probably want to merge this into a v0.7.0a hotfix, or potentially roll into v0.8.


UnitySDK/Assets/ML-Agents/Examples/Soccer/Scripts/AgentSoccer.cs
```diff
@@ -67,14 +81,14 @@ public override void CollectObservations()
         float rayDistance = 20f;
         float[] rayAngles = { 0f, 45f, 90f, 135f, 180f, 110f, 70f };
         string[] detectableObjects;
-        if (team == Team.red)
+        if (team == Team.Red)
         {
-            detectableObjects = new string[] { "ball", "redGoal", "blueGoal",
+            detectableObjects = new[] { "ball", "redGoal", "blueGoal",
```
vincentpierre: Why did you change that ? It makes it less explicit.

awjuliani: The code is more concise. The extra string is redundant, because the contents are clearly already strings.

XFFXFF:
I build soccer based on this branch, but the goalkeeper went out of the penalty area and acted more like a striker. 
image
My env info is the following
image

awjuliani:
Hi @XFFXFF 

Thanks for pointing out this bug. I've just added a commit which will address this.

PR #1849, opened 2019-03-20
fixed localhost address for macOS

djsdvg:
I cannot start the training because i get this error:
_**File "/Users/user/Documents/ml-agents/ml-agents/mlagents/envs/rpc_communicator.py", line 70, in check_ports.bind(("localhost", port))socket.gaierror: [Errno 8] nodename nor servname provided, or not known**_

Using **socket.gethostname()** instead the training starts properly
I think that the error is linked to the fact that macOS use 0.0.0.0 instead of 127.0.0.0


harperj:
Hi @djsdvg -- this looks like a good change but it's unlikely we'll be releasing a new patch version of v0.7 in the near future since we're getting ready to release v0.8.  I pointed this PR at the develop branch so it can make it into the new release.

harperj:
This fix needed to be rebased onto our develop branch but the PR has been idle for a while now.  I'm going to close it but please reopen if you'd like to add this patch on top of `develop`.

PR #1893, opened 2019-04-04
Included TFS page to redirect

unityjeffrey:
There are historical links to the old TFS page that are breaking for some users.  Added this page in order to keep the links unbroken and direct users to the new Unity Inference Engine documentation


Guidosalimbeni:
ops... that was very useful to load custom graph in unity... can you reset the link please?

xiaomaogy:
@Guidosalimbeni We haven't deleted the TFS download for older version, you can still use them.

PR #1896, opened 2019-04-05
Fix not saving .nn file after max_timesteps

ervteng:


ervteng:
Note: On Docker it does save .nn but we get this error. 

`Process Process-1:
  File "/usr/local/lib/python3.6/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/local/lib/python3.6/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/ml-agents-envs/mlagents/envs/subprocess_environment.py", line 70, in worker
    env.close()
  File "/ml-agents-envs/mlagents/envs/environment.py", line 467, in close
    raise UnityEnvironmentException("No Unity environment is loaded.")
mlagents.envs.exception.UnityEnvironmentException: No Unity environment is loaded.`

PR #1912, opened 2019-04-08
Fix environment factory pickling on Windows

harperj:
SubprocessUnityEnvironment sends an environment factory function to
each worker which it can use to create a UnityEnvironment to interact
with. We use Python's standard multiprocessing library, which pickles
all data sent to the subprocess.  The built-in pickle library doesn't
pickle function objects on Windows machines (tested with Python 3.6 on
Windows 10 Pro).

This PR adds cloudpickle as a dependency in order to serialize the
environment factory. Other implementations of subprocess environments
do the same:
https://github.com/openai/baselines/blob/master/baselines/common/vec_env/subproc_vec_env.py

### DIFF
--- a/ml-agents-envs/mlagents/envs/subprocess_environment.py
+++ b/ml-agents-envs/mlagents/envs/subprocess_environment.py
@@ -1,6 +1,7 @@
 from typing import *
 import copy
 import numpy as np
+import cloudpickle
 
 from mlagents.envs import UnityEnvironment
 from multiprocessing import Process, Pipe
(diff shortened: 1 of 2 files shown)


xiaomaogy:
@harperj Just to make sure, does the model get saved under Windows for these two cases? 
1) ctrl + c interrupt. 
2) Finish the training normally. 

Also have you tested the Linux environment for these two cases?

SetoKaiba:
1. Stop from Unity Editor.
Now: Model is not saved.
Previous: Model is saved. It's saved even just after the training begins for seconds.

2. Run out the whole training for max_steps. 
Now: The Unity Editor is frozen. And the training is never finished. The checkpoints are saved. But the nn model is not. Press Ctrl C and the Unity Editor is not frozen any more. The nn model is saved.
Previous: The Unity Editor will stop playing. The training is finished. Both checkpoints and nn model are saved. You don't have to press Ctrl C.

3. Stop with Ctrl C.
Now: Model is not saved.
Previous: Model is saved. It's saved even just after the training begins for seconds.

xiaomaogy:
@SetoKaiba Thanks for testing this for us! @harperj Just showed me a fix yesterday afternoon, which works for case 2 and case 3 now, I believe he will test the case 1 and then update this branch.

harperj:
@xiaomaogy @SetoKaiba Those are different (unrelated) issues which I'll shortly make subsequent PRs for.

Issue #1389, opened 2018-10-30
Tensorflow 1.8 & 1.9 suppport


BernardinD:
Are the Tensorflow versions after 1.7 supported? I have been trying to run a graph in Unity that was trained in Tensorflow 1.9. I tried converting it using the script segment in the Readme but it gives me an error: "TypeError: names_to_saveables must be a dict mapping string names to Tensors/Variables. Not a variable: Tensor("BoxPredictor_0/BoxEncodingPredictor/biases:0", shape=(12,), dtype=float32". Any suggestions?

Thx in advance

xiaomaogy:
Hi @BernardinD , currently we don't have any plan to support TensorFlow after 1.7.1. So the best way would be to stick with the 1.7 version. Just curious, what is your use case for Tensorflow 1.9?  I understand newer version is always better, but just wondering what are the things you can can't do on 1.7.

SetoKaiba:
@xiaomaogy Only 1.8 and later with ROCm support can run on AMD GPU. Maybe it can be a reason why 1.8 support is needed.

Amanpradhan:
@BernardinD  How did you convert your graph from tensorflow 1.9 to 1.7?

BernardinD:
@Amanpradhan I believe all that is needed is that you roll back your _models_ to a time corresponding with 1.7

SetoKaiba:
Bump this again. Tensorflow GPU 1.7.1 is working with cuda 9.0 & cudnn 7.0 while Octane for Unity is working with newer cuda version & cudnn 7.4.1.

SetoKaiba:
Ok. I found that I updated the driver and it's working well. Anyway, TensorFlow 2.0 is released. It would be great to support it. There're many improvements.

SetoKaiba:
@xiaomaogy Could you please have a look at it? I updated the ml-agents python part to TF 2.0. It's almost working except for a problem I can't find out the reason cause it. Here's my repo and output.
(https://github.com/SetoKaiba/ml-agents/tree/tf2)

> (ml-agents-2) E:\Seto\GitHub\Unity-Technologies\ml-agents>mlagents-learn config\trainer_config.yaml --train
> 
> 
>                         ▄▄▄▓▓▓▓
>                    ╓▓▓▓▓▓▓█▓▓▓▓▓
>               ,▄▄▄m▀▀▀'  ,▓▓▓▀▓▓▄                           ▓▓▓  ▓▓▌
[...]

PR #1917, opened 2019-04-09
Fix subprocess model saving on Windows

harperj:
On Windows the interrupt for subprocesses works in a different
way from OSX/Linux. The result is that child subprocesses and
their pipes may close while the parent process is still running
during a keyboard (ctrl+C) interrupt.

To handle this, this change adds handling for EOFError and
BrokenPipeError exceptions when interacting with subprocess
environments. Additional management is also added to be sure
when using parallel runs using the "num-runs" option that
the threads for each run are joined and KeyboardInterrupts are
handled.

These changes made the "_win_handler" we used to specially
manage interrupts on Windows unnecessary, so they have been
removed.

Brought up in #1389, follow up to #1912.

### DIFF
--- a/ml-agents-envs/mlagents/envs/subprocess_environment.py
+++ b/ml-agents-envs/mlagents/envs/subprocess_environment.py
@@ -27,15 +27,24 @@ class UnityEnvWorker(NamedTuple):
     conn: Connection
 
     def send(self, name: str, payload=None):
-        cmd = EnvironmentCommand(name, payload)
-        self.conn.send(cmd)
+        try:
+            cmd = EnvironmentCommand(name, payload)
+            self.conn.send(cmd)
+        except (BrokenPipeError, EOFError):
+            raise KeyboardInterrupt
 
     def recv(self) -> EnvironmentResponse:
-        response: EnvironmentResponse = self.conn.recv()
-        return response
+        try:
+            response: EnvironmentResponse = self.conn.recv()
+            return response
+        except (BrokenPipeError, EOFError):
+            raise KeyboardInterrupt
 
     def close(self):
-        self.conn.send(EnvironmentCommand("close"))
+        try:
+            self.conn.send(EnvironmentCommand('close'))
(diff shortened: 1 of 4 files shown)


ml-agents-envs/mlagents/envs/subprocess_environment.py
```diff
@@ -27,15 +27,24 @@ class UnityEnvWorker(NamedTuple):
...
+            raise KeyboardInterrupt
 
     def recv(self) -> EnvironmentResponse:
-        response: EnvironmentResponse = self.conn.recv()
-        return response
+        try:
+            response: EnvironmentResponse = self.conn.recv()
+            return response
+        except (BrokenPipeError, EOFError):
+            raise KeyboardInterrupt
```
xiaomaogy: Why do you raise KeyboardInterrupt here?

xiaomaogy:
Also have you tested the following things on both Windows and Linux?

1. Editor training
2. ctrl+c interrupt, model and summaries gets saved. 
3. Training ends normally, model and summaries gets saved. 
4. --num-runs still works correctly.

ml-agents-envs/mlagents/envs/subprocess_environment.py
harperj: This should really probably be some custom exception, but in our case KeyboardInterrupt is a proxy for training being interrupted.  This is actually being sent from the main process which has the trainer and SubprocessUnityEnvironment.  If the pipe is broken or we receive an EOF, it means the worker has died before we told it to close -- this means either a keyboard interrupt happened (like on Windows) and the worker died before the main process or the environment process was manually destroyed, etc.  This way we'll still save the model.

harperj:
@xiaomaogy I've tested on Windows and OSX, and we'll further test on Linux shortly for the release checklist.

SetoKaiba:
Thank you. All the stuff is working now. I'll merge into my TF 2.0 branch to see whether it's working correctly.

PR #1952, opened 2019-04-16
Add use curiosity in Hallway example environment

sdiao:
Hallway example takes a long time to train. To speed up training, we can enable curiosity. This change is to add a suggestion in the doc in case our customers want to speed up the Hallway example training.

### DIFF
--- a/docs/Learning-Environment-Examples.md
+++ b/docs/Learning-Environment-Examples.md
@@ -246,6 +246,7 @@ If you would like to contribute environments, please see our
     `VisualHallway` scene.
 * Reset Parameters: None.
 * Benchmark Mean Reward: 0.7
+  * To speed up training, you can enable curiosity by adding `use_curiosity: true` in `config/trainer_config.yaml`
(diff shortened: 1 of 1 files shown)


docs/Learning-Environment-Examples.md
```diff
@@ -246,6 +246,7 @@ If you would like to contribute environments, please see our
     `VisualHallway` scene.
 * Reset Parameters: None.
 * Benchmark Mean Reward: 0.7
+  * To speed up training, you can enable curiosity by adding `use_curiosity: true` in `config/trainer_config.yaml`
```
mmattar: Is the intention to have space prior to the asterisk?

sdiao: yes. It looks all examples have the same list of bullet points (Set-up, Goal,..., Benchmark Mean reward). I want to keep the overall structure, and only add enabling curiosity to achieve the benchmark mean reward faster. So I add this optimization under the benchmark mean reward. Let me know if there is a better place to add a note of enabling curiosity.

PR #2090, opened 2019-05-30
Update pip and setuptools in CircleCI

harperj:
The CircleCI checks have been broken because of outdated setuptools,
this change should fix the issue.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,6 +20,8 @@ jobs:
           command: |
             python3 -m venv venv
             . venv/bin/activate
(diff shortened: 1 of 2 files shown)


xiaomaogy:
Also maybe we should remove the `black --check ml-agents` since we don't really care about it.

harperj:
@xiaomaogy Why don't we care about `black`?  I think we should absolutely continue using that.

xiaomaogy:
Your 4587a4c commit is the answer. I personally don't mind keeping it there.

PR #2091, opened 2019-05-30
Fix broken demonstration recorder

awjuliani:
* Fixes issue caused by improper resolved conflict.


xiaomaogy:
This fixes the build error and I've tested it.

PR #2118, opened 2019-06-10
GAIL and Pretraining

ervteng:
Based on the new reward signals architecture, add BC pretrainer and GAIL for PPO. Main changes:

- A new GAILRewardSignal and GAILModel for GAIL
- A BCModule component (not a reward signal) to do pretraining during RL
- Documentation for both of these
- Change to Demo Loader that lets you load multiple demo files in a folder
- Example Demo files for all of our tested sample environments (for future regression testing)


ml-agents/mlagents/trainers/components/reward_signals/reward_signal.py
```diff
@@ -0,0 +1,57 @@
+import logging
+from mlagents.trainers.trainer import UnityTrainerException
+from mlagents.trainers.policy import Policy
+
+import tensorflow as tf
+
+logger = logging.getLogger("mlagents.trainers")
+
+
+class RewardSignal(object):
```
chriselion: Worth using abstract base class (https://docs.python.org/3/library/abc.html) and making these `@abc.abstractmethod`'s?

ml-agents/mlagents/trainers/components/reward_signals/reward_signal.py
```diff
@@ -0,0 +1,57 @@
...
+        :param training_buffer: The training buffer
+        :param n_sequences: The number of sequences in the training buffer.
+        :return: A dict of {"Stat Name": stat} to be added to Tensorboard
+        """
+        raise UnityTrainerException(
+            "The update for this RewardSignal was not implemented."
+        )
+
+    @classmethod
+    def check_config(cls, config_dict, param_keys=[]):
```
chriselion: It's probably OK here, but generally it's a bad idea to have mutable default arguments in python methods - see https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments.

A better way would be something like 
``` python
def check_config(cls, config_dict, param_keys=None):
    param_keys = param_keys or []
    for k in param_keys:
        # ...
```

ml-agents/mlagents/trainers/ppo/models.py
```diff
@@ -79,195 +77,77 @@ def create_reward_encoder():
...
         :param epsilon: Value for policy-divergence threshold
         :param lr: Learning rate
         :param max_step: Total number of training steps.
         """
-        self.returns_holder = tf.placeholder(
-            shape=[None], dtype=tf.float32, name="discounted_rewards"
-        )
+        self.returns_holders = []
+        self.old_values = []
+        for i in range(len(value_streams)):
```
chriselion: I'm a little confused on the types/uses of `value_streams` and `stream_names`. Former is a dict of str->some tf object, and latter is a list of str? Is `stream_names` a subset of `value_streams.keys()`? Can you have a `stream_name` that's not a key in `value_streams`?

If `stream_names` is a subset of `value_streams.keys()` then it feels strange to iterate over the `len(value_streams)` here. 

If you can have a stream_name that's not a key in value_streams, the `for i, name in enumerate(stream_names):` seems wrong because it uses them as keys.

If they're identical, then you should drop stream_names as a parameter and just `value_streams.keys()`

ml-agents/mlagents/trainers/ppo/policy.py
```diff
@@ -33,11 +37,28 @@ def __init__(self, seed, brain, trainer_params, is_training, load):
...
-                use_curiosity=bool(trainer_params["use_curiosity"]),
-                curiosity_strength=float(trainer_params["curiosity_strength"]),
-                curiosity_enc_size=float(trainer_params["curiosity_enc_size"]),
                 seed=seed,
+                stream_names=list(reward_signal_configs.keys()),
             )
+            self.model.create_ppo_optimizer()
+
+            # Create reward signals
+            for _rsignal in reward_signal_configs.keys():
```
chriselion: nit: `signal_name` instead of `_rsignal`?

chriselion: even better:
```
for signal_name, config in reward_signal_configs.items():
    self.reward_signals[signal_name] = create_reward_signal(self, signal_name, config)

File: ml-agents/mlagents/trainers/tests/test_bcmodule.py [hunk 5]
```diff
@@ -0,0 +1,89 @@
...
+
+from mlagents.trainers.ppo.models import PPOModel
+from mlagents.trainers.ppo.trainer import discount_rewards
+from mlagents.trainers.ppo.policy import PPOPolicy
+from mlagents.envs import UnityEnvironment
+
+
+@pytest.fixture
+def dummy_config():
+    return yaml.load(
```
chriselion: nit: I switched the tests to use `yaml.safe_load` in https://github.com/Unity-Technologies/ml-agents/pull/2124 instead to clean up some warnings

ml-agents/mlagents/trainers/components/reward_signals/reward_signal.py
chriselion: I think another way you could do this is something like
``` python
class RewardSignal(object):
  REQUIRED_PARAM_KEYS = []

  @classmethod
  def check_config(cls, config_dict):
  for k in cls.REQUIRED_CONFIG_PARAMS:
    # ...


class GAILRewardSignal(RewardSignal):
    REQUIRED_PARAM_KEYS = ["strength", "gamma", "demo_path"]
```
and then you wouldn't need to override check_config in each subclass.

ml-agents/mlagents/trainers/components/reward_signals/curiosity/signal.py
```diff
@@ -0,0 +1,168 @@
...
+        Creates the Curiosity reward generator
+        :param policy: The Learning Policy
+        :param encoding_size: The size of the Curiosity encoding
+        :param signal_strength: The scaling parameter for the reward. The scaled reward will be the unscaled
+        reward multiplied by the strength parameter
+        """
+        self.policy = policy
+        self.strength = strength
+        self.gamma = gamma
+        self.stat_name = "Policy/Curiosity Reward"
```
chriselion: Do stat_name or value_name get used anywhere? Don't see them referenced for this or any of the other RewardSignal implementations.

If they are going to be used, maybe it would be better to have them as properties in the base class; you could munge them from the `cls.__name__` by default or allow them to be overridden.

ml-agents/mlagents/trainers/ppo/models.py
```diff
@@ -285,6 +165,6 @@ def create_ppo_optimizer(
             * tf.reduce_mean(tf.dynamic_partition(entropy, self.mask, 2)[1])
         )
 
-        if self.use_curiosity:
-            self.loss += 10 * (0.2 * self.forward_loss + 0.8 * self.inverse_loss)
+    def create_ppo_optimizer(self):
```
chriselion: Is ppo_optimizer like "ATM Machine"? :)

ml-agents/mlagents/trainers/ppo/policy.py
```diff
@@ -111,13 +135,19 @@ def update(self, mini_batch, num_sequences):
...
             self.model.mask_input: mini_batch["masks"].flatten(),
-            self.model.returns_holder: mini_batch["discounted_returns"].flatten(),
-            self.model.old_value: mini_batch["value_estimates"].flatten(),
             self.model.advantage: mini_batch["advantages"].reshape([-1, 1]),
             self.model.all_old_log_probs: mini_batch["action_probs"].reshape(
                 [-1, sum(self.model.act_size)]
             ),
         }
+        for i, name in enumerate(self.reward_signals.keys()):
+            feed_dict[self.model.returns_holders[i]] = mini_batch[
```
chriselion: Are the `returns_holders` and `old_values` guaranteed to be in same order as reward_signals?  

Can we just key them by the name instead of in a list? (this goes back to my question about reward_signals vs. stream_names)

ml-agents/mlagents/trainers/ppo/models.py
ervteng: Oh yeah, yep :P

Proximal Policy Optimization Optimizer - guess the common name for the algorithm is PPO

ml-agents/mlagents/trainers/ppo/policy.py
ervteng: You're exactly right - on my other branch I've changed it so that the model only refers to them by name and not index. I think I'll make these changes here as well

ml-agents/mlagents/trainers/components/reward_signals/reward_signal.py
ervteng: I did have it like that in an earlier iteration, but wasn't sure about it since it isn't a construct we use anywhere else in the code. Maybe it's worth bringing back?

ml-agents/mlagents/trainers/components/reward_signals/reward_signal.py
ervteng: Got it - I'll fix this

ml-agents/mlagents/trainers/ppo/models.py
chriselion: Not a big deal, just sounded funny in my head :)

ml-agents/mlagents/trainers/components/reward_signals/reward_signal.py
chriselion: Your call - the other option is to provide a default implementation some of the methods, e.g. `return {}` in `update()`

ml-agents/mlagents/trainers/ppo/policy.py
ervteng: I am now referring these by name - marking as resolved

ml-agents/mlagents/trainers/ppo/policy.py
ervteng: Done!

ml-agents/mlagents/trainers/ppo/models.py
ervteng: I re-did this - now I'm always referring to things by stream name and never by index. Also removed stream_names and just use the keys from the `value_streams` (now `value_heads`) parameter. 

The `value_streams` aren't the same as the `stream_names` - renamed `value_streams` to `value_heads` (they're Tensorflow nodes) and removed stream_names. Let me know how that looks!

chriselion: Sounds cleaner, will look again tomorrow.

ml-agents/mlagents/trainers/components/reward_signals/curiosity/signal.py
ervteng: Hey @chriselion, they are used - every RewardSignal needs both of these. They're used for stats reporting.

I've added them to the base class, though rather than using the class name I'd rather have the "human friendly" name reported here (e.g. "Curiosity Reward" instead of "CuriosityRewardSignal"). So for now I'll override them - but maybe there's a better way?

ml-agents/mlagents/trainers/bc/models.py
```diff
@@ -15,7 +15,9 @@ def __init__(
         use_recurrent=False,
         seed=0,
     ):
-        LearningModel.__init__(self, m_size, normalize, use_recurrent, brain, seed)
+        LearningModel.__init__(
```
chriselion: The pythonic way to call the parent class initializer is `super().__init__` https://stackoverflow.com/a/12557639/224264

ml-agents/mlagents/trainers/components/bc/model.py
```diff
@@ -0,0 +1,107 @@
...
+    def __init__(self, policy_model: LearningModel, lr, anneal_steps):
+        """
+        Tensorflow operations to perform Behavioral Cloning on a Policy model
+        :param policy_model: The policy of the learning algorithm
+        :param lr: The initial learning Rate for behavioral cloning
+        :param anneal_steps: Number of steps over which to anneal BC training
+        """
+        self.z_size = 128
+        self.alpha = 0.0005
+        self.mutual_information = 0.5
```
chriselion: Is this used? Only see it used on the GAILModel

chriselion: Should these be options in the constructor?

ml-agents/mlagents/trainers/components/reward_signals/entropy/signal.py
```diff
@@ -0,0 +1,22 @@
+from mlagents.trainers.components.reward_signals import RewardSignal
+from mlagents.trainers.policy import Policy
+
+
+class EntropySignal(RewardSignal):
```
chriselion: `EntropyRewardSignal`?

ml-agents/mlagents/trainers/components/reward_signals/curiosity/signal.py
chriselion: OK, I see where it's used now.

I would probably do something like
``` python
class RewardSignal:
  def __init__(self):
    class_name = self.__class__.__name__
    short_name = class_name.replace("RewardSignal", "")
    self.stat_name = f"Policy/{short_name} Reward"
    self. value_name = f"Policy/{short_name} Value Estimate"
```
And then call the base constructor in each child class. That should set everything up (assuming you rename EntropySignal) with a reasonable default but still allow it to be overridden if desired.

ml-agents/mlagents/trainers/components/reward_signals/reward_signal.py
```diff
@@ -0,0 +1,64 @@
...
+import numpy as np
+import abc
+
+import tensorflow as tf
+
+logger = logging.getLogger("mlagents.trainers")
+
+
+class RewardSignal(abc.ABC):
+    value_name = "Policy/{} Value Estimate".format(__name__)
```
chriselion: Not sure `__name__` is what you want here. I think it will evaluate to something like
`mlagents.trainers.components.reward_signals.reward_signal`

ml-agents/mlagents/trainers/components/reward_signals/reward_signal.py
```diff
@@ -0,0 +1,64 @@
...
+    stat_name = "Policy/{} Reward".format(__name__)
+
+    @abc.abstractmethod
```
chriselion: Sorry to flipflop. I think if you have reasonable defaults for the methods, you don't need to make them abstract.

ml-agents/mlagents/trainers/components/reward_signals/reward_signal_factory.py
```diff
@@ -0,0 +1,41 @@
...
+)
+from mlagents.trainers.components.reward_signals.curiosity.signal import (
+    CuriosityRewardSignal,
+)
+from mlagents.trainers.policy import Policy
+
+logger = logging.getLogger("mlagents.trainers")
+
+
+NAME_TO_CLASS = {
```
chriselion: Add `Entropy[Reward]Signal`?

ml-agents/mlagents/trainers/demo_loader.py
```diff
@@ -61,10 +64,17 @@ def load_demonstration(file_path):
 
     # First 32 bytes of file dedicated to meta-data.
     INITIAL_POS = 33
-
-    if not os.path.isfile(file_path):
+    file_paths = []
```
chriselion: If you want to recurse into subfolders too, `os.walk()` is your friend. https://docs.python.org/3/library/os.html#os.walk

ml-agents/mlagents/trainers/bc/models.py
ervteng: Hmm - this code shouldn't be part of this PR, actually. Let me clean this up...

As far as the RewardSignals, I've changed them to use the super() call as appropriate.

ml-agents/mlagents/trainers/components/bc/model.py
ervteng: Yep. removed!

ml-agents/mlagents/trainers/components/reward_signals/entropy/signal.py
ervteng: Am going to remove the Entropy reward signal as it is handled already within the PPO (as entropy regularization). It also doesn't generalize well to SAC.

ml-agents/mlagents/trainers/demo_loader.py
ervteng: I actually specifically didn't want to walk through all of the subfolders. This was probably biased by my own workflow where I put the .demos that I don't want to use in an "old" folder. Open to changing it if it's less confusing to the user.

ml-agents/mlagents/trainers/components/reward_signals/reward_signal.py
ervteng: I went with the 1st method

ml-agents/mlagents/trainers/components/reward_signals/curiosity/signal.py
ervteng: This code works a treat, thanks!

ml-agents/mlagents/trainers/demo_loader.py
chriselion: Sounds good. Not sure what's more intuitive in general.

docs/Training-RewardSignals.md
```diff
@@ -112,3 +112,83 @@ This should typically be decreased if training is unstable, and the curiosity lo
...
+Typical Range: `0.8` - `0.9`
+
+#### Demo Path
+
+`demo_path` is the path to your `.demo` file or directory of `.demo` files. See the [imitation learning guide]
+(Training-ImitationLearning.md).
+
+#### Encoding Size
+
+`encoding_size` corresponds to the size of the hidden layer used by the discriminator.
```
chriselion: Typical/default values?

ml-agents/mlagents/trainers/components/bc/model.py
```diff
@@ -0,0 +1,101 @@
...
+                    )
+                    for i in range(len(self.policy_model.act_size))
+                ],
+                axis=1,
+            )
+
+    def create_loss(self, learning_rate: float, anneal_steps: int) -> None:
+        """
+        Creates the loss and update nodes for the GAIL reward generator
+        :param learning_rate: The learning rate for the optimizer
```
chriselion: nit: either add `anneal_steps` or remove this (I vote the latter)

ml-agents/mlagents/trainers/components/bc/model.py
```diff
@@ -0,0 +1,101 @@
...
+            self.action_in_expert = tf.placeholder(
+                shape=[None, action_length], dtype=tf.float32
+            )
+            self.expert_action = tf.identity(self.action_in_expert)
+        else:
+            action_length = len(self.policy_model.act_size)
+            self.action_in_expert = tf.placeholder(
+                shape=[None, action_length], dtype=tf.int32
+            )
+            self.expert_action = tf.concat(
```
chriselion: Not a big deal if you're just moving this code, but I'd go with something like 
```python
[
  tf.one_hot(
    self.action_in_expert[:, i], act_size
  )
  for i, act_size in enumerate(self.policy_model.act_size)
]
```

ml-agents/mlagents/trainers/components/bc/model.py
```diff
@@ -0,0 +1,101 @@
...
+                            )
+                            for i in range(len(action_size))
+                        ],
+                        axis=1,
+                    )
+                ),
+                axis=1,
+            )
+            self.loss = tf.reduce_mean(
+                -tf.log(tf.nn.softmax(log_probs) + 1e-7) * self.expert_action
```
chriselion: should epsilon's like this be parameters to the constructor instead?

ervteng: Not sure about a parameter as they'll never be changed by the user. But maybe we should make it a constant somewhere? Epsilons are used all over the codebase, so we could have ex. a common one for everything

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
```diff
@@ -0,0 +1,261 @@
...
+        self.update_beta = tf.assign(self.beta, new_beta)
+
+    def make_inputs(self) -> None:
+        """
+        Creates the input layers for the discriminator
+        """
+        self.done_expert = tf.placeholder(shape=[None, 1], dtype=tf.float32)
+        self.done_policy = tf.placeholder(shape=[None, 1], dtype=tf.float32)
+
+        if self.policy_model.brain.vector_action_space_type == "continuous":
```
chriselion: This setup is the same as in BCModel.make_inputs, right? Can it be extracted to a utility method?

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
```diff
@@ -0,0 +1,261 @@
...
+                dtype=tf.float32,
+                initializer=tf.ones_initializer(),
+            )
+            self.z_sigma_sq = self.z_sigma * self.z_sigma
+            self.z_log_sigma_sq = tf.log(self.z_sigma_sq + 1e-7)
+            self.use_noise = tf.placeholder(
+                shape=[1], dtype=tf.float32, name="NoiseLevel"
+            )
+        self.expert_estimate, self.z_mean_expert = self.create_encoder(
+            self.encoded_expert, self.expert_action, self.done_expert, False
```
chriselion: style nit: can you do `reuse=False` (and `reuse=True`) instead? I know it's not strictly necessary, but (imho) makes it clearer when reading (without having to look at the definition of create_encoder).

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
```diff
@@ -0,0 +1,261 @@
...
+
+    def create_loss(self, learning_rate: float) -> None:
+        """
+        Creates the loss and update nodes for the GAIL reward generator
+        :param learning_rate: The learning rate for the optimizer
+        """
+        self.mean_expert_estimate = tf.reduce_mean(self.expert_estimate)
+        self.mean_policy_estimate = tf.reduce_mean(self.policy_estimate)
+
+        self.disc_loss = -tf.reduce_mean(
```
chriselion: `discriminator_loss`?

ml-agents/mlagents/trainers/components/reward_signals/gail/signal.py
```diff
@@ -0,0 +1,277 @@
...
+            ].reshape([-1, self.policy.vec_obs_size])
+            feed_dict[self.model.obs_in_expert] = mini_batch_demo["vector_obs"].reshape(
+                [-1, self.policy.vec_obs_size]
+            )
+
+        # for reporting
+        po, pe, zlss, zme, zmp = None, None, None, None, None
+        kl_loss = None
+        # end for reporting
+        if self.model.use_vail:
```
chriselion: Seems like this would be cleaner as something like
```python
out_dict = {
    "gail_loss": self.model.loss
    # ... same for update_batch, policy_estimate, expert_estimate
}

if self.model.use_vail:
    out_dict["kl"] = kl_loss
    # same for other VAIL variables
run_out = self.policy.sess.run(out_dict, feed_dict=feed_dict)

if if self.model.use_vail:
    self.update_beta(run_out["kl_loss"])

return run_out
```
Not sure if that changes the effect of `sess.run` though. And not sure if `out_dict` is the right name, but seems to be what we use elsewhere.

docs/Training-RewardSignals.md
ervteng: added

ml-agents/mlagents/trainers/components/bc/model.py
ervteng: Hmm, remove the comment? Or remove the anneal steps?

docs/Training-Imitation-Learning.md
```diff
@@ -10,6 +10,35 @@ from the game and actions from a game controller to guide the medic's behavior.
...
+For instance, on the [Pyramids environment](Learning-Environment-Examples.md#pyramids), 
+just 6 episodes of demonstrations can reduce training steps by more than 4 times.
+
+<p align="center">
+  <img src="images/mlagents-ImitationAndRL.png"
+       alt="Using Demonstrations with Reinforcement Learning"
+       width="350" border="0" />
+</p>
+
+ML-Agents provides several ways to interact with demonstrations. For most situations,
```
awjuliani: Instead of "interact with demonstrations" maybe we want to say "learn from demonstrations" or "use demonstrations" ?

ml-agents/mlagents/trainers/components/bc/model.py
chriselion: Sorry, I mean either document both or neither. I think they're both fairly self-explanatory.

docs/Training-PPO.md
```diff
@@ -191,6 +190,73 @@ the agent will need to remember in order to successfully complete the task.
 
 Typical Range: `64` - `512`
 
+## (Optional) Pretraining Using Demonstrations
+
+In some cases, we would want to bootstrap the agent's policy using behavior recorded
```
awjuliani: Maybe replace "we would want" with "you might want" ?

docs/Training-PPO.md
```diff
@@ -191,6 +190,73 @@ the agent will need to remember in order to successfully complete the task.
...
+Typical Range (Discrete): `32` - `512`
+
+### (Optional) Number of Epochs
+
+`num_epoch` is the number of passes through the experience buffer during
+gradient descent. If not specified, it will default to the number of epochs set for PPO.
+
+Typical Range: `3` - `10`
+
+### (Optional) Max Batches
```
awjuliani: This one might be a little confusing to people.

ml-agents/mlagents/trainers/components/bc/model.py
```diff
@@ -0,0 +1,101 @@
...
+                        self.action_in_expert[:, i], self.policy_model.act_size[i]
+                    )
+                    for i in range(len(self.policy_model.act_size))
+                ],
+                axis=1,
+            )
+
+    def create_loss(self, learning_rate: float, anneal_steps: int) -> None:
+        """
+        Creates the loss and update nodes for the GAIL reward generator
```
awjuliani: This says GAIL reward generator, but the code is in the BC model?

ml-agents/mlagents/trainers/components/bc/model.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        selected_action = self.policy_model.output
+        action_size = self.policy_model.act_size
+        if self.policy_model.brain.vector_action_space_type == "continuous":
+            self.loss = tf.reduce_mean(
+                tf.squared_difference(selected_action, self.expert_action)
+            )
+        else:
+            log_probs = self.policy_model.all_log_probs
+            # self.expert_action = tf.placeholder(shape=[None, len(action_size)], dtype=tf.int32)
```
awjuliani: Should this be removed?

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
```diff
@@ -0,0 +1,261 @@
...
+        self,
+        policy_model: LearningModel,
+        h_size: int = 128,
+        learning_rate: float = 3e-4,
+        encoding_size: int = 64,
+        use_actions: bool = False,
+    ):
+        """
+        The initializer for the GAIL reward generator.
+        https://arxiv.org/abs/1606.03476
```
awjuliani: Do we want to reference VAIL as well? Since we are using it, it would probably help people reading out code to know why we are doing something so different from GAIL. Also, have we shown that VAIL does indeed outperform GAIL?

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
```diff
@@ -0,0 +1,261 @@
...
+        with tf.variable_scope("GAIL_model"):
+            if self.use_actions:
+                concat_input = tf.concat([state_in, action_in, done_in], axis=1)
+            else:
+                concat_input = state_in
+
+            hidden_1 = tf.layers.dense(
+                concat_input,
+                self.h_size,
+                activation=tf.nn.elu,
```
awjuliani: Do we want to use swish for these activations function as well (for consistency) ?

ml-agents/mlagents/trainers/components/reward_signals/gail/signal.py
```diff
@@ -0,0 +1,277 @@
...
+        n_sequences = max(n_sequences // 2, 1)  # Divide by 2 since we have two buffers
+        possible_demo_batches = (
+            len(self.demonstration_buffer.update_buffer["actions"]) // n_sequences
+        )
+        possible_policy_batches = len(update_buffer["actions"]) // n_sequences
+        possible_batches = min(possible_policy_batches, possible_demo_batches)
+
+        # for reporting
+        kl_loss = []
+        pos = []
```
awjuliani: Maybe we want to give these variables more semantically interpretable names?

ml-agents/mlagents/trainers/components/reward_signals/gail/signal.py
ervteng: I cleaned this up in the latest commit

ml-agents/mlagents/trainers/tests/test_reward_signals.py
```diff
@@ -92,64 +117,127 @@ def create_ppo_policy_mock(
...
     )
-    assert scaled_reward.shape == (12,)
-    assert unscaled_reward.shape == (12,)
+    assert rsig_result.scaled_reward.shape == (12,)
+    assert rsig_result.unscaled_reward.shape == (12,)
+
+
+def reward_signal_update(env, policy, reward_signal_name):
+    brain_info_list = []
+    for i in range(20):
```
awjuliani: There are a few numerical values hard-coded here. Can we replace them with vars/consts?

ml-agents/mlagents/trainers/components/reward_signals/gail/signal.py
ervteng: I updated these. TBH, we might be able to remove this since they're only there for printing during debug.

awjuliani: In that case I think it makes sense to remove them.

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
ervteng: Yes. Related note: should we call it `LearningModel.activation` rather than `LearningModel.swish` so that if we change it in the future we can just change one spot?

ml-agents/mlagents/trainers/components/reward_signals/gail/signal.py
```diff
@@ -0,0 +1,266 @@
...
+from mlagents.trainers.demo_loader import demo_to_buffer
+
+
+class GAILRewardSignal(RewardSignal):
+    def __init__(
+        self,
+        policy: TFPolicy,
+        strength: float,
+        gamma: float,
+        demo_path: str,
```
chriselion: (discussed offline) I think it we could make this take a Buffer instance instead of the path, it would be a little cleaner. But might not be easy to pass it down from the TrainerController.

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
ervteng: I agree. Would it be better to put a `utils.py` in the `components` directory, or should it be in one of the two classes and referred from the other class (this doesn't seem too clean)?

ml-agents/mlagents/trainers/components/reward_signals/gail/signal.py
ervteng: One way we could do this is to add a `demo_path` config to the highest level (the PPO trainer), and if it exists, the TrainerController will load the demo and produce a demo Buffer, and pass it to the Trainer. Else it sets to None. Then the PreTraining and GAIL modules will attempt to grab the buffer from the Trainer and throw an exception if it is None. 

But this removes the ability to set a different demo file for each module separately and makes it a bit less compartmentalized from before - I'm not sure if at the end of the day it's overall cleaner

ml-agents/mlagents/trainers/components/bc/model.py
ervteng: Added the comment

docs/Training-PPO.md
ervteng: I agree, but I also think it's necessary for people who have a huge demonstration dataset. We could do a couple of things:

- Remove the option and just set to the buffer size given by PPO - perhaps allow overriding 
- Change the option to Samples Per Update or Demonstration Buffer Size 
- Leave as-is

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
awjuliani: I think that makes sense.

docs/Training-PPO.md
awjuliani: I agree that it is useful to have. I think it just needs a different name that is a little more descriptive. "Samples Per Update" could fit the bill.

ml-agents/mlagents/trainers/components/reward_signals/gail/signal.py
```diff
@@ -0,0 +1,263 @@
...
+                        z_mean_expert.append(run_out["z_mean_expert"])
+
+                batch_losses.append(loss)
+        self.has_updated = True
+
+        if self.print_debug:
+            print_list = ["n_epoch", "beta", "policy_estimate", "expert_estimate"]
+            print_vals = [
+                n_epoch,
+                self.policy.sess.run(self.model.beta),
```
chriselion: Does this have any side effects? Should it be outside the `if self.print_debug`?

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
ervteng: Made VAIL an option. For simple stuff like Pyramids and Crawler, VAIL actually learns slower. I believe it makes learning more stable, but just from the learning speed on Pyramids and Crawler, I've decided to default to `use_vail = False`

ml-agents/mlagents/trainers/components/reward_signals/gail/signal.py
ervteng: removed self.print_debug entirely. Am now using the logging.debug() call.

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
chriselion: Not sure, utils.py seems a bit better than enforcing a bad class hierarchy.

ml-agents/mlagents/trainers/components/bc/model.py
ervteng: Yep - typo (fixed)

ml-agents/mlagents/trainers/components/bc/model.py
ervteng: Gone

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
ervteng: Will leave that for a future PR, for now changed to `swish`

docs/Training-PPO.md
ervteng: Done! (and yes any issues were related to the stochasticity)

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
ervteng: I'm also a bit reluctant to create a `utils.py` with just one function in it - perhaps we should wait until there are more common functions before refactoring this?

PR #2119, opened 2019-06-10
Doc/comment cleanup - Fix some occurrences of 'the the'

chriselion:
Some small doc and comment cleanups (also to test my github setup).

Also adds `/venv` to `.gitignore` for setting up virtual environments.

### DIFF
--- a/ml-agents/mlagents/trainers/ppo/trainer.py
+++ b/ml-agents/mlagents/trainers/ppo/trainer.py
@@ -29,7 +29,7 @@ class PPOTrainer(Trainer):
         :param training: Whether the trainer is set for training.
         :param load: Whether the model should be loaded.
         :param seed: The seed the model will be initialized with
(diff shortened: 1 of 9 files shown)


ml-agents/mlagents/trainers/ppo/trainer.py
```diff
@@ -29,7 +29,7 @@ def __init__(
         :param training: Whether the trainer is set for training.
         :param load: Whether the model should be loaded.
         :param seed: The seed the model will be initialized with
-        :param run_id: The The identifier of the current run
```
chriselion: I think this is the one I found originally, the others I spotted with `git grep -i`

xiaomaogy: Wow there are so many the the in our doc.

chriselion: Not anymore :)

PR #2124, opened 2019-06-12
use yaml.safe_load instead of yaml.load

chriselion:
This gets rid of these warnings

> /.../ml-agents/ml-agents/mlagents/trainers/learn.py:157: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
>   trainer_config = yaml.load(data_file)
> 

when running `mlagents-learn`. It's also just a generally good idea; see the link for more info.


chriselion:
Tests failing until https://github.com/Unity-Technologies/ml-agents/pull/2125 is merged in

PR #2133, opened 2019-06-14
Automate meta file check

chriselion:
Attempt to automate part of our release checklist. This ensures each .meta file corresponds to a normal file, and each normal file has a .meta file (except for the script itself).

We could potentially run this as a unit test instead, but that would mean the mlagents package would have to read outside its own directory. Standalone script and executing in CI seemed like a cleaner approach.

Also, please check that my understanding of the checklist item is correct.


UnitySDK/Assets/validate_meta_files.py
```diff
@@ -0,0 +1,42 @@
+import json
+import os
+
+
+def main():
+    asset_path = "UnitySDK/Assets"
```
xiaomaogy: This assumes that you are always running in the root path of the repo, which I guess is fine since you will only run it in the circleci.

xiaomaogy:
Overall looks good to me. The only things is that maybe we should not put this script inside the UnitySDK folder. This seems like a utils script and might worth its own folder?

UnitySDK/Assets/validate_meta_files.py
chriselion: Yeah, I guess we could add the path as a parameter if we need to later. But I think it's OK to leave hardcoded for now.

chriselion:
Sure, I'll look into moving it into a folder.

PR #2158, opened 2019-06-20
Fixed the import issue

xiaomaogy:


ervteng:
After some digging - looks like the structure we are using is the [native namespace package](https://packaging.python.org/guides/packaging-namespace-packages/#native-namespace-packages) introduced in Python 3.3. (we don't have an __init__.py in mlagents/). 

One thing we can do is replace the setup.py `packages=` with `packages=find_namespace_packages(exclude=["*.tests", "*.tests.*", "tests.*", "tests"])`. I've tested this and it does successfully find the trainers. This would mean we won't have to keep adding new packages to this list. Thoughts?

ml-agents/setup.py
```diff
@@ -23,7 +23,11 @@
         "License :: OSI Approved :: Apache Software License",
         "Programming Language :: Python :: 3.6",
     ],
-    packages=["mlagents.trainers", "mlagents.trainers.bc", "mlagents.trainers.ppo"],  # Required
+    packages=[
+        "mlagents.trainers",
+        "mlagents.trainers.bc",
+        "mlagents.trainers.ppo",
+    ],  # Required
```
xiaomaogy: Thanks~

chriselion: 👍

chriselion: I think @ervteng's solution is better though.

PR #2163, opened 2019-06-21
Use pre-commit for CI tests, add black hook

chriselion:
Set up pre-commit (https://pre-commit.com/hooks.html) with a single hook to run `black`. This is a no-op for now, but we can add other checks like mypy soon.

You can see the output from a failed CI run here: https://circleci.com/gh/Unity-Technologies/ml-agents/873?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link It's slightly nicer (IMHO) since it shows the diff of applying the linter, so you have a better idea of where the failure is.


.circleci/config.yml
```diff
@@ -42,11 +42,13 @@ jobs:
...
       - run:
           name: Check Code Style for ml-agents and gym_unity using black
+          # Some pre-commit hooks (namely mypy) don't work well on the entire repo because of the pacakage structure.
+          # So run it on all the files of the subdirectory instead.
           command: |
             . venv/bin/activate
-            black --check ml-agents
-            black --check ml-agents-envs
-            black --check gym-unity
+            git ls-files ml-agents      | xargs pre-commit run --show-diff-on-failure --files
```
chriselion: I don't love this because it makes it harder to run the same checks on CI and locally. After mypy checks are running (in another PR) I'll revisit this.

.circleci/config.yml
```diff
@@ -24,7 +24,7 @@ jobs:
             pip install --upgrade setuptools
             cd ml-agents-envs && pip install -e .
             cd ../ml-agents && pip install -e .
-            pip install black pytest-cov==2.6.1 codacy-coverage==1.3.11
+            pip install pre-commit pytest-cov==2.6.1 codacy-coverage==1.3.11
```
chriselion: pre-commit will install all the corresponding hooks that it needs, so we don't need to install black directly anymore.

.circleci/config.yml
chriselion: Figured out how to get mypy working with a single invocation, so can do this in one call now.

PR #2167, opened 2019-06-22
Develop new reset param crawler gravity length of limbs

sankalp04:


UnitySDK/Assets/ML-Agents/Examples/Crawler/Scripts/CrawlerAgent.cs
```diff
@@ -265,5 +273,49 @@ public override void AgentReset()
...
+    {
+        SetForeLegSize();
+        SetUpperLegSize();
+    }
+
+    public void SetAgentStartingHeight()
+    {
+        float newHeight = leg0Upper.localScale.y + leg0Lower.localScale.y + 2;
+
+        body.position = new Vector3(body.position.x, newHeight, body.position.z);
```
ervteng: Keep upper and lower in lists and iterate

PR #2177, opened 2019-06-24
Enable mypy in precommit checks

chriselion:
This enables mypy (python type annotation checks) for `ml-agents`, `ml-agents-env`, and `gym-unity`. The number of lines changed is a little intimidating, but it's mostly due to adding new stub files. I can break this into a PR per library though.

The repo layout added a few complications:
* We need 3 separate invocations of mypy, filtered by directory
* Currently need `--ignore-missing-imports` on each. Will try to remove those in subsequent PRs
* The generated `_pb2.py` files are flagged as incorrect because the stubs https://github.com/python/typeshed/tree/master/third_party/2and3/google/protobuf don't have the `serialized_options` field that the current code does https://github.com/protocolbuffers/protobuf/blob/master/python/google/protobuf/descriptor.py#L539. Need to see if we can fix the stubs or if they're missing for a reason.
* Because each directory is checked individually, I'm not sure that the type information from one is available in the other. Again, this can be improved in another PR.


ml-agents/mlagents/trainers/tests/test_trainer_controller.py
```diff
@@ -283,14 +283,8 @@ def test_initialize_invalid_trainer_raises_exception(BrainInfoMock):
     tc = basic_trainer_controller(brain_info_mock)
     bad_config = dummy_bad_config()
 
-    try:
+    with pytest.raises(UnityEnvironmentException):
         tc.initialize_trainers(bad_config)
-        assert (
```
chriselion: mypy correctly complained that the assert was always true. `pytest.raises` is a cleaner way to test that an exception is raised

ml-agents-envs/mlagents/envs/brain.py
```diff
@@ -174,17 +174,19 @@ def from_agent_proto(agent_info_list, brain_params):
...
 
 
-def safe_concat_lists(l1: Optional[List], l2: Optional[List]):
-    if l1 is None and l2 is None:
-        return None
-    if l1 is None and l2 is not None:
-        return l2.copy()
-    if l1 is not None and l2 is None:
-        return l1.copy()
+def safe_concat_lists(l1: Optional[List], l2: Optional[List]) -> Optional[List]:
```
chriselion: mypy was getting confused about the null-ness of the inputs. Personally I think this is a little nicer anyway.

ml-agents-envs/mlagents/envs/communicator_objects/agent_action_proto_pb2.pyi
```diff
@@ -0,0 +1,51 @@
+# @generated by generate_proto_mypy_stubs.py.  Do not edit!
```
chriselion: All of these `.pyi` files are stubs generated with https://github.com/dropbox/mypy-protobuf. I updated the directions with how to generate them.

ml-agents-envs/mlagents/envs/environment.py
```diff
@@ -373,7 +373,18 @@ def step(
         custom_action = {} if custom_action is None else custom_action
 
         # Check that environment is loaded, and episode is currently running.
-        if self._loaded and not self._global_done and self._global_done is not None:
+        if not self._loaded:
```
chriselion: It wasn't obvious that this block either raises or returns; I think this is a little cleaner.
Could also unindent the "main" block but trying to keep the diff smaller.

ml-agents-envs/mlagents/envs/environment.py
```diff
@@ -623,7 +623,7 @@ def _get_state(self, output: UnityRLOutput) -> (AllBrainInfo, bool):
 
     def _generate_step_input(
         self, vector_action, memory, text_action, value, custom_action
-    ) -> UnityRLInput:
+    ) -> UnityInput:
```
chriselion: These were just wrong :)

ml-agents-envs/mlagents/envs/environment.py
```diff
@@ -607,7 +607,7 @@ def _flatten(cls, arr) -> List[float]:
         arr = [float(x) for x in arr]
         return arr
 
-    def _get_state(self, output: UnityRLOutput) -> (AllBrainInfo, bool):
```
chriselion: Unfortunately that's not valid type annotation syntax, you have to use `Tuple[]`

chriselion:
@ervteng just a heads up. I don't think this conflicts much with your rewards refactor but there might be a bit of overlap.

.pre-commit-config.yaml
```diff
@@ -1,7 +1,23 @@
...
     rev: '19.3b0'
     hooks:
     -   id: black
+-   repo: https://github.com/pre-commit/mirrors-mypy
+    rev: 'v0.701'
+    hooks:
+    -   id: mypy
+        name: mypy-ml-agents
+        files: "ml-agents/.*"
+        args: [--ignore-missing-imports, --disallow-incomplete-defs]
```
xiaomaogy: What are --ignore-missing-imports, --disallow-incomplete-defs used for?

CONTRIBUTING.md
```diff
@@ -48,6 +48,18 @@ the platform, and provide a unique non-trivial challenge to modern
 machine learning algorithms. Feel free to submit these environments with a
 PR explaining the nature of the environment and task.
 
-## Style Guide
+## Continuous Integration (CI)
```
xiaomaogy: I think this is still a style guide, we just enforce our style guide through CI.

ml-agents-envs/mlagents/envs/communicator_objects/agent_action_proto_pb2.pyi
xiaomaogy: Why use stub files instead of putting the type annotation directly onto the python source file?

ml-agents-envs/mlagents/envs/rpc_communicator.py
```diff
@@ -94,7 +95,7 @@ def initialize(self, inputs: UnityInput) -> UnityOutput:
         self.unity_to_external.parent_conn.recv()
         return aca_param
 
-    def exchange(self, inputs: UnityInput) -> UnityOutput:
+    def exchange(self, inputs: UnityInput) -> Optional[UnityOutput]:
```
xiaomaogy: Why this is Optional here?

ml-agents/mlagents/trainers/demo_loader.py
```diff
@@ -4,7 +4,8 @@
 from mlagents.trainers.buffer import Buffer
 from mlagents.envs.brain import BrainParameters, BrainInfo
 from mlagents.envs.communicator_objects import *
-from google.protobuf.internal.decoder import _DecodeVarint32
+from google.protobuf.internal.decoder import _DecodeVarint32  # type: ignore
```
xiaomaogy: What does #type:ignore mean here?

.pre-commit-config.yaml
chriselion: `--disallow-incomplete-defs` means that the checker will complain about functions that have some arguments and return annotated, but not others. For example
``` python
def foo(x: int, y) -> int:  # bad, no type hint on y

def bar(x: int, y: int):  # bad, no return type
```
would be treated as errors. I didn't apply it to `ml-agents-env` because there's still some code that I'm not sure what the types should be.

`--ignore-missing-imports` is a little more complicated; without it the checker will complain about importing numpy and tensorflow. It also complains about some imports in the mlagents repo; I think is related to our directory structure (similar to the package problems we had during the release) but I'll admit that I don't fully understand it yet. I would like to remove these in the future but it's the only practical way to get checks passing today.

protobuf-definitions/make.bat
```diff
@@ -22,7 +22,7 @@ mkdir -p $DST_DIR_P/$PYTHON_PACKAGE
 # generate proto objects in python and C#
 
 protoc --proto_path=proto --csharp_out=$DST_DIR_C $SRC_DIR/*.proto
-protoc --proto_path=proto --python_out=$DST_DIR_P $SRC_DIR/*.proto 
+protoc --proto_path=proto --python_out=$DST_DIR_P --mypy_out=$DST_DIR_P $SRC_DIR/*.proto
```
xiaomaogy: Have you tested this?

ml-agents-envs/mlagents/envs/communicator_objects/agent_action_proto_pb2.pyi
chriselion: The python source here is generated by protoc and it's not generating traditional python classes. In theory we might be able to do it, but it would involve wrestling with both `protoc` and `mypy-protobuf`.

ml-agents-envs/mlagents/envs/rpc_communicator.py
chriselion: The method can potentially return `None` (line 105), so we have to change the return type. `Optional[UnityOutput]` is shorthand for `Union[None, UnityOutput]`

xiaomaogy: Oh I missed that part of the logic.

ml-agents/mlagents/trainers/demo_loader.py
chriselion: `_DecodeVarint32` isn't declared in the protobuf library stubs here: https://github.com/python/typeshed/blob/master/third_party/2and3/google/protobuf/internal/decoder.pyi
I'm not sure if this is an oversight, or because they don't want to provide type info for internal functions (it starts with a `_`), or because it's not available in all versions of protobuf. 

`# type: ignore` just tells mypy to ignore errors on that line; it will treat `_DecodeVarint32` as an `Any`. We can try to fix this up, but it didn't seem worth the effort.

xiaomaogy: I agree, this is good.

CONTRIBUTING.md
chriselion: There's a "Code Style" section below. I wanted to give a little more explanation about how the checks (black, mypy, and anything else we add later) were run. I can change this back and move it to another doc if you think it was better before.

ml-agents-envs/mlagents/envs/communicator_objects/agent_action_proto_pb2.pyi
xiaomaogy: Oh I see, make sense.

protobuf-definitions/make.bat
chriselion: Yes, that was how I generated the `.pyi` files.

CONTRIBUTING.md
xiaomaogy: No I think everything is fine here. I don't have a strong opinion on this.

PR #2180, opened 2019-06-24
python timers

chriselion:
Simple nested timers for profiling long-running jobs. To use, just import `hierarchical_timer` and use is as a context manager:

``` python
with hierarchical_timer("my_stuff"):
  do_stuff()
```
or
``` python
@timed # will generate a "my_func" timer
def my_func(x, y)
  return x + y

class Foo:
  @timed  # will generate a "Foo.bar" timer
  def bar()
```

Some follow-ups todo after the initial PR:
* See if we can capture what the existing TrainerMetrics is doing, and delete the old code there
* Track timings in the subprocess workers, and send stats back to the "main" process and merge them there.


chriselion:
Example results on the hallway scene for about a minute: 
``` json
{
  "total": 68.649203701,
  "count": 1,
  "children": [
    {
      "name": "take_step",
      "total": 65.83360267899994,
      "count": 2199,
      "children": [
        {
          "name": "get_actions",
          "total": 2.5110894320000035,
          "count": 2200,
          "self": 2.5110894320000035
        },
        {
          "name": "env_step",
          "total": 20.06926863800002,
          "count": 2200,
          "self": 20.06926863800002
        },
        {
          "name": "add_experiences",
          "total": 1.1026085590000263,
          "count": 2200,
          "self": 1.1026085590000263
        },
        {
          "name": "process_experiences",
          "total": 1.0049063659999633,
          "count": 2200,
          "self": 1.0049063659999633
        },
        {
          "name": "update_policy",
          "total": 39.75323067699998,
          "count": 31,
          "children": [
            {
              "name": "execute_model",
              "total": 40.25257996899976,
              "count": 1565,
              "self": 40.25257996899976
            }
          ],
          "self": 0.0
        }
      ],
      "self": 1.392499006999941
    }
  ],
  "self": 2.8156010220000667
}
```
There's something still a bit funny with the total of the children taking more time than the parent; will look into how that's being counted...

ml-agents-envs/mlagents/envs/timers.py
```diff
@@ -0,0 +1,145 @@
...
+from contextlib import contextmanager
+from typing import Any, Callable, Dict, Generator, TypeVar
+
+
+class TimerNode:
+    """
+    Represents the time spent in a block of code.
+    """
+
+    __slots__ = ["children", "total", "count"]
```
chriselion: Possibly a premature optimization, but this helps reduce the memory footprint and time overhead to the overall system.

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -296,36 +304,40 @@ def take_step(
...
         take_action_value = {}
         take_action_outputs = {}
-        for brain_name, trainer in self.trainers.items():
-            action_info = trainer.get_action(curr_info[brain_name])
-            take_action_vector[brain_name] = action_info.action
-            take_action_memories[brain_name] = action_info.memory
-            take_action_text[brain_name] = action_info.text
-            take_action_value[brain_name] = action_info.value
-            take_action_outputs[brain_name] = action_info.outputs
+        with hierarchical_timer("get_actions"):
```
chriselion: The decorator is more concise and doesn't need the whitespace change, but this is probably better for parts of the code where we're making calls to an abstract class.

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -118,6 +120,11 @@ def _write_training_metrics(self):
             if brain_name in self.trainer_metrics:
                 self.trainers[brain_name].write_training_metrics()
 
+    def _write_timing_tree(self) -> None:
+        timing_path = f"{self.summaries_dir}/{self.run_id}_timers.json"
```
chriselion: I'm open to suggestions on where to put this/what to call it. But this will put it in the same path as the current TrainerMetrics csv files.

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -310,9 +319,10 @@ def advance(self, env: SubprocessEnvManager) -> int:
                 if changed:
                     self.trainers[brain_name].reward_buffer.clear()
 
-        time_start_step = time()
-        new_step_infos = env.step()
-        delta_time_step = time() - time_start_step
+        with hierarchical_timer("env_step"):
```
chriselion: The decorator is more concise and doesn't need the whitespace change, but this is probably better for parts of the code where we're making calls to an abstract class.

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -333,6 +343,7 @@ def advance(self, env: SubprocessEnvManager) -> int:
                 trainer.increment_step(len(new_step_infos))
                 if trainer.is_ready_update():
                     # Perform gradient descent with experience buffer
-                    trainer.update_policy()
+                    with hierarchical_timer("update_policy"):
```
chriselion: The decorator is more concise and doesn't need the whitespace change, but this is probably better for parts of the code where we're making calls to an abstract class.

ml-agents-envs/mlagents/envs/timers.py
```diff
@@ -0,0 +1,172 @@
...
+        return res
+
+
+# Global instance of a TimerStack. This is generally all that we need for profiling, but you can potentially
+# create multiple instances and pass them to the contextmanager
+_global_timer_stack = TimerStack()
+
+
+@contextmanager
+def hierarchical_timer(name: str, timer_stack: TimerStack = None) -> Generator:
```
chriselion: Should this just be called "timer" or is that too generic?

ml-agents-envs/mlagents/envs/timers.py
```diff
@@ -0,0 +1,172 @@
+# # Unity ML-Agents Toolkit
```
chriselion: Better suggestions on where this should live?

chriselion:
based on the current PR, here's how hallway looks today:
``` python
{
  "total": 0.0,
  "count": 0,
  "self": 0.0,
  "children": [
    {
      "name": "TrainerController.advance",
      "total": 32.731460376,
      "count": 957,
      "self": 1.4661682279999795,
      "children": [
        {
          "name": "env_step",
          "total": 14.123654692000024,
          "count": 957,
          "self": 0.1792060960000832,
          "children": [
            {
              "name": "SubprocessEnvManager._take_step",
              "total": 1.2891531169999557,
              "count": 957,
              "self": 0.05311957899995745,
              "children": [
                {
                  "name": "PPOPolicy.evaluate",
                  "total": 1.2360335379999983,
                  "count": 957,
                  "self": 1.2360335379999983
                }
              ]
            },
            {
              "name": "recv",
              "total": 12.655295478999985,
              "count": 957,
              "self": 12.655295478999985
            }
          ]
        },
        {
          "name": "update_policy",
          "total": 17.141637455999998,
          "count": 14,
          "self": 0.13363037500003472,
          "children": [
            {
              "name": "PPOPolicy.update",
              "total": 17.008007080999963,
              "count": 678,
              "self": 17.008007080999963
            }
          ]
        }
      ]
    }
  ]
}
```

ml-agents-envs/mlagents/envs/timers.py
xiaomaogy: Wow I didn't know this exists. This will be super useful once we use the profiling in a tons of places.

ml-agents-envs/mlagents/envs/timers.py
xiaomaogy: I think this is good, better to be precise to help people understand.

ml-agents-envs/mlagents/envs/timers.py
xiaomaogy: This seems good to me for the short term.

ml-agents-envs/mlagents/envs/timers.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+
+    __slots__ = ["children", "total", "count"]
+
+    def __init__(self):
+        self.children: Dict[str, "TimerNode"] = {}
+        self.total: float = 0.0
+        self.count: int = 0
+
+    def get_child(self, name: str) -> "TimerNode":
```
xiaomaogy: Maybe this is a dumb question.. Why do you make the TimeNode into "TimeNode" here?

chriselion: This explains it better than I can: https://mypy.readthedocs.io/en/latest/kinds_of_types.html#class-name-forward-references

The runtime can't parse without the string: ```
______________________________________ ERROR collecting mlagents/trainers/tests/test_reward_signals.py ______________________________________
venv/lib/python3.7/site-packages/py/_path/local.py:701: in pyimport
    __import__(modname)
ml-agents/mlagents/trainers/__init__.py:6: in <module>
    from .trainer import *
ml-agents/mlagents/trainers/trainer.py:8: in <module>
    from mlagents.trainers import TrainerMetrics
ml-agents/mlagents/trainers/__init__.py:8: in <module>
    from .trainer_controller import *
ml-agents/mlagents/trainers/trainer_controller.py:16: in <module>
    from mlagents.envs.subprocess_env_manager import SubprocessEnvManager
ml-agents-envs/mlagents/envs/subprocess_env_manager.py:9: in <module>
    from mlagents.envs.timers import timed, hierarchical_timer
ml-agents-envs/mlagents/envs/timers.py:35: in <module>
    class TimerNode:
ml-agents-envs/mlagents/envs/timers.py:47: in TimerNode
    def get_child(self, name: str) -> TimerNode:
E   NameError: name 'TimerNode' is not defined
```
It looks like they're not needed in `__init__` though, so I'll remove those ones.

File: ml-agents-envs/mlagents/envs/timers.py [hunk 9]
```diff
@@ -0,0 +1,172 @@
...
+            node = self.root
+
+        res: Dict[str, Any] = {"total": node.total, "count": node.count}
+
+        child_total = 0.0
+        child_list = []
+        for child_name, child_node in node.children.items():
+            child_res: Dict[str, Any] = {
+                "name": child_name,
+                **self.get_timing_tree(child_node),
```
xiaomaogy: What is the ** before self used for here?

ml-agents-envs/mlagents/envs/timers.py
```diff
@@ -0,0 +1,172 @@
...
+            }
+            child_list.append(child_res)
+            child_total += child_res["total"]
+
+        # "self" time is total time minus all time spent on children
+        res["self"] = max(0.0, node.total - child_total)
+        if child_list:
+            res["children"] = child_list
+
+        return res
```
xiaomaogy: It seems that based on this implementation, the root TimerNode is treated as a special node (its total, count is always 0, it doesn't have a name). Since this is a recursive call, should we set a base case so that the root TimerNode also has the corresponding values?

chriselion: I agree, the handling of the root node is a little strange but I'm not sure what the best way to format the output is. A few ways I was thinking about:
* Omit total, self, and count keys from the root
* Just return the list of the root's children
* For the root, set total = program execution time, self = total - children, and count = 1
Sounds like you're suggesting the third one? Any other ways to do it?

xiaomaogy: Yes I'm suggesting the third one. Also I'm suggesting adding a "name" attribute to the TimerNode class, and give the root TimerNode a key value pair as "name":"root". Right now you are adding the "name" attribute to the child TimerNode in the line 104, which was a little bit confusing.

ml-agents-envs/mlagents/envs/timers.py
```diff
@@ -0,0 +1,172 @@
...
+
+This would produce a timer tree like
+    (root)
+        "foo"
+        "context"
+            "foo"
+
+The total time and counts are tracked for each block of code; in this example "foo" and "context.foo" are considered
+distinct blocks, and are tracked separately.
+"""
```
xiaomaogy: Maybe it is helpful to point out(or give an example) that @timed is totally equivalent to "with hierarchical_timer()" given a certain condition (when function name equals the name passed in). So that for people who are not familiar with Decorator or doesn't want to go to the implementation, it is easier to understand?

xiaomaogy:
Overall looks good to me. I've learned a lot during the review.

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -284,8 +296,10 @@ def start_learning(
         env_manager.close()
         if self.train_model:
             self._write_training_metrics()
+            self._write_timing_tree()
```
ervteng: We might want to write the timing tree even when we're not training (i.e., doing inference). I think if we just move it out of this if statement it will do that..

ml-agents-envs/mlagents/envs/timers.py
chriselion: Sounds good, I'll do that (and I forgot about adding `"name":"root"`)

I'll add some comments about the node names - since the parent has the name as the key, the node doesn't actually store it.

ml-agents/mlagents/trainers/trainer_controller.py
chriselion: Done

PR #2198, opened 2019-06-26
C# hierarchical timers

chriselion:
Hierarchical timers for C#.

Sample output (after running through https://jsonlint.com/):
```json
{
	"count": 1,
	"self": 10.327180799999999,
	"total": 10.734703,
	"children": {
		"AgentResetIfDone": {
			"count": 525,
			"self": 0.030199999999999998,
			"total": 0.030199999999999998,
			"children": null
		},
		"AgentSendState": {
			"count": 525,
			"self": 0.071646,
			"total": 0.120143,
			"children": {
				"CollectObservations": {
					"count": 1260,
					"self": 0.048497,
					"total": 0.048497,
					"children": null
				}
			}
		},
		"BrainDecideAction": {
			"count": 525,
			"self": 0.195094,
			"total": 0.195094,
			"children": null
		},
		"AcademyStep": {
			"count": 525,
			"self": 0.019004,
			"total": 0.019004,
			"children": null
		},
		"AgentAct": {
			"count": 525,
			"self": 0.041696,
			"total": 0.041696,
			"children": null
		}
	}
}
```


UnitySDK/Assets/ML-Agents/Scripts/Academy.cs
```diff
@@ -551,18 +556,40 @@ void EnvironmentStep()
...
+                AgentAct();
+            }
 
             m_StepCount += 1;
             m_TotalStepCount += 1;
+
+            if (m_TotalStepCount % 100 == 0)
+            {
+                // TODO better place to hook this in?
+                // TODO append environment ID so multiple envs don't overwrite each other
```
chriselion: Talked to @harper-u3d about this; might pass the worker ID as an env variable.

UnitySDK/Assets/ML-Agents/Scripts/Timer.cs
```diff
@@ -0,0 +1,257 @@
...
+#if UNITY_EDITOR
+using UnityEditor;
+
+#endif
+
+
+namespace MLAgents
+{
+
+    [DataContract]
```
chriselion: I used the generic C# DataContract/DataMember for JSON serialization. Could also use https://docs.unity3d.com/Manual/JSONSerialization.html - this happens very rarely so I'm not too worried about performance.

ml-agents-envs/mlagents/envs/environment.py
```diff
@@ -583,7 +583,15 @@ def _close(self):
         self._loaded = False
         self.communicator.close()
         if self.proc1 is not None:
-            self.proc1.kill()
+            # Wait a few seconds for the process to shutdown, but kill it if it takes too long
+            try:
```
chriselion: This seems necessary (but not sufficient) to the environment cleanly shutting down. Might move to another PR.

UnitySDK/Assets/ML-Agents/Scripts/Timer.cs
surfnerd: I would lean towards using Unity classes.

chriselion: Tried this as a quick proof of concept. I don't think the Unity serializer can handle recursively defined classes, even if they don't loop in practice.

I made an output class (to help match the format of python's output):
```csharp
    [Serializable]
    public class TimerOutput
    {
        public string name = "";
        public double total;
        public double self;
        public int num_calls;
        List<TimerOutput> children;

        public TimerOutput(TimerNode node)
        {
            total = node.TotalSeconds;
            self = node.SelfSeconds;
            num_calls = node.m_NumCalls;
            // note children not assigned
        }
    }
```
and used it with 
```csharp
var output = new TimerOutput(m_RootNode);
var json = JsonUtility.ToJson(output, prettyPrint: true);
```
and get this
> Serialization depth limit 7 exceeded at 'MLAgents::TimerOutput.children'. There may be an object composition cycle in one or more of your serialized classes.
and it doesn't output the `children` list.

UnitySDK/Assets/ML-Agents/Scripts/Academy.cs
```diff
@@ -594,6 +614,9 @@ protected virtual void OnDestroy()
 
             // Signal to listeners that the academy is being destroyed now
             DestroyAction();
+
+            m_Timer.SaveJsonTimers();
```
chriselion: This does not currently trigger when learn.py closes the environment. Need to debug that further, but will do so in a separate PR.

UnitySDK/Assets/ML-Agents/Scripts/Timer.cs
surfnerd: oof

UnitySDK/Assets/ML-Agents/Scripts/Timer.cs
```diff
@@ -0,0 +1,318 @@
...
+    ///             doSomeMoreWork();
+    ///         }
+    ///     }
+    /// }
+    /// </summary>
+    /// <remarks>
+    /// This implements the Singleton pattern (solution 4) as described in
+    /// https://csharpindepth.com/articles/singleton
+    /// </remarks>
+    public class TimerStack : System.IDisposable
```
chriselion: Should we just call this `Timer`?

UnitySDK/Assets/ML-Agents/Scripts/Academy.cs
chriselion: Fixed in another PR.

UnitySDK/Assets/ML-Agents/Scripts/Timer.cs
```diff
@@ -0,0 +1,315 @@
...
+
+
+namespace MLAgents
+{
+
+    [DataContract]
+    public class TimerNode
+    {
+        static string s_Separator = ".";
+        static double s_TicksToSeconds = 1.0 / 10000000.0; // 100 ns per tick
```
surfnerd: could you change to 1e-7?

UnitySDK/Assets/ML-Agents/Scripts/Timer.cs
surfnerd: TimerStack seems appropriate.

UnitySDK/Assets/ML-Agents/Scripts/Timer.cs
chriselion: Done

UnitySDK/Assets/ML-Agents/Scripts/Timer.cs
```diff
@@ -53,13 +53,31 @@ public class TimerNode
...
+        /// if the timer is still running (i.e. if m_TickStart is non-zero).
+        /// </summary>
+        public long CurrentTicks
+        {
+            get
+            {
+                long currentTicks = m_TotalTicks;
+                if (m_TickStart != 0)
+                {
+                    currentTicks += (System.DateTime.Now.Ticks - m_TickStart);
```
surfnerd: isn't this equivalent to `2 * m_TotalTicks + (System.DateTime.Now.Ticks - m_TotalTicks)`?

surfnerd: nvm, misunderstood.

UnitySDK/Assets/ML-Agents/Scripts/Academy.cs
surfnerd: makes sense.

chriselion:
@vincentpierre Going to merge this now. Agreed it doesn't feel like it belongs with the  "core" ML-Agents code, but we can easily move it later if you have a suggestion of where it should go.

PR #2207, opened 2019-07-01
Update issue templates

harperj:
Added three issue template types:
* bug reports
* feature requests
* discussion and general questions

The focus is on getting sufficient information in bug requests to help us reproduce the issue and communicating that we will be unable to generally support custom environments and participate in general discussions.

### DIFF
--- a/.github/ISSUE_TEMPLATE/discussion---general-questions.md
+++ b/.github/ISSUE_TEMPLATE/discussion---general-questions.md
@@ -1,0 +1,12 @@
+---
+name: Discussion / General Questions
+about: Discussion about ML-Agents, RL algorithms, or game integrations
+title: ''
+labels: discussion
+assignees: ''
+
+---
+
+Describe what you'd like to discuss.
+
+**Note**: The ML-Agents team has limited resources for education and community discussion.  We'll participate as we are able but encourage members of the community to support one another to discuss and support one another.
--- a/.github/ISSUE_TEMPLATE/discussion---general-questions.md
+++ b/.github/ISSUE_TEMPLATE/discussion---general-questions.md
@@ -9,4 +9,4 @@ assignees: ''
 
 Describe what you'd like to discuss.
 
-**Note**: The ML-Agents team has limited resources for education and community discussion.  We'll participate as we are able but encourage members of the community to support one another to discuss and support one another.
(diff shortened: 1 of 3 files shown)


chriselion:
I think these need to be on master, not develop:

> Issue templates are stored on the repository's default branch, in a hidden .github/ISSUE_TEMPLATE directory. If you create a template in another branch, it will not be available for collaborators to use. 

https://help.github.com/en/articles/about-issue-and-pull-request-templates

.github/ISSUE_TEMPLATE/discussion---general-questions.md
```diff
@@ -0,0 +1,12 @@
...
+about: Discussion about ML-Agents, RL algorithms, or game integrations
+title: ''
+labels: discussion
+assignees: ''
+
+---
+
+Describe what you'd like to discuss.
+
+**Note**: The ML-Agents team has limited resources for education and community discussion.  We'll participate as we are able but encourage members of the community to support one another to discuss and support one another.
```
chriselion: grammar nit: "able, but"

harperj:
@chriselion Yes, I was thinking about the branch issue -- I guess I picked develop by default since this is our usual git workflow.  Shouldn't be a problem to merge this type of change directly into master; I'll change the branch.

.github/ISSUE_TEMPLATE/discussion---general-questions.md
harperj: 👍 -- fixed.

xiaomaogy:
Looks good to me. One comment is that we also have another big category of issue for help-wanted. Do we plan to remove that category?

harperj:
@xiaomaogy my suggestion here is to merge the `help-wanted` and `discussion` categories.  The idea being that both of these are outside of the scope of what we can guarantee to support as part of our on-call process.

In "help wanted" issues we're often getting questions about technical details of RL algorithms or about how specific bits of code work.  This often involves a lot of education which seems outside of the scope of us supporting this project.  I think it would be valuable for us to help when we are able but we should lean on the community for it in general and not guarantee it.

This is just my thinking though, definitely open to adding a separate template and rethinking how we want to support those issues.

xiaomaogy:
I agree with you on this. We are currently treating them in the same way. If there is no objection on this, let's merge it.

PR #2208, opened 2019-07-01
ensure LF line endings

chriselion:
Enforce LF line endings with pre-commit. We were mostly compliant except for a barracuda file.


.pre-commit-config.yaml
```diff
@@ -21,3 +21,13 @@ repos:
...
         args: [--ignore-missing-imports, --disallow-incomplete-defs]
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.2.3
+    hooks:
+    -   id: mixed-line-ending
+        exclude: >
+            (?x)^(
+                .*cs.meta|
+                .*.css
+            )$
```
xiaomaogy: What does this part of exclude do?

UnitySDK/Assets/ML-Agents/Scripts/InferenceBrain/TensorGenerator.cs
```diff
@@ -1,106 +1,106 @@
...
-                {
-                    throw new UnityAgentsException(
-                        "Unknow tensor expected as input : " + tensor.Name);
-                }
-                _dict[tensor.Name].Generate(tensor, currentBatchSize, agentInfos);
-            }
-        }
-    }
-}
+﻿#define ENABLE_BARRACUDA
```
xiaomaogy: Why this script get changed?

xiaomaogy: Oh I see, just saw your comment.

.pre-commit-config.yaml
chriselion: I excluded them because I wasn't sure if we want to change them. The meta files are autogenerated by the editor, and docs/doxygen/unity.css was changed but wasn't sure if we wanted to bother with it.

UnitySDK/Assets/ML-Agents/Scripts/InferenceBrain/TensorGenerator.cs
chriselion: Yeah, it was using different line endings. You can change the diff settings to ignore whitespace and they're identical: 
image

PR #2209, opened 2019-07-01
Add environment manager for parallel environments

harperj:
Previously in v0.8 we added parallel environments via the
SubprocessUnityEnvironment, which exposed the same abstraction as
UnityEnvironment while actually wrapping many parallel environments
via subprocesses.

Wrapping many environments with the same interface as a single
environment had some downsides, however:
* Ordering needed to be preserved for agents across different envs,
  complicating the SubprocessEnvironment logic
* Asynchronous environments with steps taken out of sync with the
  trainer aren't viable with the Environment abstraction

This PR introduces a new EnvManager abstraction which exposes a
reduced subset of the UnityEnvironment abstraction and a
SubprocessEnvManager implementation which replaces the
SubprocessUnityEnvironment.


harperj:
Apologies in advance for the size of this PR.  Many of the changed lines are specifically around renaming files Policy / TFPolicy or otherwise moving around code.  If it would make it a lot easier for reviewers I can attempt to stage this better.

ml-agents-envs/mlagents/envs/brain.py
```diff
@@ -84,28 +84,28 @@ def merge_memories(m1, m2, agents1, agents2):
         return np.append(m1, m2, axis=0)
 
     @staticmethod
-    def process_pixels(image_bytes, gray_scale):
+    def process_pixels(image_bytes, gray_scale) -> np.ndarray:
```
chriselion: Can you add type annotations for the input args too?

ml-agents-envs/mlagents/envs/brain.py
```diff
@@ -157,13 +157,14 @@ def from_agent_proto(agent_info_list, brain_params):
             vector_obs = np.nan_to_num(
                 np.array([x.stacked_vector_observation for x in agent_info_list])
             )
+        agents = [str(worker_id) + "-" + str(x.id) for x in agent_info_list]
```
chriselion: nit: how about `[f'{worker_id}-{x.id}' for x in agent_info_list]`?

ml-agents-envs/mlagents/envs/env_manager.py
```diff
@@ -0,0 +1,38 @@
...
+    last_all_brain_info: Optional[AllBrainInfo]
+    current_all_brain_info: AllBrainInfo
+    all_action_infos: Optional[Dict[str, ActionInfo]]
+
+
+class EnvManager(ABC):
+    def __init__(self):
+        self.policies: Dict[str, Policy] = {}
+
+    def set_policy(self, brain_name: str, policy: Policy):
```
chriselion: return type annotation (`-> None:`)

ml-agents-envs/mlagents/envs/env_manager.py
```diff
@@ -0,0 +1,38 @@
...
+    def reset(self, config=None, train_mode=True) -> List[StepInfo]:
+        pass
+
+    @abstractmethod
+    def external_brains(self) -> Dict[str, BrainParameters]:
+        pass
+
+    @property
+    @abstractmethod
+    def reset_parameters(self) -> Dict[str, str]:
```
chriselion: This is probably a `Dict[str, Any]` right?

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
```diff
@@ -0,0 +1,179 @@
...
+                    outputs = {}
+                    for brain_name, action_info in all_action_info.items():
+                        actions[brain_name] = action_info.action
+                        memories[brain_name] = action_info.memory
+                        texts[brain_name] = action_info.text
+                        values[brain_name] = action_info.value
+                        outputs[brain_name] = action_info.outputs
+                    all_brain_info = env.step(actions, memories, texts, values)
+                _send_response("step", all_brain_info)
+            elif cmd.name == "external_brains":
```
chriselion: Is it worth moving these commands to an enum? (maybe not)

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
```diff
@@ -0,0 +1,179 @@
...
+    process: Process
+    worker_id: int
+    conn: Connection
+
+    def send(self, name: str, payload=None):
+        try:
+            cmd = EnvironmentCommand(name, payload)
+            self.conn.send(cmd)
+        except (BrokenPipeError, EOFError):
+            raise KeyboardInterrupt
```
chriselion: I understand this is probably reproducing the old behavior, but converting these to a KeyboardInterrupt feels strange. Maybe convert all three to a new EnvironmentDoneException (or something like that)?

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
```diff
@@ -0,0 +1,179 @@
...
+        env.close()
+
+
+class SubprocessEnvManager(EnvManager):
+    def __init__(
+        self, env_factory: Callable[[int], BaseUnityEnvironment], n_env: int = 1
+    ):
+        super().__init__()
+        self.envs: List[UnityEnvWorker] = []
+        for worker_id in range(n_env):
```
chriselion: If these are contiguous, would it be better to refer to that as "worker_index" or "worker_idx" instead?

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
```diff
@@ -0,0 +1,179 @@
...
+            self.envs.append(self.create_worker(worker_id, env_factory))
+        self.env_last_steps: List[StepInfo] = [None for _ in self.envs]
```
chriselion: nit: `[None] * n_env`?

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -233,40 +231,55 @@ def start_learning(
...
-        env.close()
+        env_manager.close()
         if self.train_model:
             self._write_training_metrics()
             self._export_graph()
 
-    def take_step(
-        self, env: BaseUnityEnvironment, curr_info: AllBrainInfo
-    ) -> AllBrainInfo:
+    def take_step(self, env: SubprocessEnvManager) -> int:
```
chriselion: Maybe rename to "advance" since it might take multiple steps?

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -233,40 +231,55 @@ def start_learning(
...
+                for i in range(n_steps):
+                    global_step += 1
+                    if (
+                        global_step % self.save_freq == 0
+                        and global_step != 0
+                        and self.train_model
+                    ):
+                        # Save Tensorflow model
+                        self._save_model(steps=global_step)
+                    for brain_name, trainer in self.trainers.items():
```
chriselion: Move this block to a `write_to_tensorboard()` method?

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -233,40 +231,55 @@ def start_learning(
...
-                    and self.global_step != 0
-                    and self.train_model
-                ):
-                    # Save Tensorflow model
-                    self._save_model(steps=self.global_step)
-                curr_info = new_info
+                n_steps = self.take_step(env_manager)
+                for i in range(n_steps):
+                    global_step += 1
+                    if (
```
chriselion: Maybe encapsulate this as `_should_save_model()`?

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -283,61 +296,40 @@ def take_step(
...
-                    lesson_num=self.meta_curriculum.brains_to_curriculums[
-                        brain_name
-                    ].lesson_num,
-                )
-            else:
-                trainer.write_summary(self.global_step, delta_train_start)
-            if self.train_model and trainer.get_step <= trainer.get_max_steps:
-                trainer.increment_step()
-        return new_info
+                env.set_policy(brain_name, trainer.policy)
```
chriselion: Why do you need to set the policies here?

ml-agents-envs/mlagents/envs/env_manager.py
```diff
@@ -0,0 +1,38 @@
+from abc import ABC, abstractmethod
+from typing import List, Dict, NamedTuple, Optional
+from mlagents.envs import AllBrainInfo, BrainParameters, Policy, ActionInfo
+
+
+class StepInfo(NamedTuple):
+    last_all_brain_info: Optional[AllBrainInfo]
+    current_all_brain_info: AllBrainInfo
+    all_action_infos: Optional[Dict[str, ActionInfo]]
```
chriselion: Can you add a comment here on what the key is? brain name? Alternatively, naming it something like `action_info_by_brain_name` or `brain_name_to_action_info` would make it clearer.

ml-agents-envs/mlagents/envs/env_manager.py
```diff
@@ -0,0 +1,38 @@
+from abc import ABC, abstractmethod
+from typing import List, Dict, NamedTuple, Optional
+from mlagents.envs import AllBrainInfo, BrainParameters, Policy, ActionInfo
+
+
+class StepInfo(NamedTuple):
+    last_all_brain_info: Optional[AllBrainInfo]
```
chriselion: "previous" instead of "last"? I think that's more consistent with other variables in the repo.

ml-agents-envs/mlagents/envs/env_manager.py
```diff
@@ -0,0 +1,38 @@
+from abc import ABC, abstractmethod
+from typing import List, Dict, NamedTuple, Optional
+from mlagents.envs import AllBrainInfo, BrainParameters, Policy, ActionInfo
+
+
+class StepInfo(NamedTuple):
```
chriselion: Just an idea: instead of having 3 dictionaries all keyed by the brain name, what about something like
```
class BrainStepInfo(NamedTuple):
  previous_brain_info: Optional[BrainInfo]
  current_brain_info: Optional[BrainInfo]
  action_info: Optional[ActionInfo]

StepInfo = Dict[str, BrainStepInfo] # or make it a class/NamedTuple
```
Not sure if that would help clean anything up.

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
```diff
@@ -0,0 +1,179 @@
...
+                    memories = {}
+                    texts = {}
+                    values = {}
+                    outputs = {}
+                    for brain_name, action_info in all_action_info.items():
+                        actions[brain_name] = action_info.action
+                        memories[brain_name] = action_info.memory
+                        texts[brain_name] = action_info.text
+                        values[brain_name] = action_info.value
+                        outputs[brain_name] = action_info.outputs
```
chriselion: outputs aren't actually used - is that intentional?

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
```diff
@@ -0,0 +1,179 @@
...
+    finally:
+        env.close()
+
+
+class SubprocessEnvManager(EnvManager):
+    def __init__(
+        self, env_factory: Callable[[int], BaseUnityEnvironment], n_env: int = 1
+    ):
+        super().__init__()
+        self.envs: List[UnityEnvWorker] = []
```
chriselion: nit: rename to `env_workers` or just `workers`? This implies they're `Environment`s but they aren't

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
chriselion: I think it might be cleaner to make `last_step` and `last_action_info` part of `UnityEnvWorker` instead? If mutability is the issue, we can use a dataclass and/or attrs. But that seems better than keeping these three lists.

ml-agents-envs/mlagents/envs/brain.py
harperj: 👍

ml-agents-envs/mlagents/envs/env_manager.py
harperj: This was essentially trying to minimize the changes elsewhere since a some parts of the codebase expect `AllBrainInfo`.  What do you think about leaving it for now and deprecating `AllBrainInfo` and splitting this into an `AgentStep` and `EnvironmentStep` in a following PR?

ml-agents-envs/mlagents/envs/env_manager.py
harperj: 👍 -- I think changing the name is the clearest thing

ml-agents-envs/mlagents/envs/env_manager.py
harperj: Actually it looks like this is required to be a `Dict[string, float]` by the protobuf definition: https://github.com/Unity-Technologies/ml-agents/blob/master/protobuf-definitions/proto/mlagents/envs/communicator_objects/environment_parameters_proto.proto

I'll change it to reflect that.

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
harperj: I agree (and should have done it to begin with) -- but this PR has grown a bit large.  Mind if I submit this as a subsequent PR?

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
harperj: 👍 it is -- I just forgot to remove that when moving code.  Will remove.

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
harperj: My current thinking is that this solution is fine unless the same values are used elsewhere.

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
harperj: Agreed, that should be cleaner.  Will fix.

ml-agents/mlagents/trainers/trainer_controller.py
harperj: 😁 I actually have a branch where I did that already.  I don't see a problem in fitting it into this PR

ml-agents/mlagents/trainers/trainer_controller.py
harperj: 👍

ml-agents/mlagents/trainers/trainer_controller.py
harperj: For the single-machine case you don't need to.  However from the perspective of the TrainerController any modification of the policy should result in a `set_policy` on the EnvManager.  In the distributed case, `set_policy` might actually be sending the serialized policy to another machine so this is a way of triggering that update.

Another way to do this would be to have a listener on policy updates, but this seemed simpler to me for this use case.

harperj:
Addressed most of your review comments @chriselion -- the change of `env_last_steps` and `last_action_info` was bigger so I haven't made the change yet.  Considering whether to move that into the next PR.  Also pushed back on a couple of points; PTAL

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
chriselion: OK, fair enough.

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
chriselion: Sounds good

ml-agents/mlagents/trainers/trainer_controller.py
chriselion: Sounds good. And I guess theoretically a trainer could completely change its policy between steps.

ml-agents-envs/mlagents/envs/env_manager.py
chriselion: Sure

chriselion:
Generally looks good. I'll give it one more look over tomorrow morning.

ml-agents-envs/mlagents/envs/tests/test_subprocess_env_manager.py
```diff
@@ -0,0 +1,107 @@
...
+class MockEnvWorker:
+    def __init__(self, worker_id, resp=None):
+        self.worker_id = worker_id
+        self.process = None
+        self.conn = None
+        self.send = Mock()
+        self.recv = Mock(return_value=resp)
+
+
+class SubprocessEnvManagerTest(unittest.TestCase):
```
chriselion: Not a huge deal, but you don't really need unittest.TestCase here - you're not using e.g. TestCase.setUp or tearDown, and you should be able to replace all the `self.assertEqual(x, y)` with `assert x == y`

ml-agents-envs/mlagents/envs/tests/test_subprocess_env_manager.py
```diff
@@ -0,0 +1,107 @@
...
+    def test_environments_are_created(self):
+        SubprocessEnvManager.create_worker = MagicMock()
+        env = SubprocessEnvManager(mock_env_factory, 2)
+        # Creates two processes
+        self.assertEqual(
```
chriselion: nit: can replace this with `env.create_worker.assert_has_calls([mock.call(0, mock_env_factory), mock.call(1, mock_env_factory)])`

ml-agents-envs/mlagents/envs/tests/test_subprocess_env_manager.py
```diff
@@ -0,0 +1,107 @@
...
+        env_mock.reset = Mock(return_value="reset_data")
+        env_mock.global_done = True
+
+        def mock_global_done_env_factory(worker_id: int):
+            return env_mock
+
+        mock_parent_connection = Mock()
+        step_command = EnvironmentCommand("step", (None, None, None, None))
+        close_command = EnvironmentCommand("close")
+        mock_parent_connection.recv = Mock()
```
chriselion: This line is actually redundant, Mock will return a Mock for any attribute you access.

ml-agents/mlagents/trainers/tests/test_trainer_controller.py
```diff
@@ -391,38 +393,26 @@ def trainer_controller_with_take_step_mocks():
...
-    brain_info_mock = MagicMock()
-    curr_info_mock.__getitem__ = MagicMock(return_value=brain_info_mock)
+    old_step_info = StepInfo(Mock(), Mock(), MagicMock())
+    new_step_info = StepInfo(Mock(), Mock(), MagicMock())
     trainer_mock.is_ready_update = MagicMock(return_value=True)
 
     env_mock = MagicMock()
-    env_step_output_mock = MagicMock()
-    env_mock.step = MagicMock(return_value=env_step_output_mock)
+    env_mock.step = MagicMock(return_value=[new_step_info])
```
chriselion: Similarly, I think you can tighten this up a bit to
```
    env_mock = MagicMock()
    env_mock.step.return_value = [new_step_info]
    env_mock.reset.return_value = [old_step_info]
    env_mock.global_done = False
```

ml-agents/mlagents/trainers/trainer.py
```diff
@@ -208,6 +196,8 @@ def write_summary(self, global_step, delta_train_start, lesson_num=0):
                 if self.is_training and self.get_step <= self.get_max_steps
                 else "Not Training."
             )
+            # step = min(self.get_step, self.get_max_steps)
```
chriselion: Do you still need to take the min here?

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -217,8 +215,34 @@ def _reset_env(self, env: BaseUnityEnvironment) -> AllBrainInfo:
         else:
             return env.reset(train_mode=self.fast_simulation)
 
+    def _should_save_model(self, global_step):
```
chriselion: type annotation
```suggestion
    def _should_save_model(self, global_step: int) -> bool:
```

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -217,8 +215,34 @@ def _reset_env(self, env: BaseUnityEnvironment) -> AllBrainInfo:
...
+        return (
+            global_step % self.save_freq == 0 and global_step != 0 and self.train_model
+        )
+
+    def _not_done_training(self):
```
chriselion:
```suggestion
    def _not_done_training(self) -> bool:
```

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -217,8 +215,34 @@ def _reset_env(self, env: BaseUnityEnvironment) -> AllBrainInfo:
...
+        return (
+            any([t.get_step <= t.get_max_steps for k, t in self.trainers.items()])
+            or not self.train_model
+        )
+
+    def write_to_tensorboard(self, global_step):
```
chriselion:
```suggestion
    def write_to_tensorboard(self, global_step: int) -> None:
```

ml-agents/mlagents/trainers/trainer.py
harperj: This was part of a set of changes around how we use steps that I think can / should be reverted.  Will fix.

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
```diff
@@ -0,0 +1,180 @@
...
+        return UnityEnvWorker(child_process, worker_id, parent_conn)
+
+    def step(self) -> List[StepInfo]:
+        for env_worker in self.env_workers:
+            all_action_info = self._take_step(env_worker.previous_step)
+            env_worker.previous_all_action_info = all_action_info
+            env_worker.send("step", all_action_info)
+
+        step_brain_infos: List[AllBrainInfo] = [
+            self.env_workers[i].recv().payload for i in range(len(self.env_workers))
```
ervteng: (nit) Shouldn't this be `[worker.recv().payload for worker in self.env_workers]`

harperj: @ervteng yes, that would be simpler.  This code will change anyway in the next PR so I'll make sure any cases like this are taken care of.

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -283,61 +303,36 @@ def take_step(
...
-            text_action=take_action_text,
-            value=take_action_value,
-        )
+        new_step_infos = env.step()
         delta_time_step = time() - time_start_step
+
+        for step_info in new_step_infos:
+            for brain_name, trainer in self.trainers.items():
+                if brain_name in self.trainer_metrics:
+                    self.trainer_metrics[brain_name].add_delta_step(delta_time_step)
```
chriselion: Note - I think you're double counting the delta_time_step. But I also think the TrainerMetrics code is broken because add_delta_step() increments the experience collection counter 🤦‍♂  https://github.com/Unity-Technologies/ml-agents/blob/ea60bc9089daa6064763463931ef6b4843d2ac4c/ml-agents/mlagents/trainers/trainer_metrics.py#L59-L66
Should be able to clean this up with the timer code soon.

PR #2210, opened 2019-07-02
Develop generalization training

sankalp04:
[Work in progress] - Need to include doc changes and tests


ml-agents-envs/mlagents/envs/sampler_class.py
```diff
@@ -0,0 +1,112 @@
...
+
+class Sampler(ABC): 
+
+    @abstractmethod
+    def sample_parameter(self, *args, **kwargs):
+        pass
+
+
+class UniformSampler(Sampler):
+    def __init__(self, param_dict):
```
chriselion: Make the min and max value explicit arguments to `__init__`, then do `sampler_cls(**param_dict)` in the factory instead.

chriselion: I think all of the samplers should either take a random seed, or an instance of a random.Random (or the numpy equivalent). That way you at least have the option of replaying a particular sequence of samples.

ml-agents-envs/mlagents/envs/sampler_class.py
```diff
@@ -0,0 +1,112 @@
...
+            if key not in param_dict:
+                raise SamplerException(
+                    key + " is required to use Uniform Sampling but" 
+                    " this value isn't provided. Please recheck"
+                    " the documentation to ensure that the sampler file"
+                    " matches the required format."
+                )
+
+    def sample_parameter(self):
+        return np.random.uniform(self.min_value, self.max_value)
```
chriselion: Is there any advantage to using numpy here instead of https://docs.python.org/3/library/random.html#random.uniform ?

ml-agents-envs/mlagents/envs/sampler_class.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class UniformSampler(Sampler):
+    def __init__(self, param_dict):
+        self.check_keys(param_dict)
+        self.min_value = param_dict["min"]
+        self.max_value = param_dict["max"]
+
+
+    def check_keys(self, param_dict):
```
chriselion: The other advantage of passing the min and max as params to init is that you don't need this check.

ml-agents-envs/mlagents/envs/sampler_class.py
```diff
@@ -0,0 +1,112 @@
...
+                raise SamplerException(
+                    key + " is required to use MultiRange Uniform sampler but" 
+                    " this value isn't provided. Please recheck"
+                    " the documentation to ensure that the sampler file"
+                    " matches the required format."
+                )
+
+    
+    def sample_parameter(self):
+        cur_min, cur_max = self.intervals[np.random.randint(len(self.intervals))]
```
chriselion: It's an OK design choice, but I think when you're writing the documentation for this, you should be explicit that each interval is equally likely to be selected (regardless of size). As a user, if I pass [[1,2], [3, 5]], I might expect the latter to be selected twice as often.

ml-agents-envs/mlagents/envs/sampler_class.py
```diff
@@ -0,0 +1,112 @@
...
+    "gaussian": GaussianSampler,
+    "multirange_uniform": MultiRangeUniformSampler,
+    }
+
+    @staticmethod
+    def register_sampler(name, sampler_cls):
+        SamplerFactory.NAME_TO_CLASS[name] = sampler_cls
+    
+    def init_sampler_class(name, param_dict):
+        sampler_cls = SamplerFactory.NAME_TO_CLASS[name]
```
chriselion: Check `if name in NAME_TO_CLASS` and raise a friendlier exception if not.

ml-agents-envs/mlagents/envs/sampler_class.py
```diff
@@ -0,0 +1,112 @@
...
+            sampler_name = cur_param_dict["sampler-type"]
+            param_sampler = SamplerFactory.init_sampler_class(sampler_name, cur_param_dict)
+
+            self.samplers[param_name] = param_sampler
+
+    def sample_all(self, train_mode = True):
+        res = {}
+        for param_name, param_sampler in self.samplers.items():
+            res[param_name] = param_sampler.sample_parameter()
+        return res
```
chriselion: nit: newline at EOF. if you're using pycharm, you can use this to make it automatic 
image

ml-agents-envs/mlagents/envs/sampler_class.py
```diff
@@ -0,0 +1,112 @@
...
+        self.sampler_factory = SamplerFactory()
+        self.samplers = {}
+        for param_name, cur_param_dict in self.reset_param_dict.items():
+            # cur_param_dict = reset_param_dict[param_name]
+            sampler_name = cur_param_dict["sampler-type"]
+            param_sampler = SamplerFactory.init_sampler_class(sampler_name, cur_param_dict)
+
+            self.samplers[param_name] = param_sampler
+
+    def sample_all(self, train_mode = True):
```
chriselion: `train_mode` unused - use it or remove it?

ml-agents-envs/mlagents/envs/sampler_class.py
```diff
@@ -0,0 +1,112 @@
...
+    "uniform": UniformSampler,
+    "gaussian": GaussianSampler,
+    "multirange_uniform": MultiRangeUniformSampler,
+    }
+
+    @staticmethod
+    def register_sampler(name, sampler_cls):
+        SamplerFactory.NAME_TO_CLASS[name] = sampler_cls
+    
+    def init_sampler_class(name, param_dict):
```
chriselion: This can be static. In fact, it either needs to be `@staticmethod` or you need to add a `self` argument (otherwise it's using `name`)

ml-agents-envs/mlagents/envs/sampler_class.py
```diff
@@ -0,0 +1,112 @@
...
+        sampler_cls = SamplerFactory.NAME_TO_CLASS[name]
+        return sampler_cls(param_dict)
+
+
+class SamplerManager:
+    def __init__(self, reset_param_dict):
+        self.reset_param_dict = reset_param_dict
+        self.sampler_factory = SamplerFactory()
```
chriselion: Don't need to instantiate this, just use static methods

ml-agents-envs/mlagents/envs/sampler_class.py
```diff
@@ -0,0 +1,112 @@
...
+        self.samplers = {}
+        for param_name, cur_param_dict in self.reset_param_dict.items():
+            # cur_param_dict = reset_param_dict[param_name]
+            sampler_name = cur_param_dict["sampler-type"]
```
chriselion: nit: `sampler_type` or just `type`?

If you're going to pass as kwargs to `init_sampler_class`, you may want to do this as `cur_param_dict.pop("sampler_type")` instead.

ml-agents-envs/mlagents/envs/sampler_class.py
```diff
@@ -0,0 +1,112 @@
...
+        SamplerFactory.NAME_TO_CLASS[name] = sampler_cls
+    
+    def init_sampler_class(name, param_dict):
+        sampler_cls = SamplerFactory.NAME_TO_CLASS[name]
+        return sampler_cls(param_dict)
+
+
+class SamplerManager:
+    def __init__(self, reset_param_dict):
+        self.reset_param_dict = reset_param_dict
```
chriselion: Don't need to keep the reference to this?

ml-agents/mlagents/trainers/learn.py
```diff
@@ -73,6 +78,12 @@ def run_training(sub_id: int, run_seed: int, run_options, process_queue):
             docker_target_name=docker_target_name
         )
 
+    sampler = None
```
chriselion: nit: rename to `sampler_config`. `sampler` implies it's a `Sampler` (which it's not)

chriselion: Can this and lesson_config default to empty dictionaries instead? That might simplify null handling later.

ml-agents/mlagents/trainers/learn.py
```diff
@@ -73,6 +78,12 @@ def run_training(sub_id: int, run_seed: int, run_options, process_queue):
...
+    lesson_config = None
+    if sampler_file_path is not None:
```
chriselion: Should raise an explicit error if `(sampler_file_path and not lesson_config_path)` (or may sampler_file_path xor lesson_config_path)

ml-agents/mlagents/trainers/lesson_controller.py
```diff
@@ -0,0 +1,122 @@
...
+import math
+
+from .exception import LessonControllerError
+
+import logging
+
+logger = logging.getLogger("mlagents.trainers")
+
+
+class LessonController(object):
```
chriselion: nit: in python3 you don't need to explicitly inherit from object. just `class LessonController:` is fine.

ml-agents/mlagents/trainers/lesson_controller.py
```diff
@@ -0,0 +1,122 @@
...
+            reward_buff_size (int): The size of the reward buffer of the trainer
+                that corresponds to the specified brain.
+
+        Returns:
+            Whether the curriculum of the specified brain should attempt to
+            increment its lesson.
+        """
+        return (reward_buff_size >= self.min_lesson_length)
+
+    def change_lesson(self, measure_val):
```
chriselion: nit: `should_change_lesson`?

chriselion: Hmm, maybe not. Ideally I'd like to separate the decision from the action, but that's harder with the smoothing. Need to think about it.

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -72,6 +83,8 @@ def __init__(
         self.fast_simulation = fast_simulation
         np.random.seed(self.seed)
         tf.set_random_seed(self.seed)
+        self.sampler_manager = SamplerManager(reset_param_dict) if reset_param_dict is not None else None
```
chriselion: This feels a little inconsistent - you create the SampleManager in here from the config, but the LessonController is created externally and passed. I think passing both as objects (instead) is slightly better.

ml-agents/mlagents/trainers/learn.py
chriselion: Sorry, didn't realize that `lesson_config` was actually a LessonController (which means it has a bad name!).

ml-agents/mlagents/trainers/lesson_controller.py
```diff
@@ -0,0 +1,122 @@
+import os
```
chriselion: Didn't realize this was (mostly) migrated code. Can curriculium.py be deleted?

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -173,7 +201,9 @@ def initialize_trainers(self, trainer_config: Dict[str, Dict[str, str]]):
                         brain_name
                     ].min_lesson_length
                     if self.meta_curriculum
-                    else 0,
+                    else self.lesson_controller.min_lesson_length
```
chriselion: The logic for this was debatably too complicated before, it definitely is now. Please extract it to a local variable or a helper method.

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -203,6 +233,18 @@ def _create_model_path(model_path):
                 "permissions are set correctly.".format(model_path)
             )
 
+    @staticmethod
+    def _check_reset_params(reset_params, new_config):
+        for k in new_config:
+            if (k in reset_params) and (isinstance(config[k], (int, float))):
```
chriselion: Do you mean `new_config` instead of `config` here?

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -90,6 +103,21 @@ def _get_measure_vals(self):
                     measure_val = np.mean(self.trainers[brain_name].reward_buffer)
                     brain_names_to_measure_vals[brain_name] = measure_val
             return brain_names_to_measure_vals
+
+        elif ((self.sampler_manager is not None) and (self.lesson_controller is not None)):
```
chriselion: Why do we need to check the sampler manager here? We don't use it.

Maybe it's just need, but I don't understand why the sampler needs to be this tied to to the lesson controller.

ml-agents/mlagents/trainers/trainer_controller.py
chriselion: It might make the code cleaner if you always create a SampleManager. A SamplerManager with an empty dictionary of reset parameters should be valid (it just won't return any reset parameters). Not sure how that interacts with the lesson_controller logic though.

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -215,6 +257,12 @@ def _reset_env(self, env: BaseUnityEnvironment):
                 train_mode=self.fast_simulation,
                 config=self.meta_curriculum.get_config(),
             )
+        elif self.sampler_manager is not None:
+            new_config = self.sampler_manager.sample_all()
```
chriselion: If you always create the SamplerManager, you need the `else` branch. Passing an empty dict for the config should be equivalent to not passing one.

I think you can simplify this whole function to something like
```python
new_config = self.sampler_manager.sample_all()
meta_curriculum_config = self.meta_curriculum.get_config() if self.meta_curriculum else {}
new_config.update(meta_curriculum_config)

return env.reset(train_mode=self.fast_simulation, config=new_config)
```

ml-agents/mlagents/trainers/lesson_controller.py
```diff
@@ -0,0 +1,122 @@
...
+        try:
+            with open(location) as data_file:
+                self.data = yaml.load(data_file)
+        except IOError:
+            raise LessonControllerError("The file {0} could not be found.".format(location))
+        except UnicodeDecodeError:
+            raise LessonControllerError("There was an error decoding {}".format(location))
+        self.smoothing_value = 0
+        self.check_keys(location)
+        self.measure = self.data["measure"]
```
chriselion: nit: I think some of this can be cleaned up - you probably don't need to keep `self.data` at all, just store these keys and data["signal_smoothing"]. That simplifies change_lesson() a bit.

ml-agents-envs/mlagents/envs/sampler_class.py
sankalp04: I decided to use np because I use np.random.choice which allows me to make weighted choices

ml-agents-envs/mlagents/envs/sampler_class.py
sankalp04: This wasn't the intended behavior; the new update of code will make the interval selection weighted based on their interval size.

ml-agents-envs/mlagents/envs/sampler_class.py
chriselion: https://docs.python.org/3/library/random.html#random.choices - you can pass the array of weights, or (even better) compute the cumulative weights in `__init__`

ml-agents/mlagents/trainers/lesson_controller.py
sankalp04: Even though the ideal scenario is to replace curriculum with some of this code, we don't want to do this rightaway. Having said that, lot of the code logic (and code) is portable here.

ml-agents-envs/mlagents/envs/sampler_class.py
chriselion: Hmm, I guess we set np.random's seed at the start. So you can ignore the comment about the seed if you want, and just stuck with numpy.

ml-agents/mlagents/trainers/trainer_controller.py
sankalp04: The idea here is that we only use the computation in this section if we are doing generalization training, which is indicated by those two objects.

ml-agents-envs/mlagents/envs/sampler_class.py
ervteng: Also: use Black to format your code, it will do this automatically as well: https://black.readthedocs.io/en/stable/

ml-agents-envs/mlagents/envs/sampler_class.py
sankalp04: So I tried setting np.random's seed, but there isn't a deterministic selection for the reset parameter sampling. I am guessing this is happening because in sample_all you can call the different sampling methods in a different order in each loop, which changes subsequent values. 

One alternative could be to maintain a seed in each of the sampler methods (i.e. - maintain the seed in each instance of UniformSampler or other sampling classes); what are your thoughts?

ml-agents-envs/mlagents/envs/sampler_class.py
sankalp04: So I tried setting np.random's seed, but there isn't a deterministic selection for the reset parameter sampling. I am guessing this is happening because in sample_all you can call the different sampling methods in a different order in each loop, which changes subsequent values. 

One alternative could be to maintain a seed in each of the sampler methods (i.e. - maintain the seed in each instance of UniformSampler or other sampling classes, though this would require a lot of calls to np.random.get_state to maintain the next seed).

Another idea is to deterministically loop over the self.samplers.items().

What are your thoughts?

PR #2144, opened 2019-06-17
Refactor reward signals into separate class

ervteng:
Refactors the reward signals (extrinsic and curiosity) into separate class that inherits from RewardSignal. The Trainer is now reward signal agnostic, and doesn't check the config whether or not each type exists - it just instantiates all of the classes declared. 

This is in preparation for additional reward signals (e.g. GAIL) as well as reuse across different trainers. 

Also, it is equivalent to the IRL PR but without the new features (GAIL and PreTraining). We're breaking up this PR into two to make it easier to review.

PR #2215, opened 2019-07-03
add some types to the reward signals

chriselion:
Add some type annotations for the changes that were introduced in https://github.com/Unity-Technologies/ml-agents/pull/2144


ml-agents/mlagents/trainers/components/reward_signals/curiosity/model.py
```diff
@@ -17,12 +18,13 @@ def __init__(
         """
         self.encoding_size = encoding_size
         self.policy_model = policy_model
+        self.next_visual_in: List[tf.Tensor] = []
```
chriselion: This is done in create_curiosity_encoders(); just trying to declare all the fields in init.

ervteng:
This looks good to me (other than the conflict, of course). I'll add this to the GAIL reward signal class as well.

PR #2236, opened 2019-07-08
Enable flake8

chriselion:
Flake8 is the standard python linter. It catches some things that black doesn't.

This enables it in pre-commit, and fixes up some of the errors that were flagged. Notably:
* black enforces a line limit of 88 by default, but this doesn't apply to docstrings. I set flake8's limit at 120 and fixed up the few violations
* Some unused variables in barracuda code (confirmed it was OK to remove)
* bare `except` statements - replaced these with `except Exception` which is just as broad.
* a few style things like `if x == False` -> `if not x`

I had to silence a lot of warnings coming from e.g. `from typing import *` - I'd like to remove all these eventually, but didn't want a huge PR for now.


xiaomaogy:
The PR looks good to me. The only thing is the "Some unused variables in barracuda code" you mentioned. Maybe it would be better to ask Barracuda team or maybe remove that change for the moment?

xiaomaogy:
@chriselion It seems that the test is broken.

chriselion:
Yes, it's broken because of this: https://github.com/Unity-Technologies/ml-agents/blob/c085438d36323da9239ceee1f1534528c051b3ab/ml-agents/mlagents/trainers/components/bc/model.py#L68-L84 Need to get @ervteng 's advice on what to do with it

PR #2239, opened 2019-07-10
Using-Docker.md miss a backslash in 3DBall command

AcelisWeaven:
Hi,

Just a quick edit because a backslash seems to be missing from the 3DBall command example.


unityjeffrey:
@AcelisWeaven - hi, can you change to develop from master?  We do not allow direct commits to master in this way.  thanks!

AcelisWeaven:
@unityjeffrey I changed the base branch, but I seems that two other commits have been included as well. Should I reopen my PR?

Also, I've noticed I needed to made some other changes for the command to run properly, namely:
- adding -it for the ability to kill the instance using Ctrl-C
- prefixing paths with the volume path (eg. /unity-volume/3DBall, /unity-volume/trainer_config.yaml)

Please let me know if I need to include thoses changes as well.

unityjeffrey:
CC: @xiaomaogy

xiaomaogy:
Hi @AcelisWeaven Thanks a lot for the PR. Your suggested changes regarding adding '-it' and '-p 6006:6006' make sense. Could you also add an example on how to use the tensorboard once you've made this change?

AcelisWeaven:
Hi @xiaomaogy 

Thanks for your reply. I've added the options and the Tensorboard example.

chriselion:
Looks good, thanks!

PR #2243, opened 2019-07-11
Add Unity command line arguments

BlueFisher:
Adding addition Unity command line arguments.
I had to build multiple executables for each scene again if I updated ml-agents. It was really tedious. So I would like to write a InitialScene script to load scenes depending on the command line arguments.
```c sharp
public class InitialScene : MonoBehaviour {
    void Start() {
        List<string> commandLineArgs = new List<string>(Environment.GetCommandLineArgs());
        int index = commandLineArgs.IndexOf("--scene");
        if(index == -1) {
            File.WriteAllText("error.log", $"No --scene in {Environment.CommandLine}");
            Application.Quit();
        }
        else {
            string sceneName = commandLineArgs[index + 1];
            SceneManager.LoadScene(sceneName);
        }
    }
}
```
Then I only have to build once.

### DIFF
--- a/ml-agents-envs/mlagents/envs/environment.py
+++ b/ml-agents-envs/mlagents/envs/environment.py
@@ -48,6 +48,7 @@ class UnityEnvironment(BaseUnityEnvironment):
         docker_training: bool = False,
         no_graphics: bool = False,
         timeout_wait: int = 30,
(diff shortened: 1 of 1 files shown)


unityjeffrey:
CC: @harperj

Issue #1798, opened 2019-03-07
Unable to shuffle if the fields are not of same length


michael20at:
Hm, tried ml-agents 0.7 with my new 2080ti (runs great on normal Tensorflow), 
updated Unity to the newest version 2018.3.7f1, training of 3D Balls example via cmd starts fine, but after a few episode (sometimes only a few, sometimes after 9) I get 

"Unable to shuffle if the fields are not of same length"!

What could be the problem?

michael20at:
Hm, it kinda works if I increase the buffer size by ten in the config file, but I still don't know why?

xiaomaogy:
This is kind of weird and unexpected, I don't see other people raising this issue, so I guess maybe delete and rerun and try to follow the basic guide step by step might help.

Ina299:
I also experienced this issue.
I use ml-agents 0.6 on 1080ti.
`       
        batch_size:     1536
        beta:   0.006
        buffer_size:    15360
        epsilon:        0.17
        gamma:  0.995
        hidden_units:   512
        lambd:  0.9
        learning_rate:  0.00012
        max_steps:      5.0e15
        normalize:      True
        num_epoch:      4
        num_layers:     4
        time_horizon:   768
        summary_freq:   2000
        use_recurrent:  False
        use_curiosity:  False
`
I saw this phenomenon when I took agent' s life time too long.
So I guess agent should be Done() before buffer_size.

ervteng:
Are you using a Numpy version greater than 1.14.1 by any chance? Newer versions of Numpy are known to have similar issues.

rainysolar:
I have the same issue.
I use a 2070 card which only support CUDA 10.
CUDA 10 needs TensorFlow >=1.13.0.
And Numpy 1.14.1 is not support TensorFlow>=1.13.0

yttiktak:
Same issue. Using Ubuntu Bionic, numpy v 1.16.2
Trying to track it down, seems a numpy array (advantages) changes shape from (N,) to (1,N) when the ppy trainer.py subtracts the mean from it, about line 325. 
 I hacked out a thing to bypass policy updates when that happens. Walker seems to be training now. My 5 yr old approves of the walker. It got up to a reward of 300 so far.

JadenTravnik:
[...]

PR #2251, opened 2019-07-12
Update buffer.py

EfveZombie:
A simple fix of "Unable to shuffle if the fields are not of same length".


unityjeffrey:
hi @EfveZombie - can you change target branch to Develop instead of master?  We do not allow PRs to Master this way.  Thanks!

ervteng:
Hey @EfveZombie, thanks for adding this! Does using `dtype=np.float32` also fix your issue? IMO that would be a bit cleaner than `float`

EfveZombie:
> Hey @EfveZombie, thanks for adding this! Does using `dtype=np.float32` also fix your issue? IMO that would be a bit cleaner than `float`

I think that would work too. 

In the old version, some part in buffer.py would return a ndarray with dtype=Object. The format was like `array([0.0, array([1.]), array([2.]), 3.0], dtype=object)`. So I tried to set the dtype manually and it work fine so far, but I still dont know what caused this problem.

chriselion:
Issue link for reference: https://github.com/Unity-Technologies/ml-agents/issues/1798

chriselion:
We think the root cause for this was fixed in https://github.com/Unity-Technologies/ml-agents/pull/2276. Going to close this, but if you have more problems, please reopen.

PR #2265, opened 2019-07-15
Make SubprocessEnvManager take asynchronous steps

harperj:
SubprocessEnvManager takes steps synchronously to reproduce old
behavior, meaning all parallel environments will need to wait for
the slowest environment to take a step.  If some steps take much
longer than others, this can lead to a substantial overall slowdown
in practice.  We've seen extreme cases where we see almost a 2x
speedup from using asynchronous stepping, with no downside for our
faster environments.

This PR changes the SubprocessEnvManager to use async stepping.
This means on the "step" call the environment manager will enqueue
step requests to workers, and then only wait until at least one
step has been completed before returning.

**Testing**

_Reacher_ environment on GCP with 8 parallel environments.  15000 steps.  16.5% improvement in overall time for Async EnvManager, comparable reward.

_Walker_ environment on GCP with 8 parallel environments.  30000 steps.  14.2% improvement in overall time for Async EnvManager, comparable reward.

### DIFF
--- a/ml-agents-envs/mlagents/envs/subprocess_env_manager.py
+++ b/ml-agents-envs/mlagents/envs/subprocess_env_manager.py
@@ -42,7 +42,7 @@ class UnityEnvWorker:
         try:
             response: EnvironmentResponse = self.conn.recv()
             return response
-        except (BrokenPipeError, EOFError) as e:
+        except (BrokenPipeError, EOFError):
             raise KeyboardInterrupt
 
     def close(self):
--- a/ml-agents-envs/mlagents/envs/subprocess_env_manager.py
+++ b/ml-agents-envs/mlagents/envs/subprocess_env_manager.py
@@ -69,6 +69,8 @@ def worker(
             cmd: EnvironmentCommand = parent_conn.recv()
             if cmd.name == "step":
                 all_action_info = cmd.payload
+                # When an environment is "global_done" it means automatic agent reset won't occur, so we need
+                # to perform an academy reset.
                 if env.global_done:
                     all_brain_info = env.reset()
                 else:
@@ -141,10 +143,13 @@ class SubprocessEnvManager(EnvManager):
                 env_worker.waiting = True
 
     def step(self) -> List[StepInfo]:
+        # Queue steps for any workers which aren't in the "waiting" state.
         self._queue_steps()
 
         worker_steps: List[EnvironmentResponse] = []
         step_workers: Set[int] = set()
+        # Poll the step queue for completed steps from environment workers until we retrieve
+        # 1 or more, which we will then return as StepInfos
         while len(worker_steps) < 1:
             try:
                 while True:
@@ -162,8 +167,10 @@ class SubprocessEnvManager(EnvManager):
     def reset(
         self, config=None, train_mode=True, custom_reset_parameters=None
     ) -> List[StepInfo]:
+        # First enqueue reset commands for all workers so that they reset in parallel
         for ew in self.env_workers:
             ew.send("reset", (config, train_mode, custom_reset_parameters))
+        # Next (synchronously) collect the reset observations from each worker in sequence
         for ew in self.env_workers:
             ew.previous_step = StepInfo(None, ew.recv().payload, None)
         return list(map(lambda ew: ew.previous_step, self.env_workers))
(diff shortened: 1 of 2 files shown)


ml-agents-envs/mlagents/envs/subprocess_env_manager.py
```diff
@@ -104,59 +109,63 @@ def __init__(
...
-            env_worker.previous_step = step_info
-            steps.append(step_info)
-        return steps
+            if not env_worker.waiting:
+                env_action_info = self._take_step(env_worker.previous_step)
+                env_worker.previous_all_action_info = env_action_info
+                env_worker.send("step", env_action_info)
+                env_worker.waiting = True
+
+    def step(self) -> List[StepInfo]:
```
chriselion: Is a list the best way to return this? Previously the steps were 1:1 with the workers (I think), but now a worker might contribute 0 or more than 1 steps. I would think Dict[str, List[StepInfo]] where the dict key is the worker ID.

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
```diff
@@ -104,59 +109,63 @@ def __init__(
...
+        self._queue_steps()
+
+        worker_steps: List[EnvironmentResponse] = []
+        step_workers: Set[int] = set()
+        while len(worker_steps) < 1:
```
chriselion: Could you add a few comments on what's going on here, e.g. "Wait until at least one environment is finished stepping". I think it's pretty clear from the PR description but not necessarily from the code.

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
harperj: @chriselion Could you elaborate on why you think it would be valuable to return a dictionary rather than a list?  I don't see any particular value in splitting the steps out by worker at this time since we just iterate over them elsewhere in the codebase.

It would mean changing the interface as well, which means this change wouldn't be only in the implementation of an EnvManager but bleed into TrainerController.

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
harperj: Yeah, I think that is a good idea 👍

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
chriselion: OK, maybe it's not necessary.

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
```diff
@@ -104,59 +109,63 @@ def __init__(
...
+                    step = self.step_queue.get_nowait()
+                    self.env_workers[step.worker_id].waiting = False
+                    if step.worker_id not in step_workers:
+                        worker_steps.append(step)
+                        step_workers.add(step.worker_id)
+            except EmptyQueue:
+                pass
+
+        step_infos = self._postprocess_steps(worker_steps)
+        return step_infos
```
ervteng: Does this change the meaning of `step` from the `trainer_controller` side? I suppose on average it should be `num_envs` of steps per step call.

harperj: No, as mentioned in person that change was made in #2209

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
```diff
@@ -2,12 +2,13 @@
 import cloudpickle
 
 from mlagents.envs import UnityEnvironment
-from multiprocessing import Process, Pipe
+from multiprocessing import Process, Pipe, Queue
 from multiprocessing.connection import Connection
+from queue import Empty as EmptyQueue
```
chriselion: nit: maybe call this `EmptyQueueException`? bad naming in the stdlib :/

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
```diff
@@ -104,59 +111,68 @@ def __init__(
...
     ) -> List[StepInfo]:
-        self._broadcast_message("reset", (config, train_mode, custom_reset_parameters))
-        reset_results = [
-            self.env_workers[i].recv().payload for i in range(len(self.env_workers))
-        ]
-        for i in range(len(reset_results)):
-            env_worker = self.env_workers[i]
-            env_worker.previous_step = StepInfo(None, reset_results[i], None)
+        # First enqueue reset commands for all workers so that they reset in parallel
+        for ew in self.env_workers:
```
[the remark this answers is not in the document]
harperj: 👍

PR #2266, opened 2019-07-15
Very simple environment for testing

chriselion:
Still needs a bit of cleanup, but the goal here is to create a simple "integration" test so we can run the full training loop.

This gets the code coverage in unit tests from 66% to 72%; mostly of the increase is from `trainers/ppo/policy.py` and `trainers/ppo/trainer.py`

Had to clean up a few interface inconsistencies (mypy should be finding these but it's not checking across files very well right now), and changed TrainerController to return some "final" stats so that we can test them.


ml-agents/mlagents/trainers/tests/test_environments/test_simple.py
```diff
@@ -0,0 +1,256 @@
...
+
+class EnvContext:
+    def __init__(self, env: BaseUnityEnvironment):
+        self.env = env
+        self.previous_step: StepInfo = StepInfo(None, {}, None)
+        self.previous_all_action_info: Dict[str, ActionInfo] = {}
+
+
+# Copied from SubprocessEnvManager and removed the subprocess part
+class LocalEnvManager(EnvManager):
```
chriselion: @harperj here's the un-subprocess'ed version of SubprocessEnvManager. I can simplify it a bit further to only ever handle a single environment.

chriselion:
@ervteng Follow up from Friday about creating a simple "game" to test training without mocking anything. Feedback welcome on how to simplify/improve this.

I think from here, we can add other tests to hit the other components like visual observations, memory, and curiosity. It should hopefully be easy enough to extend to GAIL too.

ml-agents/mlagents/trainers/tests/test_environments/test_simple.py
chriselion: If you're generally happy with the PR, I'll either move this to ml-agents-envs, or a different file in this test directory (depending on whether we think it's generally useful).

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -238,24 +239,40 @@ def _not_done_training(self) -> bool:
...
                     global_step,
                     delta_train_start,
                     lesson_num=self.meta_curriculum.brains_to_curriculums[
                         brain_name
                     ].lesson_num,
                 )
             else:
-                trainer.write_summary(global_step, delta_train_start)
+                brain_stats = trainer.write_summary(global_step, delta_train_start)
+            stats_out[brain_name] = brain_stats
```
ervteng: There's currently 2 (now 3) ways we get the reward out of the trainer - `_get_measure_vals()` for meta curriculum, the `trainer.write_summary`, and however `trainer_metrics` does it. I wonder if we could make `_get_measure_vals()` more generic and use that to get the brain_stats instead of relying on write_summary (or the other way around)?

ervteng:
> @ervteng Follow up from Friday about creating a simple "game" to test training without mocking anything. Feedback welcome on how to simplify/improve this.
> 
> I think from here, we can add other tests to hit the other components like visual observations, memory, and curiosity. It should hopefully be easy enough to extend to GAIL too.

Yeah, this should be able to be extended to GAIL/Curiosity fairly easily. We can just record a .demo file, or perhaps more easily generate a bunch of steps from an "optimal" policy (just stepping in the right direction) and creating a buffer from those. Then setting the GAILRewardSignal's internal buffer to this new buffer, and running training.

ml-agents/mlagents/trainers/tests/test_environments/test_simple.py
```diff
@@ -0,0 +1,256 @@
...
+        text_action: Dict[str, Any] = None,
+        value: Dict[str, Any] = None,
+    ) -> AllBrainInfo:
+        if vector_action:
+            delta = vector_action[BRAIN_NAME][0][0]
+            self.position += clamp(delta, -STEP_SIZE, STEP_SIZE)
+            self.position = clamp(self.position, -1, 1)
+            self.step_count += 1
+        done = self.position >= 1.0 or self.position <= -1.0
+        reward = self.position if done else 0.0
```
ervteng: We should probably add a small time penalty (~-0.001) as well so the agent is incentivized to got to the right position quickly, and change the win threshold accordingly

chriselion: Done.

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -159,35 +162,35 @@ def initialize_trainers(
...
                         brain_name
                     ].min_lesson_length
                     if self.meta_curriculum
-                    else 0,
-                    trainer_parameters_dict[brain_name],
-                    self.train_model,
-                    self.load_model,
-                    self.seed,
-                    self.run_id,
+                    else 1,
```
chriselion: This is the only real change.

ml-agents/mlagents/trainers/trainer_controller.py
chriselion: OK, undid these changes and updated `_get_measure_vals()` to get the average of the rewards buffers, and call that on the test. I'm still not sure this is a great solution, since that buffer will only be length 1 for non-curriculum learning. What do you think?

ml-agents-envs/mlagents/envs/local_env_manager.py
```diff
@@ -0,0 +1,84 @@
+from typing import Any, Dict, List
+
+from mlagents.envs.base_unity_environment import BaseUnityEnvironment
+from mlagents.envs.env_manager import EnvManager, StepInfo
+from mlagents.envs.timers import timed
+from mlagents.envs import ActionInfo, BrainParameters
+
+
+class LocalEnvManager(EnvManager):
```
chriselion: @harperj I simplified this to a single environment and moved it to `ml-agents-envs`

harperj: 👍

harperj: I think this isn't an ideal name since "local" makes me think of local machine vs. remote machine.  Maybe `SingleEnvManager`?

ml-agents-envs/mlagents/envs/local_env_manager.py
```diff
@@ -0,0 +1,84 @@
...
+    This is generally only useful for testing; see SubprocessEnvManager for a production-quality implementation.
+    """
+
+    def __init__(self, env: BaseUnityEnvironment):
+        super().__init__()
+        self.env = env
+        self.previous_step: StepInfo = StepInfo(None, {}, None)
+        self.previous_all_action_info: Dict[str, ActionInfo] = {}
+
+    def get_last_steps(self):
```
harperj: This function should be unnecessary against the latest `develop`

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -254,8 +257,12 @@ def write_to_tensorboard(self, global_step: int) -> None:
                 trainer.write_summary(global_step, delta_train_start)
 
     def start_learning(
-        self, env_manager: SubprocessEnvManager, trainer_config: Dict[str, Any]
+        self, env_manager: EnvManager, trainer_config: Dict[str, Any]
     ) -> None:
+        """
+        Returns stats for each trainer.
```
harperj: This isn't really a descriptive docstring for this function -- it really returns `None` but has a side effect of outputting trainer stats.  Probably should either change this or remove / push to a different PR.

ml-agents-envs/mlagents/envs/local_env_manager.py
chriselion: Removed.

ml-agents-envs/mlagents/envs/local_env_manager.py
chriselion: Will do.

ml-agents/mlagents/trainers/trainer_controller.py
chriselion: Actually I forgot to remove this - a previous attempt returned stats before Ervin pointed me to `_get_measure_vals`

chriselion:
@harperj Feedback addressed.

PR #2268, opened 2019-07-16
Get timers from subprocess

chriselion:
Extends the previous timer code to return the timer from the subprocess workers, and merge them into the timer tree.

Deciding to represent this is the hardest part. Instead of trying to track real/user time, I went with a simpler approach of flagging the nodes that are run in parallel; this is set when merging the trees.


ml-agents-envs/mlagents/envs/subprocess_env_manager.py
```diff
@@ -90,6 +95,11 @@ def _send_response(cmd_name, payload):
                 _send_response("reset", all_brain_info)
             elif cmd.name == "global_done":
                 _send_response("global_done", env.global_done)
+            elif cmd.name == "timers":
```
chriselion: @harperj Let's talk about this - I'm not sure if it's better to add a separate command to request the timers, or return them in `step` response. The latter is probably easier with async stepping, but the downside is that it might make it difficult to time `_send_response()`.

chriselion: (not going to fix the breaking unit test until we decide what to do)

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
```diff
@@ -78,7 +89,16 @@ def _send_response(cmd_name, payload):
                         texts[brain_name] = action_info.text
                         values[brain_name] = action_info.value
                     all_brain_info = env.step(actions, memories, texts, values)
-                _send_response("step", all_brain_info)
+
+                # The timers in this process are independent from all the processes and the "main" process
+                # So after we send back the root timer, we can safely clear them.
+                # Note that we could randomly return timers a fraction of the time if we wanted to reduce
+                # the data transferred.
+                step_response = StepResponse(
```
chriselion: @harperj  - changed the step to return a NamedTuple with (optional) timing information. This should make it pretty easy to integrate with the async changes (I hope)

ml-agents-envs/mlagents/envs/subprocess_env_manager.py
```diff
@@ -7,8 +7,14 @@
...
 from mlagents.envs.base_unity_environment import BaseUnityEnvironment
 from mlagents.envs.env_manager import EnvManager, StepInfo
-from mlagents.envs.timers import timed
-from mlagents.envs import BrainParameters, ActionInfo
+from mlagents.envs.timers import (
+    TimerNode,
+    timed,
+    hierarchical_timer,
+    reset_timers,
+    _global_timer_stack,
```
harperj: Should we change the name of this since we're not using it as private?

chriselion: I moved access to a function instead.

ml-agents-envs/mlagents/envs/environment.py
```diff
@@ -625,6 +627,7 @@ def _get_state(self, output: UnityRLOutput) -> Tuple[AllBrainInfo, bool]:
             )
         return _data, global_done
 
+    @timed
     def _generate_step_input(
```
xiaomaogy: What is the purpose of timing this in addition to timing the env.step time here?

xiaomaogy:
All seems reasonable to me.

ml-agents-envs/mlagents/envs/environment.py
chriselion: I put it there because that's where the protobuf serialization occurs. I'll put one on RpcCommunicator.communicate too.

PR #2272, opened 2019-07-16
Updated links to point to KR blog site

unityjeffrey:


unityjeffrey:
yep, it just points them to all localized kr blog site.  the local team may end up translating later.

PR #2274, opened 2019-07-16
Fix for vis obs memory leak in docker

awjuliani:
Addresses an issue when on linux and using cpu rendering where Texture2D objects are not properly cleaned up during garbage collection.

### DIFF
--- a/UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
+++ b/UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
@@ -1,4 +1,4 @@
-﻿using System.Collections.Generic;
+using System.Collections.Generic;
 using System.Linq;
 using Google.Protobuf;
 using MLAgents.CommunicatorObjects;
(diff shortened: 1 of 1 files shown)


UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -531,21 +521,15 @@ void ResetData()
 
             BrainParameters param = brain.brainParameters;
             actionMasker = new ActionMasker(param);
-            // If we haven't initialized vectorActions, initialize to 0. This should only
-            // happen during the creation of the Agent. In subsequent episodes, vectorAction
-            // should stay the previous action before the Done(), so that it is properly recorded. 
-            if (action.vectorActions == null)
```
harperj: Looks like this removes a check -- is that intentional?

awjuliani: It is not. Fixing now.

PR #2276, opened 2019-07-17
Fix get_value_estimate and buffer append

ervteng:
In newer versions of numpy (>1.14), the fact that we convert a np array into a list, append an np array, and convert it back to a list in `get_gae` causes problems. 

This commit changes `get_value_estimates` to output a `dict` of floats rather than np arrays, and uses np.append rather than converting to a list and back to resolve this issue.

### DIFF
--- a/ml-agents/mlagents/trainers/ppo/policy.py
+++ b/ml-agents/mlagents/trainers/ppo/policy.py
@@ -190,7 +190,7 @@ class PPOPolicy(TFPolicy):
         run_out = self._execute_model(feed_dict, self.update_dict)
         return run_out
 
-    def get_value_estimates(self, brain_info, idx):
+    def get_value_estimates(self, brain_info, idx, done):
         """
         Generates value estimates for bootstrapping.
         :param brain_info: BrainInfo to be used for bootstrapping.
@@ -214,6 +214,12 @@ class PPOPolicy(TFPolicy):
                 idx
             ].reshape([-1, len(self.model.act_size)])
         value_estimates = self.sess.run(self.model.value_heads, feed_dict)
+
+        for reward_signal in value_estimates:
+            value_estimates[reward_signal] = (
+                0.0 if done else value_estimates[reward_signal][0]
+            )
+
         return value_estimates
 
     def get_action(self, brain_info: BrainInfo) -> ActionInfo:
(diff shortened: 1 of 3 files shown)


ml-agents/mlagents/trainers/ppo/trainer.py
```diff
@@ -346,9 +346,12 @@ def process_experiences(
                 else:
                     bootstrapping_info = info
                     idx = l
-                value_next = self.policy.get_value_estimates(bootstrapping_info, idx)
-                if info.local_done[l] and not info.max_reached[l]:
-                    value_next["extrinsic"] = 0.0
+                value_next = self.policy.get_value_estimates(
+                    bootstrapping_info,
+                    idx,
+                    info.local_done[l] and not info.max_reached[l],
```
chriselion: Is it possible to determine this just from `bootstrapping_info` and `idx` within `get_value_estimates()`? I couldn't quite convince myself when I was looking at it.

chriselion:
Note that this should fix https://github.com/Unity-Technologies/ml-agents/issues/1798

ml-agents/mlagents/trainers/ppo/policy.py
```diff
@@ -190,14 +190,18 @@ def update(self, mini_batch, num_sequences):
         run_out = self._execute_model(feed_dict, self.update_dict)
         return run_out
 
-    def get_value_estimates(self, brain_info, idx):
+    def get_value_estimates(self, brain_info, idx, done):
```
chriselion: type annotations if it's not too hard to add?

ml-agents/mlagents/trainers/ppo/trainer.py
ervteng: Hmm, I'm not sure - it seems like `bootstrapping_info` becomes the previous `info` if the `not` condition is met. Seems like we'll run into an issue if both of those conditions are met, since `bootstrapping_info` will be sth different and we can't figure out if `info.local_done` is True.

chriselion: OK, that's more or less what I thought. Sounds good as it is!

ml-agents/mlagents/trainers/ppo/policy.py
ervteng: Added type annotation and test for this function

PR #2277, opened 2019-07-17
Fix respawn part of BananaLogic

LuoYouRen:
Fix the bug of "respawn" part that cause all the banana respawn in the first Area.


ervteng:
I can confirm that this does fix VisualBanana. However, it breaks the regular Banana scene - all of the bananas will fall through the ground. 

I think if you remove the `transform.position.y` from that sum it will work for both scenes.

LuoYouRen:
> I can confirm that this does fix VisualBanana. However, it breaks the regular Banana scene - all of the bananas will fall through the ground.
> 
> I think if you remove the `transform.position.y` from that sum it will work for both scenes.

That's right. I just fixed it and committed. Is that OK?

PR #2289, opened 2019-07-18
Add different types of visual encoder (nature cnn/resnet)

dongruoping:
Add different types of visual encoder (nature cnn/resnet)


ml-agents/mlagents/trainers/models.py
```diff
@@ -229,28 +230,101 @@ def create_visual_observation_encoder(
...
-                conv1,
-                32,
-                kernel_size=[4, 4],
-                strides=[2, 2],
-                activation=tf.nn.elu,
-                reuse=reuse,
-                name="conv_2",
-            )
-            hidden = c_layers.flatten(conv2)
+            if encoder_type == "resnet":
```
chriselion: Can we add an enum for these instead? https://docs.python.org/3/library/enum.html

ml-agents/mlagents/trainers/models.py
```diff
@@ -229,28 +230,101 @@ def create_visual_observation_encoder(
...
+                            kernel_size=[3, 3],
+                            strides=[1, 1],
+                            padding="same",
+                            reuse=reuse,
+                            name="layer%d_%d_conv2" % (i, j),
+                        )
+                        hidden = tf.add(block_input, hidden)
+                hidden = tf.nn.relu(hidden)
+                hidden = c_layers.flatten(hidden)
+            elif encoder_type == "nature_cnn":
```
harperj: I think in this case it would be better to put each of the different encoder types in its own `build_encoder` function -- so `build_nature_encoder` and `build_resnet_encoder` for example.

harperj:
Noticed this doesn't actually hook the different encoder types up to any config.  Is that intended to be a separate PR / planned at all?

BTW, also please make sure to make a descriptive commit message and PR description (they can be the same) for this.

ml-agents/mlagents/trainers/models.py
dongruoping: added

ml-agents/mlagents/trainers/models.py
dongruoping: yes sure. done.

dongruoping:
> Noticed this doesn't actually hook the different encoder types up to any config. Is that intended to be a separate PR / planned at all?
> 
> BTW, also please make sure to make a descriptive commit message and PR description (they can be the same) for this.

Added config option for encoder type in trainer config

config/trainer_config.yaml
```diff
@@ -16,6 +16,7 @@ default:
     sequence_length: 64
     summary_freq: 1000
     use_recurrent: false
+    vis_encode_type: "default"
```
harperj: I don't think quotes are needed in yaml.

config/trainer_config.yaml
```diff
@@ -16,6 +16,7 @@ default:
     sequence_length: 64
     summary_freq: 1000
     use_recurrent: false
+    vis_encode_type: default
```
chriselion: If "default" is really the default, you shouldn't need to put it in the configs.

ml-agents/mlagents/trainers/models.py
```diff
@@ -302,7 +434,9 @@ def create_discrete_action_masking_layer(all_logits, action_masks, action_size):
             ),
         )
 
-    def create_observation_streams(self, num_streams, h_size, num_layers):
+    def create_observation_streams(
```
chriselion: Please add type annotations here. That would have told you that you've got a mismatch between the default value of vis_encode_type (a string) and the comparisons (using the enum).

ml-agents/mlagents/trainers/models.py
```diff
@@ -401,13 +559,15 @@ def create_value_heads(self, stream_names, hidden_input):
             self.value_heads[name] = value
         self.value = tf.reduce_mean(list(self.value_heads.values()), 0)
 
-    def create_cc_actor_critic(self, h_size, num_layers):
+    def create_cc_actor_critic(self, h_size, num_layers, vis_encode_type):
```
chriselion: type annotations

ml-agents/mlagents/trainers/ppo/policy.py
```diff
@@ -44,6 +44,7 @@ def __init__(self, seed, brain, trainer_params, is_training, load):
                 m_size=self.m_size,
                 seed=seed,
                 stream_names=list(reward_signal_configs.keys()),
+                vis_encode_type=trainer_params["vis_encode_type"],
```
chriselion: Convert to EncoderType enum here and pass that through instead of the string type. Handle a missing value with .get(). For example

``` python
vis_encode_type = EncoderType(trainer_params.get("vis_encode_type", "default"))
```

harperj:
@dongruoping can you make sure to address the comments here from @chriselion in a following PR?

PR #2294, opened 2019-07-19
fix bug in RandomNormal

chriselion:
A typo in `RandomNormal.NextDouble()` was causing it to generate values with the right mean but incorrect standard deviation. This fixes the bug and also adds a test that the sample mean and stddev are close to what's expected.

Note on whitespace: I used the standard (I think) `.editorconfig` and applied "Format Document" in Visual Studio. This changed a lot of tabs to spaces.

### DIFF
--- a/UnitySDK/Assets/ML-Agents/Editor/Tests/RandomNormalTest.cs
+++ b/UnitySDK/Assets/ML-Agents/Editor/Tests/RandomNormalTest.cs
@@ -1,122 +1,163 @@
-﻿using System;
+using System;
 using NUnit.Framework;
 using MLAgents.InferenceBrain;
 using MLAgents.InferenceBrain.Utils;
 
+using UnityEngine;
+using System.Collections;
+
+
+
 namespace MLAgents.Tests
 {
 
-	public class RandomNormalTest
-	{
+    public class RandomNormalTest
+    {
+        private const float first = -1.19580f;
+        private const float second = -0.97345f;
 
-		[Test]
-		public void RandomNormalTestTwoDouble()
-		{
-			RandomNormal rn = new RandomNormal(2018);
+        [Test]
+        public void RandomNormalTestTwoDouble ()
+        {
+            RandomNormal rn = new RandomNormal (2018);
 
-			Assert.AreEqual(-0.46666, rn.NextDouble(), 0.0001);
-			Assert.AreEqual(-0.37989, rn.NextDouble(), 0.0001);
-		}
+            Assert.AreEqual (first, rn.NextDouble (), 0.0001);
+            Assert.AreEqual (second, rn.NextDouble (), 0.0001);
+        }
 
-		[Test]
-		public void RandomNormalTestWithMean()
-		{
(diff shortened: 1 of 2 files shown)


UnitySDK/Assets/ML-Agents/Editor/Tests/RandomNormalTest.cs
```diff
@@ -1,122 +1,163 @@
...
 
-	public class RandomNormalTest
-	{
-
-		[Test]
-		public void RandomNormalTestTwoDouble()
-		{
-			RandomNormal rn = new RandomNormal(2018);
-
-			Assert.AreEqual(-0.46666, rn.NextDouble(), 0.0001);
```
chriselion: These tests were all hardcoded to the values; I redid them to check that the mean and stddev were accounted for correctly, based on the first two values that the generator would return.

UnitySDK/Assets/ML-Agents/Editor/Tests/RandomNormalTest.cs
```diff
@@ -1,122 +1,163 @@
...
+            RandomNormal rn = new RandomNormal (1982);
+            Tensor t = new Tensor {
+                ValueType = Tensor.TensorType.FloatingPoint
+            };
+
+            Assert.Throws<ArgumentNullException> (() => rn.FillTensor (t));
+        }
+
+        [Test]
+        public void RandomNormalTestDistribution ()
```
chriselion: This test fails before the fix; the mean comes out right but the stddev is way off.

UnitySDK/Assets/ML-Agents/Scripts/InferenceBrain/Utils/RandomNormal.cs
```diff
@@ -44,7 +44,7 @@ public double NextDouble()
                 s = u * u + v * v;
             } while (s >= 1.0 || s == 0.0);
 
-            s = Math.Sqrt(-2.0 * Math.Log(s) / 2);
+            s = Math.Sqrt(-2.0 * Math.Log(s) / s);
```
chriselion: Here's the bug, as per https://en.wikipedia.org/wiki/Marsaglia_polar_method

UnitySDK/Assets/ML-Agents/Editor/Tests/RandomNormalTest.cs
```diff
@@ -1,122 +1,163 @@
...
+    {
+        private const float first = -1.19580f;
+        private const float second = -0.97345f;
+
+        [Test]
+        public void RandomNormalTestTwoDouble ()
+        {
+            RandomNormal rn = new RandomNormal (2018);
+
+            Assert.AreEqual (first, rn.NextDouble (), 0.0001);
```
surfnerd: should the epsilon be extracted as a variable as well?

chriselion: Yeah, could do. It's a different (looser) epsilon for the distribution test, but I could tighten it with more samples.

PR #2296, opened 2019-07-19
Improvements for GAIL

ervteng:
- Don't 0 value bootstrap for GAIL and Curiosity
- Add gradient penalties to GAN to help with stability 
- Add `gail_config.yaml` with GAIL examples
- Cleaned up `trainer_config.yaml` and unnecessary gammas
- Documentation updates
- Code cleanup


ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
```diff
@@ -261,5 +292,8 @@ def create_loss(self, learning_rate: float) -> None:
             )
         else:
             self.loss = self.discriminator_loss
+
+        self.loss = self.loss + self.gradient_penalty * self.compute_gradient_penalty()
```
awjuliani: I think we need better var names here. From the current names I would expect `self.compute_gradient_penalty()` to return the gradient penalty, but it returns the magnitude of the gradient.

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
```diff
@@ -228,7 +231,35 @@ def create_network(self) -> None:
...
+            interp.append(alpha * _expert_in + (1 - alpha) * _policy_in)
+
+        grad_estimate, _, grad_input = self.create_encoder(
+            interp[0], interp[1], interp[2], reuse=True
+        )
+
+        grad = tf.gradients(grad_estimate, [grad_input])[0]
+
+        # Norm, like log, can return NaN. Use our own safe_norm
+        safe_norm = tf.sqrt(tf.reduce_sum(grad ** 2, axis=-1) + EPSILON)
```
chriselion: How does norm result in NaN? I could see that happening if there was overflow, but in that case adding an epsilon isn't going to help.

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
```diff
@@ -228,7 +231,35 @@ def create_network(self) -> None:
         self.discriminator_score = tf.reshape(
             self.policy_estimate, [-1], name="GAIL_reward"
         )
-        self.intrinsic_reward = -tf.log(1.0 - self.discriminator_score + 1e-7)
+        self.intrinsic_reward = -tf.log(1.0 - self.discriminator_score + EPSILON)
+
+    def compute_gradient_penalty(self) -> tf.Tensor:
```
awjuliani: What is the performance improvement from this? Faster/more stable convergence?

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
ervteng: Not the norm, it's the gradient of the norm. At 0 the gradient of sqrt() is a horizontal line. I'll update the comment to reflect this.

ml-agents/mlagents/trainers/components/reward_signals/reward_signal.py
```diff
@@ -32,6 +32,9 @@ def __init__(self, policy: TFPolicy, strength: float, gamma: float):
         short_name = class_name.replace("RewardSignal", "")
         self.stat_name = f"Policy/{short_name} Reward"
         self.value_name = f"Policy/{short_name} Value Estimate"
+        # Don't terminate discounted reward computation at Done. Useful for eliminating positive bias in rewards with
+        # no natural end, e.g. GAIL or Curiosity
+        self.ignore_terminal_states = False
```
chriselion: nit (feel free to ignore): could you name this in terms of what it does instead of what it doesn't do? I find double-negatives like this a little harder to understand when trying to reason about the code.

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
ervteng: Faster convergence, esp. for Crawler. I'm seeing about 25% less steps required with PPO. TBH a large motivation for this is for SAC, where without GP the discriminator will overfit very quickly - GAIL + SAC doesn't work at all without it.

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
chriselion: Maybe don't use `self.compute_gradient_penalty()` if `self.gradient_penalty == 0.0` (to allow turning it off if you really don't want it for some reason)?

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
ervteng: Plots for Crawler + PPO

awjuliani: The squiggly line does not lie.

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
ervteng: Changed this to `create_gradient_magnitude` and the weight to be `gradient_penalty_weight`

ml-agents/mlagents/trainers/components/reward_signals/reward_signal.py
ervteng: It was confusing for me too while writing it - I reversed it

ml-agents/mlagents/trainers/components/reward_signals/gail/model.py
ervteng: Made it only happen if the penalty is greater than 0

ml-agents/mlagents/trainers/components/reward_signals/reward_signal.py
chriselion: Cool, not sure if it's on any official naming scheme guides, but I generally try avoid using variables with negative names for this reason.

PR #2311, opened 2019-07-22
proof of concept - attrs TrainerParameters

chriselion:
Any thoughts on converting to something like this? I think it would let us simplify the yaml config files (remove anything that's set to the default) and clean up some of the dictionary-lookup code like https://github.com/Unity-Technologies/ml-agents/blob/e6dff56adfb99cca5c8854f66603e1a964fcdab4/ml-agents/mlagents/trainers/ppo/policy.py#L34-L48

### DIFF
--- a/ml-agents/mlagents/trainers/config.py
+++ b/ml-agents/mlagents/trainers/config.py
@@ -1,31 +1,93 @@
 import attr
-from typing import Dict
+from functools import partial
+from typing import Any, Callable, Dict, TypeVar, Type
+
+T = TypeVar("T")
+
+
+def generate_subclass_mapping(base_cls):
+    """
+    Takes a base class, and constructs a mapping from the names of its sublclasses to the subclass type. For example
+
+        class IceCream:
+            pass
+
+        class ChocolateIceCream(IceCream):
+            pass
+
+        class VanillaIceCream(IceCream):
+            pass
+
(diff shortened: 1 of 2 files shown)


ml-agents/mlagents/trainers/config.py
```diff
@@ -0,0 +1,58 @@
...
+}
+
+@attr.s(auto_attribs=True)
+class TrainerParameters:
+    trainer: str = "ppo"  # TODO enum
+    batch_size: int = 1024
+    beta: float = 5.0e-3
+    buffer_size: int = 10240
+    epsilon: float = 0.2
+    gamma: float = .99  # TODO is this the current default?
```
ervteng: We can remove this as it's defined in the reward signals

chriselion: Ah, I think a few configs are still passing it in :)

ervteng:
This looks pretty good. One consideration is that each trainer type (e.g. ppo, sac, online_bc, etc.) has different parameters and different defaults. So maybe we need a different TrainerParameters object for each type rather than one for all.

ml-agents/mlagents/trainers/config.py
ervteng: Uh oh, that's a bug - I removed them and pushed it to #2296

PR #2315, opened 2019-07-22
Cleanup unused code in TrainerController

harperj:
* Removes unused SubprocessEnvManager import in trainer_controller
* Removes unused `steps` argument to `TrainerController._save_model`
* Consolidates unnecessary branching for curricula in
  `TrainerController.advance`
* Moves `reward_buffer` into `TFPolicy` from `PPOPolicy` and adds
  `BCTrainer` support so that we don't have a broken interface /
  undefined behavior when BCTrainer is used with curricula.

### DIFF
--- a/ml-agents/mlagents/trainers/ppo/trainer.py
+++ b/ml-agents/mlagents/trainers/ppo/trainer.py
@@ -32,9 +32,7 @@ class PPOTrainer(Trainer):
         :param seed: The seed the model will be initialized with
         :param run_id: The identifier of the current run
         """
-        super(PPOTrainer, self).__init__(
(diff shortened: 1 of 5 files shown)


ml-agents/mlagents/trainers/ppo/trainer.py
```diff
@@ -32,7 +32,9 @@ def __init__(
         :param seed: The seed the model will be initialized with
         :param run_id: The identifier of the current run
         """
-        super(PPOTrainer, self).__init__(brain, trainer_parameters, training, run_id)
+        super(PPOTrainer, self).__init__(
```
chriselion: nit: while you're in the neighborhood, clean this up to `super().__init__`

PR #2319, opened 2019-07-23
Add markdown link check to precommit hooks

chriselion:
Automates part of the release process. Expecting this to fail in the short term; need to whitelist some things (like localhost:6006 for tensorboard)

### DIFF
--- a/docs/localized/zh-CN/README.md
+++ b/docs/localized/zh-CN/README.md
@@ -32,7 +32,7 @@ TensorFlow），让游戏开发者和业余爱好者能够轻松地
 **除了安装和使用说明外，如需更多信息，
 请参阅我们的[文档主页](docs/Readme.md)。**如果您使用的
 是 v0.3 之前的 ML-Agents 版本，强烈建议您参考
-我们的[关于迁移到 v0.3 的指南](/docs/Migrating.md)。
+我们的[关于迁移到 v0.3 的指南](../../../docs/Migrating.md)。
 
 我们还发布了一系列与 ML-Agents 相关的博客文章：
 - reinforcement learning（强化学习）概念概述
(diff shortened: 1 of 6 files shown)


docs/localized/zh-CN/README.md
```diff
@@ -32,7 +32,7 @@ TensorFlow），让游戏开发者和业余爱好者能够轻松地
 **除了安装和使用说明外，如需更多信息，
 请参阅我们的[文档主页](docs/Readme.md)。**如果您使用的
 是 v0.3 之前的 ML-Agents 版本，强烈建议您参考
-我们的[关于迁移到 v0.3 的指南](/docs/Migrating.md)。
+我们的[关于迁移到 v0.3 的指南](../../../docs/Migrating.md)。
```
chriselion: @ervteng @unityjeffrey Let me know what you think about this - to "officially" get the link checker running, we need to have the links from the localized docs point somewhere, because they doc they do point to hasn't been localized. I figure this is better since it's at least explicit in the markdown that it's linking somewhere outside the localization.

Alternative is just whitelist the localized docs and skip the checks on then.

chriselion:
To run full check locally:
`git ls-files | grep \.md$ | grep -v localized | xargs -n1 -t markdown-link-check -c markdown-link-check.config.json`
This currently only checks a subset of files - pre-commit splits them into batches, but `markdown-link-check` only runs on the first file in the batch (because it doesn't take multiple files as arguments).

chriselion:
Summary:
* Fixed broken links in https://github.com/Unity-Technologies/ml-agents/pull/2327
* Will update release instructions with full commandline
* Blocked on either https://github.com/pre-commit/pre-commit/issues/394 or https://github.com/tcort/markdown-link-check/issues/78 to implement this in pre-commit right now.

PR #2325, opened 2019-07-24
Profiling docs

chriselion:
Docs on the profiling system. Not sure if there's a good place to link to them from.

### DIFF
--- a/docs/Profiling.md
+++ b/docs/Profiling.md
@@ -1,6 +1,6 @@
 # Profiling ML-Agents in Python
 
-ml-agents provides a lightweight profiling system, in order to identity hotspots in the training process and help spot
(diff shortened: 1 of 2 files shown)


ervteng:
Maybe we can link them in Training-ML-Agents.md - since that's where the CSV is mentioned.

docs/Profiling.md
```diff
@@ -0,0 +1,53 @@
+# Profiling ML-Agents in Python
+
+ml-agents provides a lightweight profiling system, in order to identity hotspots in the training process and help spot
```
ervteng: Capitalize ML-Agents

chriselion:
Actually the docs on `--debug` are wrong - it always outputs the csv, and setting the flag only changes the log level. I'll update that too.

PR #2329, opened 2019-07-25
Add "gauges" to timer system

chriselion:
This adds the ability to track arbitrary values in the timer system. For a given key, we only track the most recent value (as well as the min and max values over all time). 

These is conceptually the same as gauges in statsd (see https://statsd.readthedocs.io/en/v0.5.0/types.html#gauges) so that's what I called. I'm definitely open to another name, though - I misspell it as "guage" more than 50% of the time.

Note that these aren't hierarchical like timers are; I'm not sure if makes sense in this case.

One missing piece at the moment is merging these from subprocesses (same as we do with timers). If this looks good otherwise, I'll add that.


ml-agents/mlagents/trainers/trainer.py
```diff
@@ -228,6 +229,7 @@ def write_summary(
                         is_training,
                     )
                 )
+                set_gauge(f"{self.brain_name}.mean_reward", mean_reward)
```
chriselion: @xiaomaogy This is what you mentioned you wanted to track. In the output json, it adds an entry like
```
  "gauges": [
    {
      "name": "HallwayLearning.mean_reward",
      "value": -0.5,
      "min": -1.0
      "max": -0.5,
      "count": 3
    }
  ],
```

xiaomaogy: This looks exactly what I need. Is this PR ready to be merged?

chriselion: Sure, like I mentioned, I didn't add support for gauges from subprocesses yet, but that should be OK for your purpose. Let me just add a TODO.

PR #2331, opened 2019-07-25
Tick version number for 0.9

ervteng:
- Tick version for Python packages
- Tick communication API to API-9

### DIFF
--- a/UnitySDK/Assets/ML-Agents/Scripts/Academy.cs
+++ b/UnitySDK/Assets/ML-Agents/Scripts/Academy.cs
@@ -95,7 +95,7 @@ namespace MLAgents
         [SerializeField]
         public BroadcastHub broadcastHub = new BroadcastHub();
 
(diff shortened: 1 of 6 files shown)


UnitySDK/Assets/ML-Agents/Scripts/Academy.cs
```diff
@@ -95,7 +95,7 @@ public abstract class Academy : MonoBehaviour
         [SerializeField]
         public BroadcastHub broadcastHub = new BroadcastHub();
 
-        private const string kApiVersion = "API-8";
+        private const string kApiVersion = "API-9";
```
chriselion: What's the procedure on these? We didn't actually change the API for this release, right?

xiaomaogy: We usually just change the API along with the major release (even though they are still compatible).

PR #2345, opened 2019-07-26
Improve test_simple.py and check discrete actions

chriselion:
* Use pytest.mark.parametrize to run multiple variants of the same basic test
* Pick a random goal each "run" and pass that as an observation (instead of always going to +1)
* Add a variant that covers discrete actions


ml-agents/mlagents/trainers/tests/test_environments/test_simple.py
```diff
@@ -94,7 +103,7 @@ def reset(
         self._reset_agent()
 
         agent_info = AgentInfoProto(
-            stacked_vector_observation=[self.position] * OBS_SIZE,
+            stacked_vector_observation=[self.goal] * OBS_SIZE,
```
ervteng: Shouldn't the vector observation also include the position in some way (maybe concat the position and goal)? I suppose it will still train if just the goal is given by taking the same action always.

ervteng:
I think we should move the actual running of the test to the test_ppo.py file - what do you think?

ml-agents/mlagents/trainers/tests/test_environments/test_simple.py
chriselion: The way this is set up, the agent always starts at 0 and gets a reward at +/- 1, so the "best" policy is always action=observation[0] (for continuous) or (0 if observation[0] < 0 else 1) for discrete.

chriselion:
Right now it's only testing PPO, but I think it's general enough to test other combinations of settings (memory, visual obs) and other trainers (SAC).

ml-agents/mlagents/trainers/tests/test_environments/test_simple.py
ervteng: Got it. So it should train either way.

xiaomaogy:
I guess it is testing the rl aspect of the environment, so either rename it something like test_simple_rl or move it to test_ppo? The PR looks good to me, it is ok to merge.

chriselion:
Sounds good, I renamed to test_simple_rl.py and moved it to the main `trainers/tests` directory.

PR #2360, opened 2019-07-29
Refactor Trainer and Model

ervteng:
Refactor trainer, model in preparation for SAC. 

Specifically:
- Move common functions to `trainer.py`, model.py` from `ppo/trainer.py`, `ppo/policy.py` and `ppo/model.py'
- Introduce `RLTrainer` class and move most of `add_experiences` and some common reward 
signal code there. PPO and SAC will inherit from this, not so much BC Trainer.
- Add methods to Buffer to enable sampling, truncating, and save/loading. 
- Add scoping to create encoders in `model.py`


ml-agents/mlagents/trainers/buffer.py
```diff
@@ -240,6 +282,22 @@ def reset_update_buffer(self):
         """
         self.update_buffer.reset_agent()
 
+    def truncate_update_buffer(self, max_length):
+        """
+        Truncates the update buffer to a certain length.
+
+        The skew makes a bigger truncation each time, hopefully requiring less truncations overall.
```
awjuliani: What is the skew?

ml-agents/mlagents/trainers/trainer.py
```diff
@@ -268,3 +281,237 @@ def write_tensorboard_text(self, key, input_dict):
                 "Cannot write text summary for Tensorboard. Tensorflow version must be r1.2 or above."
             )
             pass
+
+
+class RLTrainer(Trainer):
```
awjuliani: Does it make sense to make this its own file?

ml-agents/mlagents/trainers/buffer.py
ervteng: Seems like an old comment from an older implementation - removed.

ml-agents/mlagents/trainers/trainer.py
ervteng: Was debating this - I think it does, but I was also concerned with creating more files. Seems like it's long enough to break off, though.

ervteng: Done

ml-agents/mlagents/trainers/models.py
```diff
@@ -407,7 +409,7 @@ def create_discrete_action_masking_layer(all_logits, action_masks, action_size):
             for i in range(len(action_size))
         ]
         raw_probs = [
-            tf.multiply(tf.nn.softmax(branches_logits[k]) + 1.0e-10, branch_masks[k])
+            tf.multiply(tf.nn.softmax(branches_logits[k]) + 1.0e-7, branch_masks[k])
```
harperj: Can you explain the change from `1.0e-10` to `1.0e-7` here and below?

ml-agents/mlagents/trainers/models.py
```diff
@@ -438,12 +438,15 @@ def create_observation_streams(
         h_size: int,
         num_layers: int,
         vis_encode_type: EncoderType = EncoderType.SIMPLE,
+        stream_scopes: List[str] = [],
```
harperj: We shouldn't use default values of `[]` -- if two separate models are created using the defaults and one changes the value of `stream_scopes`, it will change for the other.

ml-agents/mlagents/trainers/models.py
ervteng: In SAC I found that 1.0e-10 was too small to prevent NaNs. I'm changing it here to be safe, though I don't think there were any known NaNs for PPO.

ml-agents/mlagents/trainers/models.py
ervteng: Fixed, changed to None

ervteng:
Fixed a couple bugs (mostly to do with SAC) I've tested with 3DBall, Hallway, and GridWorld.

PR #2362, opened 2019-07-29
Update reward signals in parallel with policy

ervteng:
Updates reward signal in parallel with policy. This means that all batching must be handled by the Trainer, and not the reward signal (loses some generality). However, it produces quite a bit performance boost in training and a lot less code. 

Note: I'm seeing about a 20-30% speedup using Curiosity (or GAIL + Curiosity) on CPU. I expect it to be bigger on GPU - would want to test before merging. 

Slight change of behavior: Reported policy loss is not the absolute value of policy loss anymore.


ml-agents/mlagents/trainers/components/reward_signals/curiosity/signal.py
```diff
@@ -171,5 +147,5 @@ def _update_batch(
                     feed_dict[self.model.next_visual_in[i]] = _obs
 
         self.has_updated = True
-        run_out = self.policy._execute_model(feed_dict, self.update_dict)
-        return run_out
+        # run_out = self.policy._execute_model(feed_dict, self.update_dict)
```
harperj: 🤔

ervteng: Should probably delete the comment - `feed_dict` is right. I also renamed this method to `prepare_update`.

ml-agents/mlagents/trainers/ppo/trainer.py
```diff
@@ -451,19 +452,26 @@ def update_policy(self):
...
-            )
-            for stat, val in update_stats.items():
-                self.stats[stat].append(val)
+                mini_batch = buffer.make_mini_batch(start, end)
+
+                feed_dict = {}
+                update_dict = self.policy.update_dict
+                feed_dict.update(self.policy.prepare_update(mini_batch, n_sequences))
+                stats_needed = self.policy.stats_name_to_update_name
+                for _, reward_signal in self.policy.reward_signals.items():
```
dongruoping: Will it be better moving this part (constructing feed_dict and update_dict) into policy, since all the reward_signals and stats_name_to_update_name are stored in policy? Maybe keep the policy update function and move it there?

ervteng: Good idea, I think bringing back the update() method makes sense. This would also make it so that we don't have to change the BC trainer to keep the Policy interface the same, since the BC trainer has no reward signals.

ervteng:
> Code looks good to me -- my only thought is about our discussion of this tying us to certain numbers of epochs, ordering of updates, etc. on each of the reward signals and that removing flexibility. On the other hand, maybe that's just an implementation detail best left up to the trainer. In any case, that's a bridge to cross when we come to it I think-- 🚢 🇮🇹

One thought I had about this - the user really shouldn't be touching them IMO. For instance, setting num_epochs to > 1 for GAIL in SAC breaks training. This change will make it easier for us to enforce "good defaults" across trainers, which at the end of the day might be better.

ervteng:
Turns out this broke multi-GPU. We're working on a fix - will wait until it's done before pushing.

ml-agents/mlagents/trainers/ppo/multi_gpu_policy.py
```diff
@@ -84,6 +68,82 @@ def create_model(self, brain, trainer_params, reward_signal_configs, seed):
...
+                        for reward_signal, config in reward_signal_configs.items():
+                            reward_tower[reward_signal] = create_reward_signal(
+                                self, self.towers[device_id], reward_signal, config
+                            )
+                            for k, v in reward_tower[reward_signal].update_dict.items():
+                                self.update_dict[k + "_" + str(device_id)] = v
+                        self.reward_signal_towers.append(reward_tower)
+                for _, reward_tower in self.reward_signal_towers[0].items():
+                    for _, update_key in reward_tower.stats_name_to_update_name.items():
+                        self.update_dict.update(
```
harperj: Minor but this statement is 13 lines long.  I think we should have some intermediate variable names to make it easier to parse.

PR #2371, opened 2019-07-30
Return list instead of np array for make_mini_batch()

dongruoping:
Not sure if it's a good idea to change this and if it'll break anything, but avoiding creating the np.array speeds up a lot and also we can get rid of the reshaping for feed_dict.


harperj:
Looks like this broke some tests.  We'll need to fix them but otherwise this looks like a really good change to me.  Even with the really small buffer used for Hallway I'm seeing a ~15% improvement and of course you have seen an even bigger one with Obstacle Tower.

dongruoping:
> Looks like this broke some tests. We'll need to fix them but otherwise this looks like a really good change to me. Even with the really small buffer used for Hallway I'm seeing a ~15% improvement and of course you have seen an even bigger one with Obstacle Tower.

I only made changes to ppo so tests for other models failed. I'll fix them soon.

harperj:
Any update on the tests here?  I'd also update this from a draft PR to a normal PR.

PR #2373, opened 2019-07-31
Crawler Relative Observations

RunSwimFlyRich:
Observations for the static and dynamic target seeking crawlers are now made relative to the vector joining crawler to target rather than in world space. 
Height determined by raycast rather than absolute to fix behaviour on raised platforms.
Crawler body mass increased to 20.

Apologies for the lack of commit messages. I've not used GitHub much before. I couldn't get the desktop client to authorise my commits so I resorted to the Web Interface and had to commit each folder separately. I'd thought I might be able to add a commit message with the pull request or at least edit the messages later.

### DIFF
--- a/UnitySDK/Assets/ML-Agents/Examples/Crawler/Brains/CrawlerDynamicLearning.asset
+++ b/UnitySDK/Assets/ML-Agents/Examples/Crawler/Brains/CrawlerDynamicLearning.asset
@@ -3,8 +3,9 @@
 --- !u!114 &11400000
 MonoBehaviour:
   m_ObjectHideFlags: 0
-  m_PrefabParentObject: {fileID: 0}
-  m_PrefabInternal: {fileID: 0}
+  m_CorrespondingSourceObject: {fileID: 0}
+  m_PrefabInstance: {fileID: 0}
+  m_PrefabAsset: {fileID: 0}
   m_GameObject: {fileID: 0}
   m_Enabled: 1
   m_EditorHideFlags: 0
@@ -12,7 +13,7 @@ MonoBehaviour:
   m_Name: CrawlerDynamicLearning
   m_EditorClassIdentifier: 
   brainParameters:
-    vectorObservationSize: 129
+    vectorObservationSize: 126
     numStackedVectorObservations: 1
     vectorActionSize: 14000000
     cameraResolutions: []
@@ -38,4 +39,5 @@ MonoBehaviour:
     - 
     - 
     vectorActionSpaceType: 1
-  model: {fileID: 11400000, guid: 94d1889cee00f4361b74cbc4eb129f11, type: 3}
+  model: {fileID: 5022602860645237092, guid: 94d1889cee00f4361b74cbc4eb129f11, type: 3}
+  inferenceDevice: 0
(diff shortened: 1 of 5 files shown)


awjuliani:
Hi @RunSwimFlyRich 

Thanks for making this PR. Have you tested training these models again, to compare performance?

RunSwimFlyRich:
Yes, the trained models for static and dynamic target seeking are part of the commit. 
The static one runs at the same speed as the old one but will now actually run to the designated target regardless of direction (the old version broke if the target direction was different to the one it was exposed to during training.)
The dynamic version gets to targets a little faster and looks more natural than the old version as it reliably turns to face the target rather than strafing and crab walking.

Hunter-Unity:
Thanks for the PR! i'm having a look at the project. The walk/body rotation def looks better to me. Can you tell me a bit more about training times? Do you know how many steps this CrawlerDynamicTraining model was trained for? Less than 1M steps?

RunSwimFlyRich:
I've attached the log which was fortuitously still available in my Terminal window. It was trained for a reported total of 261,000 steps though it had plateaued by step 200,000.  However I was running 5 environments. Once again I'm new to this but based upon the rewards I've seen the actual number of steps needs to be multiplied by the number of environments. So based on that assumption it was trained for 1.3 million steps but had plateaued by step 1 million.

On a previous 5 environment run it hit a maximum mean reward of 672 at reported step 261,000 but then fell into irretrievable decline around step 400,000.  

Note that I've lifted the mass of the body object from 1 to 20 (part of the commit) for all of these. 

In a seperate uncommitted project I've increased default friction. I've also trialed training with increased joint strength or decreased decision interval time and even scaling the model up to try to get more speed but to no avail. Still with a simulated speed of 80km/h the static version might just be fit for my purposes.

Also in this seperate project I've found that I can retain the same performance by setting the Physics Project Settings Default Solver Iterations and Velocity Iterations back to their defaults of 6 and 1 respectively (ultimately I'm running on mobile and possibly Oculus Quest) and then lifting these values back to 12 and 12 for the forelegs in code.

[DynamicTraining.pdf](https://github.com/Unity-Technologies/ml-agents/files/3454269/DynamicTraining.pdf)

Hunter-Unity:
Thanks for the info. I'm finishing up my code review. I'll have some suggested changes but overall i think the new observation vectors are working much better.

RunSwimFlyRich:
Sure thing. I've made numerous changes to my local project but the only change I'd want to commit is to parent my workingTransform under its CrawlerAgent to stop it spamming the hierarchy view and leaving gameObjects behind when an agent is destroyed. I'll leave it for now though seeing as you have some changes.
This stuff is a lot of fun to play around with. I was initially looking to just add some quadrupedal robots to my game but I think I'll be adding a number of different contraptions and configurations now.

RunSwimFlyRich:
Thank you for the feedback. I’ve committed the requested change in CrawlerAgent. There are conflicts with the model data files that have arisen in the time since I made my pull request which I'm not authorised to resolve. Please let me know if there’s anything else I need to do.

Hunter-Unity:
great i'll have a look. thanks!

Hunter-Unity:
confirmed new Matrix4x4 code works. if you don't have any objections @awjuliani lets merge this to develop. thanks again for the PR @RunSwimFlyRich. please keep us updated on your experiments with ML-Agents! :D

RunSwimFlyRich:
Great. Currently playing around with allowing them to evolve their morphology while they are learning. A preliminary experiment saw them develop a galloping gait rather than the trot they currently employ. Will let you know how I go.

PR #2376, opened 2019-08-01
Fix training not quitting when play button is unchecked

ervteng:


harperj:
This should probably be squashed and we should add a good commit message.  Glad to take care of that, just give a 👍 if you're ready for it to be merged.

PR #2400, opened 2019-08-07
Fix issue with visual obs were destroyed too early

ervteng:
- Was causing issues with Demonstration Recorder where visual observations were deleted, then attempted to access. 
- Seems like my editor also deleted all trailing white spaces, causing many lines of change.


UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -8,7 +8,7 @@
 namespace MLAgents
 {
     /// <summary>
-    /// Struct that contains all the information for an Agent, including its
```
ervteng: Seems like my editor deleted trailing white spaces.

PR #2412, opened 2019-08-08
Move trainer initialization into a utility function

harperj:
This change moves trainer initialization outside of TrainerController,
reducing some of the constructor arguments of TrainerController and
setting up the ability for trainers to be initialized in the case where
a TrainerController isn't needed.

### DIFF
--- a/ml-agents/mlagents/trainers/trainer_util.py
+++ b/ml-agents/mlagents/trainers/trainer_util.py
@@ -1,0 +1,95 @@
+from typing import Any, Dict
+
+from mlagents.trainers import MetaCurriculum
+from mlagents.envs.exception import UnityEnvironmentException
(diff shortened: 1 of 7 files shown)


ml-agents/mlagents/trainers/trainer_util.py
```diff
@@ -0,0 +1,95 @@
...
+def initialize_trainers(
+    trainer_config: Dict[str, Any],
+    external_brains: Dict[str, BrainParameters],
+    summaries_dir: str,
+    run_id: str,
+    model_path: str,
+    keep_checkpoints: int,
+    train_model: bool,
+    load_model: bool,
+    seed: int,
```
ervteng: Should we have a default seed?

Issue #1402, opened 2018-11-05
Memory Leak on Linux


atapley:
Hi,

I have been using Unity ml-agents on my Mac to create environments. I then build the environment for Linux and upload it to git in order to pull the environment down on a Linux machine. When I use the environment on the Linux machine, I run the learn.py algorithm, but it continues to use more and more memory until it runs out and it crashes. 

I am running on Ubuntu 16.04 with ml-agents 0.4.0b and Unity 2017.2.0f1.

I know this was an issue with textures in the past, but the version of ml-agents that I have is after this issue was fixed.

icaro56:
My ML version 0.5 is also has this problem in Linux. The memory increase, increase ....

xiaomaogy:
This is a known issue that is on our to fix bug list.

valanchik:
Resources.UnloadUnusedAssets();

arixlin:
@xiaomaogy  is fix bug on Version 0.6  ?

xiaomaogy:
@arixlin Not yet.

arixlin:
@xiaomaogy  Thanks!

tjad:
any pointers to what the cause of this is ?

atapley:
In my opinion it could either be an issue on Unity's end with the visual observations, or an issue with Tensorflow itself. I have seen a few things saying that Tensorflow has a memory leak like this. I don't have the issue with vector observations though, so I'm not sure.

tjad:
@atapley Thanks for the fast reply. Yes, I suspected TF too, but will need to do some profiling to confirm that. The unity process (executable) itself seems stable, the python side is doing horrible things. Will post updates.

tjad:
Using mlagents 0.7 with TF 1.13.1 results with the same memory leak. Could be a TF usage error - not freeing memory when it should be freed from the session? 

Onwards...

tjad:
Why I say it may be a usage error, is that utilizing rainbow (dopamine) with the same environment does not result in the memory leak. So it can't be TF alone - well it could, but not likely ?

tjad:
@xiaomaogy let me in on any knowns - perhaps including where this bug is being managed please-, would like to keep eye on progress and provide input.

tjad:
From what I can tell, it's got to do with the trainer's accumulation of experiences

tjad:
Located the problem. Will see if I can get to the bottom of this and submit a PR.

[...]

PR #2413, opened 2019-08-08
Avoid memory leak during evalution by not filling vis obs training buffer during evaluation (#1402)

tjad:


ervteng:
Hey @tjad, this will work for visual obs, but not in general. We'll look into a better solution for this issue. In the meantime, you can use Barracuda to do inference without memory leaks.

tjad:
@ervteng Yes, I understand. It appeared at the time that this was solely responsible for eating the majority of the memory (and when applying this patch, I could not reproduce.

I do see that you've changed the composition of PPO considerably too.

Closing for now. - bear in mind (https://github.com/Unity-Technologies/ml-agents/issues/1402) is still open

PR #2439, opened 2019-08-15
normalized crawler raycast hit dist and added some code comments

Hunter-Unity:


ervteng:
Side-concern - it seems like on the current Crawler on `develop` the scenes aren't pointing to the right prefabs, and they're all missing. Maybe we can update the scene files in this PR (or another one)?

chriselion:

surfnerd:
@Hunter-Unity, is this something you still want to merge?

PR #2443, opened 2019-08-16
Update development release version to 0.10.0.dev0

harperj:
In order for downstream packages to make use of the latest
pre-release features, we can pre-release versions of our packages.
For packages ending in `devN` pip will not install that package
version by default.  This change manually updates our package version
to a development version with the idea that we can manually perform
development versions with the potential for future automated / nightly
dev releases.

### DIFF
--- a/gym-unity/setup.py
+++ b/gym-unity/setup.py
@@ -4,12 +4,12 @@ from setuptools import setup, find_packages
 
 setup(
     name="gym_unity",
-    version="0.4.4",
+    version="0.4.5.dev0",
     description="Unity Machine Learning Agents Gym Interface",
     license="Apache License 2.0",
     author="Unity Technologies",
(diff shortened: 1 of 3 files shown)


ervteng:
Looks fine to me - I'd just update the release checklist to be sure to remove the `dev` from the `setup.py`. 

What do you think about auto-ticking the dev version, e.g. https://the-hitchhikers-guide-to-packaging.readthedocs.io/en/latest/specification.html#development-releases?

harperj:
@ervteng I was thinking of this as a stepping stone towards automatically ticking the development version -- we should confirm the process works manually and then potentially add it to an automated process.  I think it's OK if we have `0.10.devX` be at the same position as `0.9.2` if something changes with our release schedule.

Will update the release checklist.

PR #2461, opened 2019-08-22
Reduce update buffer float precision

ervteng:
Update buffer stores everything as float64. This is unnecessary, as TF defaults to using float32 anyways. In addition, it balloons to a ridiculous size with large buffer sizes (> 100,000). Seems to also produce a small speedup (<10%) in the policy update. 

This PR modifies `get_batch` to truncate everything to float32 before placing it in the update buffer.


harperj:
Changes LGTM -- but do you think we should do this at the BrainInfo level as well?

ervteng:
> Changes LGTM -- but do you think we should do this at the BrainInfo level as well?

Possibly - maybe that's the better place to do it? So it's float32 all the way through, and we don't have to explicitly change it here in the buffer.

PR #2473, opened 2019-08-28
python coverage: specify dirs, exclude test files

chriselion:
* When running coverage checks locally, I noticed it was excluding some files from counting (in particular, buffer.py). This seems to be a better way to specify and includes the expected files.
* Save the coverage results as html and upload them as artifacts in circleci. This is a bit noisy, but the main coverage results can be found in the `index.html` file.
* Exclude the test files from coverage calculations.

Example coverage result: https://1701-102904613-gh.circle-artifacts.com/0/htmlcov/index.html

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,7 +37,7 @@ jobs:
           command: |
             . venv/bin/activate
             mkdir test-reports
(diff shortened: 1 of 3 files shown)


xiaomaogy:
LGTM

Issue #1835, opened 2019-03-18
'function' object has no attribute 'num_envs'


mikiwong:
Hi,

I am trying to run the unity gym wrapper +baselines example given in the docs using the Banana Collector environment. But I am encountering this error:
'AttributeError: 'function' object has no attribute 'num_envs' 
in line 91, in ppo2.py. 

I am using baselines 0.1.5 and ml-agents 0.7.0

ervteng:
Hi @mikiwong, baselines continuously changes their API, so our guide might indeed be broken. Looking at the code, it looks like you can solve this by simply setting nenvs in `ppo2.py` to the number of environments you're spawning, and not trying to read it from the env object.

xiaomaogy:
Thanks for reaching out to us. Hopefully you were able to resolve your issue.  We are closing this due to inactivity, but if you need additional assistance, feel free to reopen the issue.

ryujaehun:
I am also encountering the error like ```AttributeError: 'function' object has no attribute 'num_envs'``` when  [gym-unity](https://github.com/Unity-Technologies/ml-agents/tree/master/gym-unity) 
This problem was solved as follows.
```
env=env => env=env()
```

qfettes:
This is still an issue, but has a simple fix. Baselines expects the environments to be wrapped as SubprocessVecEnv or DummyVecEnv. Thus, change the make_unity_env function in the tutorial to:

```
from gym_unity.envs import UnityEnv

from baselines.common.vec_env.subproc_vec_env import SubprocVecEnv
from baselines.common.vec_env.dummy_vec_env import DummyVecEnv
from baselines.bench import Monitor
from baselines import logger
import baselines.ppo2.ppo2 as ppo2

try:
    from mpi4py import MPI
except ImportError:
    MPI = None

def make_unity_env(env_directory, num_env, visual, start_index=0):
    """
    Create a wrapped, monitored Unity environment.
    """
    def make_env(rank, use_visual=True): # pylint: disable=C0111
        def _thunk():
            env = UnityEnv(env_directory, rank, use_visual=use_visual, uint8_visual=True)
            env = Monitor(env, logger.get_dir() and os.path.join(logger.get_dir(), str(rank)))
            return env
        return _thunk
    if visual:
        return SubprocVecEnv([make_env(i + start_index) for i in range(num_env)])
    else:
        rank = MPI.COMM_WORLD.Get_rank() if MPI else 0
[...]

PR #2489, opened 2019-09-03
This addresses #1835. Baselines expects single environments used with…

qfettes:
… their ppo2 algorithm to be wrapped in a DummyVecEnv. The old readme did not instruct the reader to do so and the code failed to run with the latest version of baselines. This imports the correct function from baselines and fixes the make_unity_env function described in the readme.


ervteng:
@qfettes thank you again for submitting the PR. Can you target `develop` instead of `master`? I can merge it after that. Thanks!

qfettes:
Good idea. I switched the target branch and added a note about the baselines version I used in the readme.

Issue #2418, opened 2019-08-09
'command line arguments' feature does not work 


Sohojoe:
**Describe the bug**
The new feature 'support for command line arguments' #2243 does not work

**To Reproduce**
`mlagents-learn config/my_config.yaml --train  --run-id=my-id  --my-custom-opton=value`

**Environment (please complete the following information):**
- _ML-Agents version_: v0.9

**Notes / Observations**
* learn.py does not support options which are not defined in the `_USAGE` string. Its not clear if docopt supports this.
* learn.py does not pass options through to UnityEnvironment via `args`

ervteng:
Hi @Sohojoe, big thanks for spotting this issue! It's an error on our part in the release notes - the PR enabled the parameters to be passable when creating a `UnityEnvironment` class, but it was never intended to pass down through `learn.py`. We're looking into doing this properly - but basically I would add a parameter to the docopt, turn it into a list of strings, and pass it into the `create_environment_factory` method in `learn.py`.

chriselion:
This is fixed on develop and will be available in the next release. See https://github.com/Unity-Technologies/ml-agents/pull/2594

PR #2491, opened 2019-09-04
Added option to use environment arguments in learn

tomatenbrei:
#### Detailed description of the changes performed

Environment arguments passed to `mlagents-learn` via `--env-args` are now passed to the environment executables.

#### Corresponding changes to documentation, unit tests and sample environments (if applicable)

I added the new argument to `docs/Training-ML-Agents.md`.

#### Summary of the tests performed to validate your changes

I used a windows build of the 'Basic' example environment to check that it is working. The passed arguments can be accessed with `System.Environment.GetCommandLineArgs()`. Tested with single (`--num-envs=1`) and multiple (`--num-envs=4`) executables.

For example:

The argument `--env-args="--num 1 --str 'Hello World' --foo"` causes `string.Join(", ", System.Environment.GetCommandLineArgs())` to return `"BUILD_PATH\Unity Environment.exe, --port, 5005, --num, 1, --str, Hello World, --foo"`.

Without that argument, `string.Join(", ", System.Environment.GetCommandLineArgs())` returns `"BUILD_PATH\Unity Environment.exe, --port, 5005"`.

#### Issue numbers that the PR resolves (if any)

Solves https://github.com/Unity-Technologies/ml-agents/issues/2418.


ervteng:
Hey @tomatenbrei, thank you for submitting this change! The functionality seems good to me. If possible would it be OK to add an example of args to the help string? e.g. `--env-args="--arg1=2 --arg2=3"` 

Would help people format them properly when using it. Thanks!

tomatenbrei:
Hey @ervteng!

Please note that the provided implementation does not include argument parsing (aside from not splitting strings at whitespaces). It basically only forwards the value of `--env-args` to the executable environment(s). So there is actually no "proper" format in that sense.

The simplest argument format (which is straightforward and does not require any "complicated" parsing routines inside Unity) looks like

````--env-args="--arg1 2 --arg2 'string value'"````

with spaces between the arguments name and value. This allows you to manually parse arguments just like it is done with the the native `--port` parameter used by the `Academy` class inside `ReadArgs`.

I think the type `string` in `--env-args=<string>` already implies that one should use quotation marks when using multiple environment arguments. Aside from that, there are no real formatting constraints.

However, I can still add an example to the help text and/or documentation, if you find that helpful!

chriselion:
Hi @tomatenbrei,
Taking over this from @ervteng 
I played around with docopt a bit, and I think this is slightly cleaner by allowing repeated arguments. Here's an [example](
http://try.docopt.org/?doc=++++Usage%3A%0D%0A++++++mlagents-learn+%3Ctrainer-config-path%3E+%5B--env-args%3D%3Cstring%3E...%5D+%5Boptions%5D%0D%0A++++++mlagents-learn+--help%0D%0A%0D%0A++++Options%3A%0D%0A++++++--env%3D%3Cfile%3E++++++++++++++++Name+of+the+Unity+executable+%5Bdefault%3A+None%5D.%0D%0A++++++--env-args%3D%3Cstring%3E+++++++++Arguments+of+the+Unity+executable.%0D%0A++++++--curriculum%3D%3Cdirectory%3E++++Curriculum+json+directory+for+environment+%5Bdefault%3A+None%5D.%0D%0A++++++--sampler%3D%3Cfile%3E++++++++++++Reset+parameter+yaml+file+for+environment+%5Bdefault%3A+None%5D.%0D%0A++++++--keep-checkpoints%3D%3Cn%3E++++++How+many+model+checkpoints+to+keep+%5Bdefault%3A+5%5D.%0D%0A++++++--lesson%3D%3Cn%3E++++++++++++++++Start+learning+from+this+lesson+%5Bdefault%3A+0%5D.%0D%0A++++++--load++++++++++++++++++++++Whether+to+load+the+model+or+randomly+initialize+%5Bdefault%3A+False%5D.%0D%0A++++++--run-id%3D%3Cpath%3E+++++++++++++The+directory+name+for+model+and+summary+statistics+%5Bdefault%3A+ppo%5D.%0D%0A++++++--num-runs%3D%3Cn%3E++++++++++++++Number+of+concurrent+training+sessions+%5Bdefault%3A+1%5D.%0D%0A++++++--save-freq%3D%3Cn%3E+++++++++++++Frequency+at+which+to+save+model+%5Bdefault%3A+50000%5D.%0D%0A++++++--seed%3D%3Cn%3E++++++++++++++++++Random+seed+used+for+training+%5Bdefault%3A+-1%5D.%0D%0A++++++--slow++++++++++++++++++++++Whether+to+run+the+game+at+training+speed+%5Bdefault%3A+False%5D.%0D%0A++++++--train+++++++++++++++++++++Whether+to+train+model%2C+or+only+run+inference+%5Bdefault%3A+False%5D.%0D%0A++++++--base-port%3D%3Cn%3E+++++++++++++Base+port+for+environment+communication+%5Bdefault%3A+5005%5D.%0D%0A++++++--num-envs%3D%3Cn%3E++++++++++++++Number+of+parallel+environments+to+use+for+training+%5Bdefault%3A+1%5D%0D%0A++++++--docker-target-name%3D%3Cdt%3E+++Docker+volume+to+store+training-specific+files+%5Bdefault%3A+None%5D.%0D%0A++++++--no-graphics+++++++++++++++Whether+to+run+the+environment+in+no-graphics+mode+%5Bdefault%3A+False%5D.%0D%0A++++++--debug+++++++++++++++++++++Whether+to+run+ML-Agents+in+debug+mode+with+detailed+logging+%5Bdefault%3A+False%5D.%0D%0A++++++--multi-gpu+++++++++++++++++Setting+this+flag+enables+the+use+of+multiple+GPU%27s+%28if+available%29+during+training%0D%0A++++++++++++++++++++++++++++++++++%5Bdefault%3A+False%5D.&argv=mypath+--env-args+foo+--env-args%3Dbar%3D3) (with all the Unity options; please forgive the long url).

The relevant parts are 
* Add ` [--env-args=<string>...]` in the usage section
* Add `      --env-args=<string>         Arguments of the Unity executable.` to the options section (no default)

I couldn't decipher the syntax enough to get `...` working in just the options, but if you can, great.

This way you don't need to include `shlex` and do the splitting in mlagents code; it's all handled by the parser (and would also be supported by `argparse` if we ever switch to that).

Do you mind switching to this approach?

tomatenbrei:
Although using `--args` instead of  `--env-args` makes repeated arguments less verbose, I still think that splitting all env arguments into separate `mlagents-learn` arguments is not that user friendly. 

It is more verbose than the non-repeated `--env-args` approach and potentially increases the syntactic distance between argument names and values. You will need to write `--args=--arg1 --args=2` instead of `--env-args="--arg1 2"` which makes it harder to recognize that this is an argument-value pair. To overcome this problem, you could always use  `--args=--arg1=2` instead. However, this requires additional parsing in Unity.

I would prefer the `shlex.split` approach as it has less syntactic overhead and allows you to define all env arguments in one place.

chriselion:
Sorry for the back-and-forth on this. I think what I'd like to do is replace docopt with python's built-in argparse, and use [`argparse.REMAINDER`](https://docs.python.org/3/library/argparse.html#nargs) to handle the environment arguments. That keeps the usage concise, doesn't need extra parsing, and gets rid of docopt (my main motivation 😄). Does that sound OK? I'll try to get to it today or tomorrow...

tomatenbrei:
> Does that sound OK?

Yes, that sounds like the best of both worlds 😄. I already tried something similar with docopt but could not figure out a working syntax definition. It had problems with `--` and split strings at spaces. 

It looks like `argparse.REMAINDER` is designed for argument forwarding, so I guess that should work.

chriselion:
Sorry, this took longer than I expected.
https://github.com/Unity-Technologies/ml-agents/pull/2586 converts arg parsing from docopt to argparse
https://github.com/Unity-Technologies/ml-agents/pull/2587/files shows how to add this feature to the command line with the new system.

I'll get the first one approved tomorrow, and then combine the second one with this PR. We should be able to get the whole thing merged before we branch for the next release...

chriselion:
In order to make sure we got this in the next release (branching today), I made a new branch off of develop, merged your change into it, and fixed things up to use argparse. 
https://github.com/Unity-Technologies/ml-agents/pull/2594
Hope that's OK.

Closing this since the work is done.

Issue #2496, opened 2019-09-05
AttributeError: 'OnlineBCTrainer' object has no attribute 'step'


mnsmuts:
New installation of 0.9.2 with new conda py env.
-using the UnitySDK and running the 3DBall
-player brain on one of the agents Learning brain on the rest, academy with two brains (player and learner)
Nothing unusual with the setup or command (mlagents-learn config\online_bc_config.yaml --train --slow)

Obviously I changed the brain to imitate line in the default onlineBC yaml file to 3DBallPlayer
![mlagents0 9 2error](https://user-images.githubusercontent.com/22680369/64345952-26fdaf00-cfe9-11e9-87fb-56dc5774aef8.PNG)

Windows 10 ML-Agents 0.9.2 3DBall

Any ideas?

Just tried 0.9.1 with the hotfix and can get offline training working, then switched back to 0.9.2 and tried the same (with the correct env) and received the attrib error again.

mnsmuts:
![mlagents0 9 2error2](https://user-images.githubusercontent.com/22680369/64347272-878deb80-cfeb-11e9-88b7-4604e0d4449a.PNG)
Same error with offline training using the basic model and a demo file.

chriselion:
Hi @mnsmuts - thanks for the bug report. I think I see the problem, I'll work on a fix today.

chriselion:
Fixed in https://github.com/Unity-Technologies/ml-agents/pull/2498 - during a refactor, the step initialization got moved to `RLTrainer` instead of `Trainer` which explains why it didn't affect PPO training.

We'll work on some more automated testing to make sure there's better coverage on the BC Trainer code.

mnsmuts:
Many thanks

chriselion:
We're going to be doing a bugfix release on Monday with the fix for this.

chriselion:
Version 0.9.3 was released yesterday, which contains the fix for this. I'll close this issue now.

PR #2498, opened 2019-09-05
initialize trainer step count

chriselion:
This was being used by `TrainerController._not_done_training()` before it was set in `Trainer.increment_step()`

We still need better test coverage on the BC trainers, but this fixes the immediate problem.

(as reported in https://github.com/Unity-Technologies/ml-agents/issues/2496)


ml-agents/mlagents/trainers/rl_trainer.py
```diff
@@ -36,7 +36,6 @@ class RLTrainer(Trainer):
 
     def __init__(self, *args, **kwargs):
         super(RLTrainer, self).__init__(*args, **kwargs)
-        self.step = 0
```
chriselion: Removing from here since it's done in the base class now.

PR #2506, opened 2019-09-07
Add more extensive tests for BC trainer

ervteng:
Tests for step() bug as well as add_experiences and process_experiences. This brings the `bc/trainer.py` test coverage to 96%.

Does not test OnlineBCTrainer or OfflineBCTrainer, just the base class.


ervteng:
Yep, the first test breaks if `step` isn't initialized

PR #2521, opened 2019-09-09
unit test - don't use global random generator

chriselion:
(hopefully) make this test more deterministic. It's creating its own `random.Random` instance but not using it.


chriselion:
For background, we set the tf and np seeds here https://github.com/Unity-Technologies/ml-agents/blob/5ada924bab9b57d1f6fbc078dedaf07a088ace5b/ml-agents/mlagents/trainers/trainer_controller.py#L66-L67
but not python's global random seed.

The only other place this appears to be used is https://github.com/Unity-Technologies/ml-agents/blob/5ada924bab9b57d1f6fbc078dedaf07a088ace5b/ml-agents/mlagents/trainers/buffer.py#L218

PR #2526, opened 2019-09-10
Update the offline_bc_config path

xiaomaogy:


mmattar:
@xiaomaogy - I'm curious as to why we're making this change? (The PR description doesn't provide any documentation as to why.) Additionally, should we change the default path?

xiaomaogy:
@mmattar Oh I should have provided more description for this change :). This PR is to enable us to use ml-agents-cloud toolkit for the offline bc testing. We don't put the UnitySDK folder into the docker image since it is too big, that's why we point the config to demo folder, which we put into the docker image. 

It is not very necessary to change the default path, because that's the default path the demonstration recorder uses, and we refer to this path in our documentation here. https://github.com/Unity-Technologies/ml-agents/blob/3ee09631089e7e25425e1a5c1795bf995eb76ac0/docs/Training-Behavioral-Cloning.md

mmattar:
@xiaomaogy - I think having a consistent place for users to place and manage their .demo files is helpful. It is confusing to have a default directory for reading the .demo file that is different from the default directory of where the file is saved. (Esp if the reason for it is to support an internal training tool.) I would propose that we explore a new default directory that works scenarios. 

cc: @ervteng

ervteng:
To add to what @xiaomaogy says - this change also makes it so that a user could just run the example BC code as-is and have it work. 

As for the consistency issue - the basic solution seems to be just to change either the DemonstrationRecorder to save in the `demos` folder, or put the example demos in the `UnitySDK/Assets/Demonstrations` folder. I'm inclined to go with the latter. The main reason they're not in there today is that we didn't want the user to have to import these demos when they open the editor. 

But I think this is also coupled with where we put our example scenes. e.g. today, they are part of the UnitySDK folder. But ideally, if you only want to use ML-Agents as a library, you're not importing the example environments (and the demo files that go with it) into Unity. Furthermore, if you're using ML-Agents in your own project, it's unlikely that your project will be in the same directory as the Python scripts. If we were to separate these at all, I'd have the demos with the examples and not the SDK.

xiaomaogy:
@ervteng I also agree with the latter approach to move the example demos to the UnitySDK/Assets/Demonstrations folder.  Another reason why the latter approach is better is we can't write to the demos folder path from Unity, since the demos folder path is not under UnitySDK folder. 

@mmattar That's a good point, we should have a consistent place for users to place and manage their demo files. I will make a PR to change it as suggested by Ervin.

mmattar:
Agree @ervteng - a lot will change when we move to package. Can you please create an internal card to capture this? Should be something we fix as we move towards packman.

@xiaomaogy - thanks for doing a new PR. Please add me as a reviewer.

ervteng:
Trello card here: https://trello.com/c/My3kIfVq/803-move-example-demo-files-to-outside-of-unitysdk-package-when-we-shift-to-packman

Will need to be really careful such that the configs still work when we move to separate 📦 's

PR #2543, opened 2019-09-11
Develop link check xargs

chriselion:


.pre-commit-config.yaml
```diff
@@ -37,3 +37,18 @@ repos:
...
             )$
+# "Local" hooks, see https://pre-commit.com/#repository-local-hooks
+-   repo: local
+    hooks:
+    -   id: markdown-link-check
+        name: markdown-link-check
+        # markdown-link-check doesn't support multiple files on the commandline, so this hacks around that.
+        # Note that you must install the package separately via npm. For example:
+        #  brew install npm; npm install -g markdown-link-check
+        entry: bash -xc 'echo "$@" | xargs -n1 -t markdown-link-check -c markdown-link-check.config.json' --
```
chriselion: Cribbed from https://github.com/pre-commit/pre-commit/issues/394#issuecomment-421841513

.pre-commit-config.yaml
```diff
@@ -37,3 +37,18 @@ repos:
...
+        language: system
+        types: [markdown]
+        # Don't check localized files since their target might not be localized.
+        exclude: ".*localized.*"
```
chriselion: This might be controversial - if a localized file tries to link to something that hasn't been translated, what should we do? We could link to the english version, but that might be bad for the user. Currently we allow the dead links but we can change them if we really want.

PR #2545, opened 2019-09-12
One place to manage all of the demo files

xiaomaogy:


xiaomaogy:
@mmattar @ervteng

xiaomaogy:
This PR is to resolve the issue raise under https://github.com/Unity-Technologies/ml-agents/pull/2526#issuecomment-530612998. 

To summarize, this PR moves all of the .demo files under /demo to /UnitySDK/Assets/Demonstrations/ in order to have a consistent place for users to place and manage their .demo files.

xiaomaogy:
I totally agree with your point that the Expert demo files we provided won't be confused with the user recorded demo files when we put all them under Demonstrations folder, because 1) the files are based on scene name 2) user can choose their custom name. 

My concern here comes from a user workflow perspective. When a user try to record some demo file, they don't just try once, they will try a lot of times until they get a satisfying demo file. Imagine you have a Demonstration folder with 20+ demo files you just recorded, along with all of the Expert demo files we've provided, and now you want to clean this folder up without removing the Expert demo files. It would be pretty painful for the user having to select out their own demo files and delete them when they have all of these Expert demo files in the same folder. 

@ervteng @mmattar

ervteng:
@xiaomaogy @mmattar Hmm, the overcrowding in the directory is pretty much the reason for moving it into the `demos` directory and not the `Demonstrations` directory. On more thought I don't particularly like having the examples in the root `Demonstrations` folder, it's just too cluttered. My vote is for a subdirectory called `Examples` and renaming the files from `ExpertX.demo` to just `X.demo`. I don't think it's a stretch for the user to see the word "Example" in a path and know that you have to remove it - less so for `Expert`.  

Just my two cents, and apologize for changing my mind again :P

xiaomaogy:
> I am wondering if we should add an IL example to the basic guide to help onboard a new user quickly to IL and the use of demo files.

Yes that would be really helpful! However since we already have GAIL, should we add GAIL to the basic guide instead and suggest that as the default way of doing IL, since it achieves everything offline BC has, and more? @ervteng @mmattar

ervteng:
@xiaomaogy we can add a guide to use GAIL + PreTraining to speed up training (in the same way as Getting-Started-3D-Ball). Basically to replicate the plots on the Imitation Learning page

mmattar:
@xiaomaogy / @ervteng - one proposal is to include the .demo files for the sample environments in their respective folders. This way, each sample environment is self-contained within its folder. Thoughts?

ervteng:
@mmattar Would be kind of messy to put all the files everywhere. I'd rather have them in the same folder so a user knows exactly which demos are available

xiaomaogy:
@mmattar In the ml-agents-cloud, currently we don't download the while UnitySDK folder, just the demos folder to save the download time. If we put the demo files within their corresponding folder, it would be messy to download the demo files from each of the corresponding folder.

xiaomaogy:
@mmattar @ervteng According our previous discussion, I've moved the demo files under their corresponding Example folders and renamed they accordingly.

config/gail_config.yaml
```diff
@@ -32,7 +32,7 @@ PyramidsLearning:
     max_steps: 5.0e5
     num_epoch: 3
     pretraining:
-        demo_path: ./demos/ExpertPyramid.demo
```
ervteng: Dem paths :( :( 

But it is unavoidable

ml-agents/mlagents/trainers/tests/test_bcmodule.py
```diff
@@ -31,7 +31,7 @@ def ppo_dummy_config():
         use_recurrent: false
         memory_size: 8
         pretraining:
-          demo_path: ./demos/ExpertPyramid.demo
```
ervteng: I think we can change this to anything as the demo_path is simply overridden in the test

chriselion:

xiaomaogy:
Making a new PR based on this PR since a lot of things has changed.

PR #2549, opened 2019-09-13
Fix crash in construct_curr_info when next_info doesn't have any agents

### LINKED ISSUE #1687: Crash when use_recurrent=true when code thinks "0 agents" are in the scene
Hello ML Agent folk!

I'm encountering a very odd, difficult to reproduce bug that I think I've tracked down reasonably (and hopefully you can shed some light on things!)

in `ppo/trainer.py` in the `construct_curr_info` function, line 150 to 171 has the following:

code snippet

If `next_info.agents` is empty, line 171 still attempts to `vstack(memory)`, yielding an error/crash.  I've changed this section to the following

bug fix?

And my program no longer crashes.  As for my questions/concerns:

1. Is this a reasonable fix?  Will this cause some long term damage on my agent learning that I won't know about and perhaps a crash was correct?
2. Without the fix, my program appears to crash at RANDOM times.  That is, with ZERO changes to code, unity scene, anything, sometimes the code would crash after 3000 steps, sometimes never, sometimes after 150,000 steps.  My guess is that there's some race condition happening somewhere -- possibly that `next_info.agents` isn't getting the data in time on some ticks.
3. To answer some other expect questions -- I'm not deleting/adding any agents to my scene.  I have 1 agent that moves around an empty field and resets to its initial position on `AgentReset`.  If there's other things I should troubleshoot, do let me know!

ervteng:
Fixes #1687

### DIFF
--- a/ml-agents/mlagents/trainers/rl_trainer.py
+++ b/ml-agents/mlagents/trainers/rl_trainer.py
@@ -99,6 +99,7 @@ class RLTrainer(Trainer):
                 agent_brain_info.previous_text_actions[agent_index]
             )
             action_masks.append(agent_brain_info.action_masks[agent_index])
+        # Check if memories exists (i.e. next_info is not empty) before attempting vstack
         if self.policy.use_recurrent and memories:
             memories = np.vstack(memories)
         curr_info = BrainInfo(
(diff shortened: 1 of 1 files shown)


ervteng:
Added comment

PR #2556, opened 2019-09-13
Normalize observations when adding experiences

harperj:
This change moves normalization of vector observations into the trainer's
"add_experiences" interface.

Prior to this change, normalization occurred at inference time.  This
was somewhat confusing since usually executing a forward pass shouldn't
have side-effects which would change the training step.  Also, in a
asynchronous or distributed setting where we copy the neural network
weights from a trainer to a remote actor / inference worker we'd end up
with training issues because of the weights being different on the trainer
than the workers.

**Tested on:** CrawlerDynamic with PPO and SAC.  Saw no major discrepancies in the performance  when compared against `develop`.

### DIFF
--- a/ml-agents/mlagents/trainers/ppo/policy.py
+++ b/ml-agents/mlagents/trainers/ppo/policy.py
@@ -169,7 +169,6 @@ class PPOPolicy(TFPolicy):
         :param num_sequences: Number of sequences to process.
         :return: Results of update.
         """
-
         feed_dict = self.construct_feed_dict(self.model, mini_batch, num_sequences)
         stats_needed = self.stats_name_to_update_name
         update_stats = {}
@@ -183,7 +182,6 @@ class PPOPolicy(TFPolicy):
         update_vals = self._execute_model(feed_dict, self.update_dict)
         for stat_name, update_name in stats_needed.items():
             update_stats[stat_name] = update_vals[update_name]
-
         return update_stats
 
     def construct_feed_dict(self, model, mini_batch, num_sequences):
--- a/ml-agents/mlagents/trainers/ppo/policy.py
+++ b/ml-agents/mlagents/trainers/ppo/policy.py
@@ -104,13 +104,6 @@ class PPOPolicy(TFPolicy):
             self.inference_dict["pre_action"] = self.model.output_pre
         if self.use_recurrent:
             self.inference_dict["memory_out"] = self.model.memory_out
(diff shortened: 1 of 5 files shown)


ml-agents/mlagents/trainers/tf_policy.py
```diff
@@ -54,6 +54,7 @@ def __init__(self, seed, brain, trainer_parameters):
         self.seed = seed
         self.brain = brain
         self.use_recurrent = trainer_parameters["use_recurrent"]
+        self.normalize = trainer_parameters.get("normalize", False)
```
chriselion: Don't you want this to default to True?

ervteng: Default is False as normalize modifies the observation values

ml-agents/mlagents/trainers/ppo/policy.py
```diff
@@ -176,6 +169,7 @@ def update(self, mini_batch, num_sequences):
         :param num_sequences: Number of sequences to process.
         :return: Results of update.
         """
+
```
ervteng: Why the newlines?

ml-agents/mlagents/trainers/ppo/trainer.py
```diff
@@ -91,6 +91,8 @@ def process_experiences(
         :param new_info: Dictionary of all next brains and corresponding BrainInfo.
         """
         info = new_info[self.brain_name]
+        if self.is_training:
```
ervteng: Does this crash if vector_observations is []?

ml-agents/mlagents/trainers/ppo/policy.py
harperj: Should have been removed, will fix.

ml-agents/mlagents/trainers/ppo/trainer.py
harperj: It should not.  Firstly because it should be impossible to use vector observations and not return any. Second of all because the tensor is of variable length.

ml-agents/mlagents/trainers/tf_policy.py
harperj: I'm actually a little torn on what the default should be longer term since most users will likely need to normalize their vector obs -- but I think that for now it makes sense to stay with the more conservative choice.

PR #2567, opened 2019-09-16
Enable learning rate decay to be disabled

ervteng:
Normally, learning rate is decayed until `max_steps` is reached. This is unnecessary (or even undesirable) for SAC as the Q functions will naturally converge over time, and we don't see the same instabilities in training. 

Removing this decay in SAC improves performance for WallJump and Crawler environments and doesn't regress performance for any other environments

In the interest of keeping the option list slim, learning_rate_schedule is an optional param (docs included).


ml-agents/mlagents/trainers/trainer.py
```diff
@@ -8,7 +8,7 @@
 
 from mlagents.envs import UnityException, AllBrainInfo, ActionInfoOutputs, BrainInfo
 from mlagents.envs.timers import set_gauge
-from mlagents.trainers import TrainerMetrics
+from mlagents.trainers.trainer_metrics import TrainerMetrics
```
ervteng: This fixes some import errors

docs/Training-SAC.md
```diff
@@ -137,6 +137,18 @@ does not consistently increase.
 
 Typical Range: `1e-5` - `1e-3`
 
+### (Optional) Learning Rate Schedule
+
+`learning_rate_schedule` corresponds to how the learning rate is changed over time.
+For SAC, we typically hold learning rate constant so that the agent can continue to
```
awjuliani: Maybe instead of "we typically" we can say "we recommend."

ml-agents/mlagents/trainers/ppo/multi_gpu_policy.py
```diff
@@ -50,6 +50,11 @@ def create_model(
                             PPOModel(
                                 brain=brain,
                                 lr=float(trainer_params["learning_rate"]),
+                                lr_schedule=LearningRateSchedule(
+                                    trainer_params.get(
+                                        "learning_rate_schedule", "linear"
```
chriselion: This should LearningRateSchedule.LINEAR, right? (same for the other places that it's used)

ervteng: I'm using the same convention that the `vis_encode_type` uses - maybe we can change both?

chriselion: Oh my bad, it's creating a new enum instance there. All good!

PR #2584, opened 2019-09-18
Fix flake8 import warnings

harperj:
We have been ignoring unused imports and star imports via flake8. These are
both bad practice and grow over time without automated checking.  This
commit attempts to fix all existing import errors and add back the corresponding
flake8 checks.


ml-agents-envs/mlagents/envs/environment.py
```diff
@@ -68,7 +68,7 @@ def __init__(
         atexit.register(self._close)
         self.port = base_port + worker_id
         self._buffer_size = 12000
-        self._version_ = "API-8"
+        self._version_ = "API-9"
```
chriselion: Why is this changing?

chriselion: Oops, realized this was one commit not the whole PR.

PR #2585, opened 2019-09-18
Allow mypy to reject incomplete defs for mlagents-envs

harperj:
This wasn't working before because of several remaining partially defined
function definitions.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,8 +15,7 @@ repos:
         files: "ml-agents-envs/.*"
         # Exclude protobuf files and don't follow them when imported
         exclude: ".*_pb2.py"
-        # TODO get disallow-incomplete-defs working
-        args: [--ignore-missing-imports, --follow-imports=silent]
+        args: [--ignore-missing-imports, --disallow-incomplete-defs]
     -   id: mypy
         name: mypy-gym-unity
         files: "gym-unity/.*"
--- a/ml-agents-envs/mlagents/envs/base_unity_environment.py
+++ b/ml-agents-envs/mlagents/envs/base_unity_environment.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from typing import Dict
+from typing import Dict, Optional, Any
 
 from mlagents.envs.brain import AllBrainInfo, BrainParameters
 
@@ -7,13 +7,20 @@ from mlagents.envs.brain import AllBrainInfo, BrainParameters
 class BaseUnityEnvironment(ABC):
     @abstractmethod
     def step(
-        self, vector_action=None, memory=None, text_action=None, value=None
+        self,
+        vector_action: Dict,
+        memory: Optional[Dict] = None,
+        text_action: Optional[Dict] = None,
+        value: Optional[Dict] = None,
     ) -> AllBrainInfo:
         pass
 
     @abstractmethod
     def reset(
-        self, config=None, train_mode=True, custom_reset_parameters=None
+        self,
+        config: Optional[Dict] = None,
+        train_mode: bool = True,
+        custom_reset_parameters: Any = None,
     ) -> AllBrainInfo:
         pass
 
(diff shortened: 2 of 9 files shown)


.pre-commit-config.yaml
```diff
@@ -15,8 +15,7 @@ repos:
         files: "ml-agents-envs/.*"
         # Exclude protobuf files and don't follow them when imported
         exclude: ".*_pb2.py"
-        # TODO get disallow-incomplete-defs working
-        args: [--ignore-missing-imports, --follow-imports=silent]
+        args: [--ignore-missing-imports, --disallow-incomplete-defs]
```
chriselion: I think you still need `--follow-imports=silent` - when something imports protobuf code (even though it's ignored) that generates mypy errors (unless this option is on)

ml-agents-envs/mlagents/envs/base_unity_environment.py
```diff
@@ -1,19 +1,26 @@
...
 
 from mlagents.envs.brain import AllBrainInfo, BrainParameters
 
 
 class BaseUnityEnvironment(ABC):
     @abstractmethod
     def step(
-        self, vector_action=None, memory=None, text_action=None, value=None
+        self,
+        vector_action: Dict,
```
chriselion: Looks like this needs to be `Optional[Dict] = None`

ml-agents-envs/mlagents/envs/brain.py
```diff
@@ -2,12 +2,78 @@
...
+            str(self.number_visual_observations),
+            str(self.vector_observation_space_size),
+            str(self.num_stacked_vector_observations),
+            self.vector_action_space_type,
+            str(self.vector_action_space_size),
+            ", ".join(self.vector_action_descriptions),
+        )
+
+    @staticmethod
+    def from_proto(brain_param_proto):
```
chriselion: nit: add types while you're here?

ml-agents-envs/mlagents/envs/environment.py
```diff
@@ -355,11 +358,11 @@ def reset(
     @timed
     def step(
         self,
-        vector_action=None,
-        memory=None,
-        text_action=None,
-        value=None,
-        custom_action=None,
+        vector_action: Dict,
```
chriselion: Any chance you can be more specific on the values of these Dicts? I assume the types are strings?

ml-agents-envs/mlagents/envs/base_unity_environment.py
harperj: I actually think that, unlike the others, an action is mandatory -- that's why I switched it to `Dict`

ml-agents-envs/mlagents/envs/environment.py
harperj: Done

ml-agents-envs/mlagents/envs/brain.py
harperj: 👍

ml-agents-envs/mlagents/envs/base_unity_environment.py
chriselion: The unit tests disagree :)

harperj: I thought it was because the test was wrong, but I now remember a recent change means this should actually be the way it was before :-)

.pre-commit-config.yaml
harperj: Looks like this isn't needed, checks are passing 🤷‍♂

ml-agents-envs/mlagents/envs/environment.py
```diff
@@ -495,8 +498,6 @@ def step(
                 else:
                     if text_action[brain_name] is None:
                         text_action[brain_name] = [""] * n_agent
-                    if isinstance(text_action[brain_name], str):
```
harperj: Note this is an actual functional change.  Requires that if we are given a `text_action` that it will be a `Dict[str, List[str]]`.  Otherwise the types will be inconsistent.  I don't believe we use this feature but I thought it was worth noting.

.pre-commit-config.yaml
chriselion: Weird, sounds good though.

PR #2596, opened 2019-09-19
image decompress timer

chriselion:
We want to gather timings on how much we currently spend decompressing during visual learning.

### DIFF
--- a/ml-agents-envs/mlagents/envs/brain.py
+++ b/ml-agents-envs/mlagents/envs/brain.py
@@ -6,7 +6,7 @@ from mlagents.envs.communicator_objects.agent_info_proto_pb2 import AgentInfoPro
 from mlagents.envs.communicator_objects.brain_parameters_proto_pb2 import (
     BrainParametersProto,
 )
-from mlagents.envs.timers import hierarchical_timer
+from mlagents.envs.timers import hierarchical_timer, timed
 from typing import Dict, List, Optional
 from PIL import Image
 
@@ -154,6 +154,7 @@ class BrainInfo:
         return np.append(m1, m2, axis=0)
 
     @staticmethod
+    @timed
     def process_pixels(image_bytes: bytes, gray_scale: bool) -> np.ndarray:
         """
         Converts byte array observation image into numpy array, re-sizes it,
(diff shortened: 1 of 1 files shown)


ml-agents-envs/mlagents/envs/brain.py
```diff
@@ -161,8 +162,9 @@ def process_pixels(image_bytes: bytes, gray_scale: bool) -> np.ndarray:
         :param image_bytes: input byte array corresponding to image
         :return: processed numpy array of observation from environment
         """
-        image_bytearray = bytearray(image_bytes)
-        image = Image.open(io.BytesIO(image_bytearray))
+        with hierarchical_timer("image_decompress"):
+            image_bytearray = bytearray(image_bytes)
+            image = Image.open(io.BytesIO(image_bytearray))
```
chriselion: Actually, I might need to add the next line under the timer too. According to the [docs](https://pillow.readthedocs.io/en/4.2.x/reference/Image.html#PIL.Image.open), `Image.open()` 
> is a lazy operation; this function identifies the file, but the file remains open and the actual image data is not read from the file until you try to process the data (or call the load() method).

surfnerd: 👍

chriselion:
Just a note on the lazy loading. Here's timings from the same run of GridWorld with and without explicitly calling Image.load():

`without load()`
``` json
{
  "name": "BrainInfo.process_pixels",
  "total": 3.584165224000029,
  "count": 10002,
  "is_parallel": true,
  "self": 2.7557900760001024,
  "children": [
    {
      "name": "image_decompress",
      "total": 0.8283751479999264,
      "count": 10002,
      "is_parallel": true,
      "self": 0.8283751479999264
    }
  ]
}
```

`with load()`
``` json
{
  "name": "BrainInfo.process_pixels",
  "total": 3.646127201000195,
  "count": 10001,
  "is_parallel": true,
  "self": 0.9565701650002723,
  "children": [
    {
      "name": "image_decompress",
      "total": 2.689557035999923,
      "count": 10001,
      "is_parallel": true,
      "self": 2.689557035999923
    }
  ]
}
```
So total time is about the same (within error bars) but forcing the load shifts the time into the `image_decompress` block (where it belongs).

chriselion:
BTW, total train time (wallclock) was about 36.78seconds for those runs.

PR #2598, opened 2019-09-19
Fix crash with VAIL + GAIL

ervteng:
`tf.control_dependencies` requires a list, we were giving it a single Tensor. This was causing a crash if VAIL was enabled. This fixes that issue.


chriselion:
Please add a unit test that would have exposed this (or at least make a follow-up item to do it later)

ervteng:
@chriselion VAIL is a superset of the functionality in GAIL, so I changed the test config to use VAIL.

chriselion:
Cool, thanks for updating the test.

PR #2601, opened 2019-09-20
Develop yaml json loading errors

chriselion:
Since a common source of user problems and github issues is yaml/json configuration problems, I wanted to make the loading errors more explicit and point users to where they can validate their files.

I also split the loading into two parts to make testing and error checking a little clearer:
1) One part that loads the file and passes the file object to another function
2) One part that does the actual loading from the file object
It's easier to test the second function by using an `io.StringIO` (otherwise you have to mock open() and it's ugly).

I also split the CurriculumError into CurriculumLoadingError (for errors loading the file) and CurriculumConfigError (for semantic errors using the loaded file).

I noticed both json and yaml loading had some exception handling around UnicodeDecodeError but I'm not sure how to reproduce this in a test (but I assume it was there for a good reason).


ml-agents/mlagents/trainers/trainer_util.py
```diff
@@ -108,3 +109,31 @@ def initialize_trainers(
...
+    try:
+        with open(trainer_config_path) as data_file:
+            return _load_config(data_file)
+    except IOError:
+        raise UnityEnvironmentException(
+            f"Config file could not be found at {trainer_config_path}."
+        )
+    except UnicodeDecodeError:
+        raise UnityEnvironmentException(
+            f"There was an error decoding Config file from {trainer_config_path}. "
```
chriselion: I made these exceptions a bit more generic since they're used for both Trainer and Sampler config loading.

ml-agents/mlagents/trainers/curriculum.py
```diff
@@ -117,3 +111,27 @@ def get_config(self, lesson=None):
         for key in parameters:
             config[key] = parameters[key][lesson]
         return config
+
+    @staticmethod
+    def load_curriculum_file(location):
```
chriselion: I kept these in the same file, but I can move them to trainer_util.py if that's cleaner. It's currently the only place we load json.

ml-agents/mlagents/trainers/curriculum.py
```diff
@@ -117,3 +111,27 @@ def get_config(self, lesson=None):
...
+        try:
+            with open(location) as data_file:
+                return Curriculum._load_curriculum(data_file)
+        except IOError:
+            raise CurriculumLoadingError(
+                "The file {0} could not be found.".format(location)
+            )
+        except UnicodeDecodeError:
```
chriselion: This should probably be handled in _load_curriculum instead, but I couldn't reproduce it to confirm :/

ml-agents/mlagents/trainers/trainer_util.py
harperj: In that case I'd recommend renaming the argument to just `config_path`

PR #2610, opened 2019-09-24
Tick version of API and pypi packages to 10

ervteng:


UnitySDK/Assets/ML-Agents/Scripts/Academy.cs
```diff
@@ -97,7 +97,7 @@ public abstract class Academy : MonoBehaviour
         [SerializeField]
         public BroadcastHub broadcastHub = new BroadcastHub();
 
-        private const string k_KApiVersion = "API-9";
+        private const string k_KApiVersion = "API-10";
```
chriselion: Not sure if we want to increment this since I don't think anything has changed.

ervteng: I had the same discussion with Vince during 0.9.0. Our current process is to tick at every version - not sure if we should continue to do so. It's a pain for our developers TBH, esp. if we don't change anything on the C# API side.

PR #2615, opened 2019-09-24
Delete Soccer TFModels

ervteng:
Soccer didn't train properly, and our trained models didn't work well. This PR removes them, leaving Soccer as an example but not providing any trained models.


ervteng:
@Hunter-Unity I discussed with @mmattar and we decided that we should leave Soccer in as an example env, but not provide any trained models until we have a good behavior.

PR #2620, opened 2019-09-25
Better environment shutdown

chriselion:
While working on the timer code in https://github.com/Unity-Technologies/ml-agents/pull/2198 I noticed we weren't executing code at shutdown (namely so that I could write out the timer data).

This fixes two issues:
1) Don't kill the subprocess right away, give it a chance to shutdown. This should happen in the Academy code when the QUIT command is received*:
https://github.com/Unity-Technologies/ml-agents/blob/c916801475950ef92016b51c3fe5806ac60afc90/UnitySDK/Assets/ML-Agents/Scripts/Academy.cs#L536-L544
2) Launch the executable process with `start_new_session=True`. Without this, hitting Ctrl-C would propagate the signal (SIGINT) to the subprocess, so it dies before getting a chance to do anything.

Related to 2), also nicely format the return code (would have saved me some time if I know negative return corresponded to the signal ID).

I'm not sure if we should add configuration parameters for
* shutdown timeout (I'm currently using the communicator's `timeout_wait` which defaults to 30s, but that seems too long)
* whether or not to set start_new_session - I think this should default to True but might need an option to override.

*(NOTE) We don't actually send QUIT here, but when the communicator gets the `400` status that python sends
https://github.com/Unity-Technologies/ml-agents/blob/c916801475950ef92016b51c3fe5806ac60afc90/UnitySDK/Assets/ML-Agents/Scripts/RpcCommunicator.cs#L107-L116
it returns null, so the batcher treats this as a QUIT
https://github.com/Unity-Technologies/ml-agents/blob/c916801475950ef92016b51c3fe5806ac60afc90/UnitySDK/Assets/ML-Agents/Scripts/Batcher.cs#L241-L245

### DIFF
--- a/ml-agents-envs/mlagents/envs/environment.py
+++ b/ml-agents-envs/mlagents/envs/environment.py
@@ -33,6 +33,7 @@ from mlagents.envs.communicator_objects.custom_action_pb2 import CustomAction
 
 from .rpc_communicator import RpcCommunicator
 from sys import platform
+import signal
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger("mlagents.envs")
@@ -79,6 +80,7 @@ class UnityEnvironment(BaseUnityEnvironment):
         self.proc1 = (
             None
         )  # The process that is started. If None, no process was started
+        self.timeout_wait: int = timeout_wait
         self.communicator = self.get_communicator(worker_id, base_port, timeout_wait)
         self.worker_id = worker_id
 
@@ -249,7 +251,14 @@ class UnityEnvironment(BaseUnityEnvironment):
                 subprocess_args += ["--port", str(self.port)]
                 subprocess_args += args
                 try:
-                    self.proc1 = subprocess.Popen(subprocess_args)
+                    self.proc1 = subprocess.Popen(
+                        subprocess_args,
+                        # start_new_session=True means that signals to the parent python process
+                        # (e.g. SIGINT from keyboard interrupt) will not be sent to the new process.
+                        # This is generally good since we want the environment to have a chance to shutdown,
+                        # but may be undesirable in come cases.
+                        start_new_session=True,
+                    )
                 except PermissionError as perm:
                     # This is likely due to missing read or execute permissions on file.
                     raise UnityEnvironmentException(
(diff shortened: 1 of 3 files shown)


ml-agents-envs/mlagents/envs/subprocess_env_manager.py
```diff
@@ -119,18 +119,18 @@ def _send_response(cmd_name, payload):
             elif cmd.name == "close":
                 break
     except (KeyboardInterrupt, UnityCommunicationException):
-        print("UnityEnvironment worker: environment stopping.")
+        logger.info(f"UnityEnvironment worker {worker_id}: environment stopping.")
```
chriselion: Cleaned this up to use `logger` and made it and the debug's below more consistent.

ml-agents-envs/mlagents/envs/environment.py
```diff
@@ -583,7 +592,16 @@ def _close(self):
         self._loaded = False
         self.communicator.close()
         if self.proc1 is not None:
-            self.proc1.kill()
+            # Wait a bit for the process to shutdown, but kill it if it takes too long
+            try:
+                self.proc1.wait(timeout=self.timeout_wait)
```
surfnerd: I think this addresses the main concern I had, which was a user would have to manually close 15 unity processes.  But it looks like you are waiting here to see if they play nice, then drop the hammer if not.

chriselion: True, it'll still eventually get killed. But we're relying on the Academy interpreting the commands and closing the application; a game developer might expect the signals to get propagated.

Maybe we should make Academy listen for signals instead?

chriselion:
@harperj Haven't tried on Windows. The start_new_session parameter only applies to POSIX systems. From https://docs.python.org/3/library/subprocess.html#popen-constructor
> If start_new_session is true the setsid() system call will be made in the child process prior to the execution of the subprocess. (POSIX only)

So it at least shouldn't make anything worse on Windows (I'm not sure if `KeyboardInterrupt` will propagate there anyway)

I think the "wait, else kill" approach is generally better.

ml-agents-envs/mlagents/envs/environment.py
surfnerd: ehhh, I don't like the sound of that.

harperj:
I'd recommend at least giving it a test on a Windows machine before merging.  Otherwise 🚢 .

chriselion:
Tested on windows - CTRL-C still gets sent to the executable, but it's no worse (and not sure we can do better)

ml-agents-envs/mlagents/envs/environment.py
chriselion: OK, punting on this for now. If it comes up, we'll add a flag to toggle `start_new_session`

PR #2621, opened 2019-09-25
Move gRPC code to its own special place

surfnerd:


UnitySDK/Assets/ML-Agents/Editor/DemonstrationImporter.cs
```diff
@@ -29,7 +29,7 @@ public override void OnImportAsset(AssetImportContext ctx)
                 Stream reader = File.OpenRead(ctx.assetPath);
 
                 var metaDataProto = DemonstrationMetaProto.Parser.ParseDelimitedFrom(reader);
-                var metaData = new DemonstrationMetaData(metaDataProto);
+                var metaData = metaDataProto.FromProto();
```
chriselion: The naming feels weird on this. Guess I'm just used to `Foo.FromBar()` be a static method in `Foo` not an instance method on `Bar`

UnitySDK/Assets/ML-Agents/Scripts/Grpc/GrpcExtensions.cs
```diff
@@ -0,0 +1,115 @@
...
+                NumberEpisodes = dm.numberEpisodes,
+                DemonstrationName = dm.demonstrationName
+            };
+            return demoProto;
+        }
+        
+        /// <summary>
+        /// Initialize metadata values based on proto object.
+        /// </summary>
+        public static DemonstrationMetaData FromProto(this DemonstrationMetaProto demoProto)
```
chriselion: Maybe `ToDemonstrationMetaData`? `ToManaged`? not sure what else to call it though.

UnitySDK/Assets/ML-Agents/Editor/DemonstrationImporter.cs
surfnerd: It's technically a static method, just looks like an instance method.

surfnerd: But I agree with your sentiment.  I just wanted to remove as much Grpc code away from code that doesn't need to know about it

UnitySDK/Assets/ML-Agents/Scripts/Grpc/GrpcExtensions.cs
surfnerd: sounds good.

PR #2637, opened 2019-09-28
small mypy cleanup

chriselion:
I'm trying to get `--namespace-packages` in mypy working for our project (it currently hits an internal assert). In the meantime, here are a few things that flagged.

### DIFF
--- a/ml-agents/mlagents/trainers/sac/policy.py
+++ b/ml-agents/mlagents/trainers/sac/policy.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Dict, Any
+from typing import Dict, Any, Optional
 import numpy as np
 import tensorflow as tf
 
@@ -58,6 +58,7 @@ class SACPolicy(TFPolicy):
 
         with self.graph.as_default():
             # Create pretrainer if needed
+            self.bc_module: Optional[BCModule] = None
             if "pretraining" in trainer_params:
                 BCModule.check_config(trainer_params["pretraining"])
                 self.bc_module = BCModule(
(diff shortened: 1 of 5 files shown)


ml-agents/mlagents/trainers/sac/trainer.py
```diff
@@ -121,7 +120,7 @@ def save_replay_buffer(self) -> None:
         with open(filename, "wb") as file_object:
             self.training_buffer.update_buffer.save_to_file(file_object)
 
-    def load_replay_buffer(self) -> Buffer:
+    def load_replay_buffer(self) -> None:
```
chriselion: Fixed the return type here, which also lets you not import `Buffer`

ml-agents/mlagents/trainers/sac/policy.py
```diff
@@ -58,6 +58,7 @@ def __init__(
 
         with self.graph.as_default():
             # Create pretrainer if needed
+            self.bc_module: Optional[BCModule] = None
```
ervteng: Should we use the same logic for PPO?

chriselion: It wasn't complaining about PPO (probably because no type hints there) but makes sense to do it there too.

chriselion: Done now.

ervteng: :+1: Thanks for adding the type hints to PPO too.

PR #2646, opened 2019-09-30
Rename protobuf objects to be suffixed with 'Proto' in python and C#.

surfnerd:
In order to have consistency in our protobuf object naming, I've renamed all of our message types to end with a suffix: 'proto'.  I've also renamed all of our protofiles to remove a redundant '_proto' from their file name.  Having the protobuf objects end with 'Proto' will disambiguate the mlagents data structures from the gRPC data structures.  

The reason behind this is that I am working on a PR that will push all references to protobufs down to the implementation of the RpcCommunicator on the C# side.  This will help to make the communicator module more modular, enabling implementation of different Communictors in the future.


UnitySDK/ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +1 @@
-m_EditorVersion: 2017.4.29f1
+m_EditorVersion: 2017.4.10f1
```
chriselion: Expected?

surfnerd: nope

surfnerd: fixed 758e2c66

chriselion:
Does this break recorded demos, or are the old files compatible? Wasn't sure if it used name matching or something else.

surfnerd:
> Does this break recorded demos, or are the old files compatible? Wasn't sure if it used name matching or something else.

Good question.  I am not sure.  Let me dig into this.

surfnerd:
@chriselion as we mentioned in our chat offline there are tests in place for loading demonstration files here:
https://github.com/Unity-Technologies/ml-agents/blob/758e2c66c8e4b0da4f7fb23bc0d6cbdab97e7791/ml-agents/mlagents/trainers/tests/test_demo_loader.py#L6

I think this should be sufficient enough for us to know if we've broken backwards compatibility.

PR #2647, opened 2019-09-30
Cleanup visual obs setup

chriselion:
"Don't repeat yourself" - the visual encoder setup was pretty repetitive.

If you think this is a bad abstraction (i.e. the input parameters for the various encoders might differ in the future), feel free to reject this.

### DIFF
--- a/ml-agents/mlagents/trainers/models.py
+++ b/ml-agents/mlagents/trainers/models.py
@@ -470,7 +470,7 @@ class LearningModel(object):
         num_layers: int,
         vis_encode_type: EncoderType = EncoderType.SIMPLE,
         stream_scopes: List[str] = None,
-    ) -> tf.Tensor:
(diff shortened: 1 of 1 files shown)


ml-agents/mlagents/trainers/models.py
```diff
@@ -491,54 +491,40 @@ def create_observation_streams(
             self.visual_in.append(visual_input)
         vector_observation_input = self.create_vector_input()
 
+        # Pick the encoder function based on the EncoderType
+        create_encoder_func = LearningModel.create_visual_observation_encoder
```
chriselion: Note that neither the old or new versions raise an error on an invalid enum, they just fall through to `create_visual_observation_encoder`. Could do this as a dict from EncoderType -> Callable but didn't seem like much advantage for this small number of choices.

PR #2627, opened 2019-09-26
Curricula json fix

Vaspra:
I was having an issue with `meta_curriculum` attempting to generate curricula instances from Unity-made `curricula/MyThingLearning/MyThingLearning.json.meta` files. Obviously the meta files are _not_ intended to be used as .json sources.

The change simply excludes filenames ending in `.meta`.

Issue #2113, opened 2019-06-09
Trying to train with curricula causes a JsonDecodeError


DanAmador:
I have a Curriculum file which supposedly helps to teach my agent how to tackle some other objects in the scene. 

It uses a brain called SumoTackleLearn and I tried loading it by using the command 

> MLAgents/Config/trainer_config.yaml --debug --train --curriculum=MLAgents/Config/curricula/Tackle --run-id=tackle-curriculum

This at first gave me an error stating that a reset parameter was expecting more values than it actually had, this in turn showed me that the file was being correctly loaded, but I had just misread the documentation.

After fixing that mistake I tried training with said curriculum again but I've bumped into this:

![image](https://user-images.githubusercontent.com/7833567/59161955-76673880-8ae9-11e9-814e-ea0b14cdccad.png)

The tackle folder contains a file called "SumoTackleLearn.json" with the following content: 

```
{
  "measure": "reward",
  "thresholds": [5, 5, 6, 7],
  "min_lesson_length": 100,
  "signal_smoothing": false,
  "parameters": {
    "EnemySpawnVariance": [0, 5.0, 10.0, 15.0, 20.0]
  }
}
```

And my Academy object, which trains the SumoTackleLearn brain has the EnemySpawnVariance as a Reset Parameter. 

I've tried following what the documentation said or what the few tutorials online showed, but I can't seem to get it running. 

I know for a fact that the file gets correctly loaded because of my previous issue which leads me to believe that this might be caused by some sort of bug. 
If not what's wrong with my configuration?

harperj:
Hi @DanAmador -- I'm able to load the JSON file as pasted here, so I'm not sure why you're running into an issue.  Perhaps it could be related to Windows.

Maybe you could try just parsing the file via the Python REPL:

```
>>> import json
>>> with open("path/to/your/curriculum.json") as data_file:
...     json.load(data_file)
{'measure': 'reward', 'thresholds': [5, 5, 6, 7], 'min_lesson_length': 100, 'signal_smoothing': False, 'parameters': {'EnemySpawnVariance': [0, 5.0, 10.0, 15.0, 20.0]}}
```

If this works then we'll need to figure out why the Curriculum class isn't loading the correct file or how it differs.

DanAmador:
Everything works as expected while loading it with the repl 

![image](https://user-images.githubusercontent.com/7833567/59229668-531cb600-8bdb-11e9-88de-bb448501cc50.png)

[...]

PR #2651, opened 2019-10-01
Modified meta_curriculum JSON reading loop to only deal with .json

Vaspra:
See #2627
This is a bit cleaner!

Causes a `continue` in curriculum generation when the source file is not an expected JSON file.

### DIFF
--- a/ml-agents/mlagents/trainers/meta_curriculum.py
+++ b/ml-agents/mlagents/trainers/meta_curriculum.py
@@ -30,6 +30,11 @@ class MetaCurriculum(object):
 
         try:
             for curriculum_filename in os.listdir(curriculum_folder):
+                
+                # This process requires JSON files
+                if not curriculum_filename.lower().endswith('.json'):
+                    continue
+
                 brain_name = curriculum_filename.split(".")[0]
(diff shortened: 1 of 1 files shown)


DanAmador:
Is this the reason that caused [this](https://github.com/Unity-Technologies/ml-agents/issues/2113) ?

Vaspra:
@DanAmador Yes it is, I had not seen this issue. I was wracking my head with it last week but realised it was simply trying to json.load the Unity .meta files.

With this modification mine works fine where it previously did not.

ml-agents/mlagents/trainers/meta_curriculum.py
```diff
@@ -30,6 +30,11 @@ def __init__(self, curriculum_folder, default_reset_parameters):
 
         try:
             for curriculum_filename in os.listdir(curriculum_folder):
+
```
chriselion: I think black (the style checker) is complaining about whitespace on this line.

chriselion:
As a side note, https://github.com/Unity-Technologies/ml-agents/pull/2601 hopefully made these issues easier to debug.

PR #2656, opened 2019-10-02
Use a namedtuple for Camera Resolution, not a dict

chriselion:
The string keys were bothering me, and it now encapsulates some of the greyscale/channels logic.

### DIFF
--- a/gym-unity/gym_unity/envs/__init__.py
+++ b/gym-unity/gym_unity/envs/__init__.py
@@ -132,18 +132,14 @@ class UnityEnv(gym.Env):
         high = np.array([np.inf] * brain.vector_observation_space_size)
         self.action_meanings = brain.vector_action_descriptions
         if self.use_visual:
-            if brain.camera_resolutions[0]["blackAndWhite"]:
-                depth = 1
-            else:
-                depth = 3
             self._observation_space = spaces.Box(
                 0,
                 1,
                 dtype=np.float32,
                 shape=(
-                    brain.camera_resolutions[0]["height"],
-                    brain.camera_resolutions[0]["width"],
-                    depth,
+                    brain.camera_resolutions.height,
+                    brain.camera_resolutions.width,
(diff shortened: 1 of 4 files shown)


ml-agents-envs/mlagents/envs/brain.py
```diff
@@ -5,19 +5,33 @@
...
 
 
+class CameraResolution(NamedTuple):
+    height: int
+    width: int
+    gray_scale: bool
+
+    @property
+    def num_channels(self) -> int:
+        return 1 if self.gray_scale else 3
```
ervteng: I almost want to say CameraResolution shouldn't be aware of whether it's grey scale or not, just how many channels it has. Seems more generalizable that way. We could pass it in 1 or 3 on creation from the proto instead - what do you think?

chriselion: Agreed, but that would be a much bigger change (would require communicator proto changes), and will (hopefully) come for free with the sensor refactor.

ervteng: OK, if it's coming anyways in the sensor refactor, no sense in doing it twice.

chriselion: Yeah, should be coming then, and also gives me some more confidence while doing the refactoring.

gym-unity/gym_unity/envs/__init__.py
```diff
@@ -132,18 +132,14 @@ def __init__(
...
-                depth = 3
             self._observation_space = spaces.Box(
                 0,
                 1,
                 dtype=np.float32,
                 shape=(
-                    brain.camera_resolutions[0]["height"],
-                    brain.camera_resolutions[0]["width"],
-                    depth,
+                    brain.camera_resolutions.height,
```
chriselion: Note: this is broken - should be brain.camera_resolutions[0].height, etc.

PR #2657, opened 2019-10-02
add config for stalebot

chriselion:
(Note: this needs to be on the default branch, not develop)

The initial list of "exempt" tags is pretty generous; we can un-whitelist as needed later.

After this merges, I'll set up the bot from here https://github.com/marketplace/stale

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -16,7 +16,7 @@ issues:
     Thank you for your contributions.
   closeComment: >
     This issue has been automatically closed because it has not had activity in the
-    last 28 days. If this issue is still valid, please ping a maintainer and ask them to relabel it.
+    last 28 days. If this issue is still valid, please ping a maintainer.
     Thank you for your contributions.
   exemptLabels:
     - bug
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -21,6 +21,4 @@ issues:
   exemptLabels:
     - bug
     - request
-    - request/roadmap
-    - request/contributions-welcome
     - help-wanted
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -4,7 +4,6 @@
 # Label to use when marking as stale
 staleLabel: stale
 
-
 # Issue specific configuration
 issues:
   # TODO: Raise the limitPerRun once we are satisfied with the bot's performance
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,27 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# General configuration
+# Label to use when marking as stale
+staleLabel: stale
+
+
+# Issue specific configuration
+issues:
+  # TODO: Raise the limitPerRun once we are satisfied with the bot's performance
+  limitPerRun: 1
+  daysUntilStale: 14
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,27 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# General configuration
+# Label to use when marking as stale
+staleLabel: stale
+
+
```
mmattar: Remove empty line

.github/stale.yml
```diff
@@ -0,0 +1,27 @@
...
+    Thank you for your contributions.
+  closeComment: >
+    This issue has been automatically closed because it has not had activity in the
+    last 28 days. If this issue is still valid, please ping a maintainer and ask them to relabel it.
+    Thank you for your contributions.
+  exemptLabels:
+    - bug
+    - request
+    - request/roadmap
+    - request/contributions-welcome
```
mmattar: Since the plan is to deprecate these labels, we should either 1. Remove them from here and enable stalebot post deprecating the labels, 2. Make a note to remove these two lines at a later date.

.github/stale.yml
```diff
@@ -0,0 +1,27 @@
...
+  limitPerRun: 1
+  daysUntilStale: 14
+  daysUntilClose: 14
+  markComment: >
+    This issue has been automatically marked as stale because it has not had activity in the
+    last 14 days. It will be closed in the next 14 days if no further activity occurs.
+    Thank you for your contributions.
+  closeComment: >
+    This issue has been automatically closed because it has not had activity in the
+    last 28 days. If this issue is still valid, please ping a maintainer and ask them to relabel it.
```
mmattar: Does re-opening a closed issue reset the timer on the stalebot? What scenario do you have in mind for the proposal "please ping a maintainer and ask them to relabel it."

.github/stale.yml
chriselion: Done

.github/stale.yml
chriselion: I think it's worthwhile to give a user recourse if they think the issue has been closed in error. Especially if they provided information and are waiting for a response from us.

I adapted some of the wording here from https://github.com/envoyproxy/envoy/blob/master/.github/stale.yml

Via Google search, some other configs have similar wordings:
https://github.com/rails/rails/blob/master/.github/stale.yml
```
  This issue has been automatically marked as stale because it has not been commented on for at least three months.

  The resources of the Rails team are limited, and so we are asking for your help.

  If you can still reproduce this error on the `6-0-stable` branch or on `master`,
  please reply with all of the information you have about it in order to keep the issue open.

  Thank you for all your contributions.
```
https://www.codefactor.io/repository/github/facebook/create-react-app/source/master/.github/stale.yml
```
This issue has been automatically closed because it has not had any recent
activity. If you have a question or comment, please open a new issue.
```
We can also apparently use `closeComment: false` and it won't say anything, but I'd be annoyed by that as a user.

.github/stale.yml
chriselion: Will update all those issues and remove the labels.

chriselion: All issues updated, removed the labels and updated this section.

PR #2661, opened 2019-10-02
Fix bug when batch size is a non-multiple of sequence length

ervteng:
- In PPO trainer, always round batch size down to the nearest multiple of sequence length, with at least one sequence always (same as 0.9 and SAC behavior). 
- Add test to detect this error

### DIFF
--- a/ml-agents/mlagents/trainers/ppo/trainer.py
+++ b/ml-agents/mlagents/trainers/ppo/trainer.py
@@ -230,7 +230,8 @@ class PPOTrainer(RLTrainer):
         )
         self.cumulative_returns_since_policy_update = []
 
-        # Make sure batch_size is a multiple of sequence length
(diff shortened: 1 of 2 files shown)


ml-agents/mlagents/trainers/ppo/trainer.py
```diff
@@ -229,7 +229,15 @@ def update_policy(self):
             mean_return=float(np.mean(self.cumulative_returns_since_policy_update)),
         )
         self.cumulative_returns_since_policy_update = []
-        batch_size = self.trainer_parameters["batch_size"]
+
+        # Make sure batch_size is a multiple of sequence length
```
chriselion: Nit: can you expand the comment to explain why we need to make sure about this? E.g. it's a tensorflow requirement?

ervteng: Added

PR #2662, opened 2019-10-02
Fix issue which prevented training when code stripping was enabled

surfnerd:
Fixes MLA-62.

Add a link.xml file to prevent the Grpc.Core.dll from being code stripped.


UnitySDK/Assets/ML-Agents/Plugins/ProtoBuffer/link.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<linker>
+    <assembly fullname="Grpc.Core" preserve="all"/>
+</linker>
```
chriselion: EOL? if CI doesn't complain, no big deal.

surfnerd: fixed

UnitySDK/Assets/ML-Agents/Plugins/ProtoBuffer/link.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<linker>
+    <assembly fullname="Grpc.Core" preserve="all"/>
```
chriselion: Can you add a comment here? Something like
"Prevents Managed code stripping from removing code necessary for training. You may wish to disable this preservation (that is, re-enable code stripping) to reduce code size if you no longer need training. See also https://docs.unity3d.com/Manual/ManagedCodeStripping.html"

Just imagining someone trying to ship their game and things getting bloated by code they don't need.

surfnerd: Done

Issue #1850, opened 2019-03-20
TF 2.0 support?


SetoKaiba:
TensorFlow 2.0 is released. It would be great to support it. There're many improvements.

**EDITED:**
https://github.com/SetoKaiba/ml-agents/tree/tf2
I make a tf2 branch for it.
The training is tested working with 3DBall.
And the generated nn file is working with Barracuda as well.

https://github.com/SetoKaiba/ml-agents/tree/tf2-develop
And also, the develop branch splits ml-agents into two packages.
I made a tf2-develop branch for it.
The training is tested working with 3DBall.
And the generated nn file is working with Barracuda as well.

awjuliani:
Hi @SetoKaiba 

TensorFlow 2.0 is still in alpha, and therefore not supported at this time. That being said, it looks like the error you are getting is related to the buffer, rather than tensorflow directly. It may be that this new version is outputting things in a different shape than in 1.0. 

Could you also share your numpy and python versions?

SetoKaiba:
Hi @awjuliani 

Yes. I know that. So I just tried to migrate to tf2 myself. But I encounter this problem.
I tried to print out the buffer and get this output. It seems that the dimension of advantages is incorrect. But the problem doesn't occur in the previous 9000 steps. It's strange.
The output is the last two iteration including the problem case.
Previous iteration is correct (45,), But it's incorrect int that iteration (1,12018)

> key:vector_obs,value:(45, 8)
> key:next_vector_in,value:(45, 8)
> key:actions_pre,value:(45, 2)
> key:random_normal_epsilon,value:(45, 2)
> key:actions,value:(45, 2)
> key:prev_action,value:(45, 2)
> key:masks,value:(45,)
> key:rewards,value:(45,)
> key:action_probs,value:(45, 2)
> key:value_estimates,value:(45,)
> key:advantages,value:(45,)
> key:discounted_returns,value:(45,)
> key:actions,value:(12018, 2)
> key:vector_obs,value:(12018, 8)
> key:next_vector_in,value:(12018, 8)
> key:actions_pre,value:(12018, 2)
> key:random_normal_epsilon,value:(12018, 2)
> key:prev_action,value:(12018, 2)
> key:masks,value:(12018,)
> key:rewards,value:(12018,)
> key:action_probs,value:(12018, 2)
> key:value_estimates,value:(12018,)
> key:advantages,value:(1, 12018)
> previous:12018,current:1
> Traceback (most recent call last):
[...]

PR #2665, opened 2019-10-03
Allow usage with tensorflow 2.0.0 (via tf.compat.v1)

chriselion:
Still hacky but most unit tests are passing locally. Haven't tried any "real" training yet.

HUGE thank to @SetoKaiba for figuring out most of the issues in the branch mentioned in https://github.com/Unity-Technologies/ml-agents/issues/1850.

Main issues:
- [x] Better version checking 
- [x] cleanup some of the imports (I don't like tf_flatten, not sure mlagents.trainers is the right spot, etc).
- [x] Figure out what's up with tf2bc
- [x] Ban raw "import tensorflow" except in one location
- [x] Set up CI to work on 1.7, 1.14, and 2.0
- [x] Measure inference quality - quick test with 3dball didn't look good


ml-agents/mlagents/trainers/bc/models.py
```diff
@@ -44,9 +44,7 @@ def __init__(
                         size,
                         activation=None,
                         use_bias=False,
-                        kernel_initializer=c_layers.variance_scaling_initializer(
-                            factor=0.01
-                        ),
+                        kernel_initializer=tf_variance_scaling(0.01),
```
chriselion: As mentioned in the PR, I don't love the name. Would like to find a better way to handle this.

Also, the parameter name changed (factor vs scale) to I just left it as positional.

ml-agents/mlagents/trainers/tf.py
```diff
@@ -0,0 +1,19 @@
+import tensorflow as tf
+
+# TODO better version check, this will do for now though
+is_tf2 = tf.__version__ == "2.0.0"
```
chriselion: There's definitely better ways to do this, just wanted something up and running initially.

unityjeffrey:
Hi @chriselion, had a question.  I believe 2.0 does offer some backward compatibility to 1.x versions.  However, my understanding is, without making the traininer code 2.0 native, you don't get the major advantages of 2.0.  With that, when we say "2.0.0 support", does that mean we are going to (or need to) also make our code 2.0 native?

chriselion:
@unityjeffrey - correct, the goal of this PR is not to take advantage of any 2.0 features. Currently if someone has tensorflow==2.0.0 installed, ml-agents will break (and it will complain at `pip install` time first). After this PR, we'll be compatible with 2.0.0

As for "2.0 native", I'll have to let other tf experts chime in. I doubt that there's a way to do that without breaking 1.x compat, so we'll have to decide if/when the new features are compelling enough to drop support for users on the old version.

ervteng:
@unityjeffrey IMO TF 2.0 will make our trainer codebase a lot cleaner and more modular. However, as @chriselion says, it will completely break 1.0 compatibility, and it will take a substantial effort to change over.

unityjeffrey:
ah i see.  makes sense.  would it be suffice to say that we upgrade to TF 2.0 now (but do not make it 2.0 ready).  But this will enable us later in the future to take on more work to make things cleaner and more modular (when we are ready)?

ervteng:
> ah i see. makes sense. would it be suffice to say that we upgrade to TF 2.0 now (but do not make it 2.0 ready). But this will enable us later in the future to take on more work to make things cleaner and more modular (when we are ready)?

Yes

chriselion:
@unityjeffrey Updated the PR title to "Allow usage with tensorflow 2.0.0 (via tf.compat.v1)", so hopefully that's less ambiguous about the goals.

ml-agents/mlagents/trainers/components/bc/model.py
```diff
@@ -1,4 +1,5 @@
-import tensorflow as tf
+from mlagents.trainers import tf
```
chriselion: @ervteng @awjuliani @andrewcoh (research pod folks) How do you feel about importing this way instead (instead of `import tensorflow as tf`)? Would something like `import mlagents.tensorflow as tf` be nicer?

Note that either way, we should prevent directly importing "raw" tensorflow except for one central spot; I added some checks to flake8 to prevent this.

ervteng: Either should be OK. I kind of like the current way rather than the other one since it seems less like we have a "special" version of TensorFlow built into mlagents.

chriselion: By "current way", do you mean the one in the PR now, or on develop?

I feel like `mlagents.trainers` is the wrong place for it; maybe `mlagents.utils` or something like that?

ervteng: The one in the PR. 

There could definitely be an `mlagents.tf_utils` file - there are actually some other generic trainer functions that could go in there (e.g. `get_variable` in the SAC model, any checkpointing functionality, etc.).

chriselion: Moved the import wrapper to `mlagents.tf_utils`

chriselion:
I did a full training run last week using tf 2.0, and ran some of the models for inference; behavior looked good.

ml-agents/mlagents/tf_utils/tf.py
```diff
@@ -0,0 +1,26 @@
...
+    tf_flatten = tf.layers.flatten
+    tf_rnn = tf.nn.rnn_cell
+
+    tf.disable_v2_behavior()
+else:
+    import tensorflow.contrib.layers as c_layers
+
+    tf_variance_scaling = c_layers.variance_scaling_initializer
+    tf_flatten = c_layers.flatten
+    tf_rnn = tf.contrib.rnn
```
ervteng: I actually think we might not need these contrib layers. The RNN and flatten even in TF 1.7 exist as `tf.layers.flatten` and `tf.nn.rnn` which is aliased to `tf.contrib` equivalents... I think

chriselion: Yeah, you're right. Not sure why I thought I needed those originally. I think we still need the "import tensorflow.compat.v1 as tf" though.

PR #2679, opened 2019-10-04
Fix LSTM+BC in discrete case

ervteng:
Looks like the LSTM in the discrete + BC case wasn't being used. This fixes the bug.


chriselion:
:shipit:

PR #2693, opened 2019-10-08
Scale rendertexture with screen

ervteng:
The GridWorld UI gets too big depending on your screen resolution. Scale it with the screen size.


ervteng:
Fixed in main branch - closing this PR

PR #2702, opened 2019-10-09
exclude bair.berkeley.edu temporarily

chriselion:
These aren't responding, so temporarily whitelist them.

I created MLA-162 as a reminder to undo this one it's fixed.

### DIFF
--- a/markdown-link-check.config.json
+++ b/markdown-link-check.config.json
@@ -7,6 +7,10 @@
         {
             "pattern": "^https://developer.nvidia.com/compute/machine-learning/cudnn/secure",
             "comment": "Requires login"
+        },
+        {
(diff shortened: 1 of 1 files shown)


markdown-link-check.config.json
```diff
@@ -7,6 +7,10 @@
         {
             "pattern": "^https://developer.nvidia.com/compute/machine-learning/cudnn/secure",
             "comment": "Requires login"
+        },
+        {
+            "pattern": "^https?://bair.berkeley.edu",
```
chriselion: We link to both http and https, so this catches both.

chriselion:
The failing link right now is a unity one, rerunning.

PR #2703, opened 2019-10-09
Barracuda hotfix for TF2 AddV2 support

mantasp:


chriselion:
markdown_link_check is a known problem that's fixed on develop.

PR #2707, opened 2019-10-10
Check for unnecessary list comprehensions

chriselion:
Small python cleanup - lines like `any([x for x in y if z])` are slightly wasteful because they form the whole list (no opportunity to early out). This PR adds enforcement for this, and fixes the few cases where it was being used.

### DIFF
--- a/ml-agents/mlagents/trainers/meta_curriculum.py
+++ b/ml-agents/mlagents/trainers/meta_curriculum.py
@@ -1,6 +1,7 @@
 """Contains the MetaCurriculum class."""
 
 import os
+from typing import Any, Dict, Set
 from mlagents.trainers.curriculum import Curriculum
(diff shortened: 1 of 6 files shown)


ml-agents/mlagents/trainers/meta_curriculum.py
```diff
@@ -38,22 +41,18 @@ def __init__(self, curriculum_folder, default_reset_parameters):
...
+                config_keys: Set[str] = set(curriculum.get_config().keys())
 
                 # Check if any two curriculums use the same reset params.
-                if any(
-                    [
-                        (parameter in curriculum.get_config().keys())
-                        for parameter in used_reset_parameters
-                    ]
-                ):
+                if config_keys & used_reset_parameters:
```
chriselion: This seems cleaner as a set intersection check. Also added some type hints while I was in the neighborhood.

PR #2712, opened 2019-10-10
Develop unified version

chriselion:
This makes it so that we only need to change one file to change the version of all 3 packages.

Note that it will cause a jump in the gym-unity version, but that should be fine.

Will update the release docs after this.


gym-unity/setup.py
```diff
@@ -1,6 +1,12 @@
...
 from setuptools import setup, find_packages
+from os import path
+
+here = path.abspath(path.dirname(__file__))
+
+with open(path.join(here, "..", "VERSION")) as f:
+    version_string = f.read().strip()
 
 setup(
     name="gym_unity",
```
ervteng: gym version str

chriselion:
@xiaomaogy Yeah, that's the only downside, but I think we can mention it in the release notes.

ervteng:
Closing due to #2712

PR #2717, opened 2019-10-10
Speed up processing large vector observations

chriselion:
The speeds up processing large numbers of vector observations. For my tests on the 3DBall scene with 10000 observations and 1000 steps:

baseline: 85.593s
optimized: 66.714s

There are a few improvements over the old code:
1) Only convert from python->numpy once (used to convert once for nan check and once for final results)
2) Make the NaN checks faster https://stackoverflow.com/questions/6736590/fast-check-for-nan-in-numpy
3) Only call `np.nan_to_num` if there was a NaN or Inf in the inputs (this accounts for about 5s in the benchmark improvement.

### DIFF
--- a/ml-agents-envs/mlagents/envs/tests/test_brain.py
+++ b/ml-agents-envs/mlagents/envs/tests/test_brain.py
@@ -1,0 +1,62 @@
+import logging
+import numpy as np
+import sys
+from unittest import mock
+
+from mlagents.envs.communicator_objects.agent_info_pb2 import AgentInfoProto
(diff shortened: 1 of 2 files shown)


ml-agents-envs/mlagents/envs/tests/test_brain.py
```diff
@@ -0,0 +1,62 @@
+import logging
```
chriselion: Confirmed that this test passed on develop, aside from checking that nan_to_num wasn't called (because it always is currently).

PR #2719, opened 2019-10-10
Revert "Develop unified version"

chriselion:
Reverts Unity-Technologies/ml-agents#2712

This was breaking non-editable pip installs.


gym-unity/setup.py
```diff
@@ -1,21 +1,15 @@
...
-
-here = path.abspath(path.dirname(__file__))
-
-with open(path.join(here, "..", "VERSION")) as f:
-    version_string = f.read().strip()
 
 setup(
     name="gym_unity",
-    version=version_string,
+    version="0.4.7",
```
ervteng: Should we just tick this to 0.4.8?

chriselion: Hrm... I'd rather just undo the old PR, and make any other changes separately.

ervteng: OK, I'll re-open that PR after this one

PR #2722, opened 2019-10-11
Fix "memory leak" during inference

ervteng:
We don't clear the update buffer if we don't update the policy. This fixes that. 

Wanted to get some thoughts on this. Also, how do we test it?

### DIFF
--- a/ml-agents/mlagents/trainers/tests/test_rl_trainer.py
+++ b/ml-agents/mlagents/trainers/tests/test_rl_trainer.py
@@ -4,6 +4,7 @@ import yaml
 import mlagents.trainers.tests.mock_brain as mb
 import numpy as np
 from mlagents.trainers.rl_trainer import RLTrainer
+from mlagents.trainers.tests.test_buffer import construct_fake_buffer
 
 
 @pytest.fixture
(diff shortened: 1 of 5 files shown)


chriselion:
> Also, how do we test it?

Make a test that runs in inference mode, and assert that the buffer is empty after stepping a few times?

ervteng:
Tests added @chriselion

ml-agents/mlagents/trainers/tests/test_rl_trainer.py
```diff
@@ -92,3 +93,12 @@ def test_rl_trainer(add_policy_outputs, add_rewards_outputs, num_vis_obs):
...
         for agent_id in rewards:
             assert rewards[agent_id] == 0
+
+
+def test_clear_update_buffer():
+    trainer = create_rl_trainer()
+    trainer.training_buffer = construct_fake_buffer()
+    trainer.training_buffer.append_update_buffer(2, batch_size=None, training_length=2)
+    trainer.clear_update_buffer()
+    for _, arr in trainer.training_buffer.update_buffer.items():
```
chriselion: No is_empty() method on the buffer? :(

ervteng: No... maybe we should add it. Also I use this in the buffer: `len(next(iter(self.update_buffer.values())))` to get its length - there probably should be a length method as well.

chriselion: Your call - I'm OK with this as it is, but I think that would be a little nicer.

ervteng: I'll make a separate PR, unless we decide to refactor the buffer anyways

PR #2731, opened 2019-10-15
[WIP] ISensor interface and use for visual observations

chriselion:
Adds an ISensor interface, and implements Camera and RenderTexture Sensors. This will be more extensible for vector observations in the future, but should only be used for visual observations at the moment.

Notable interface changes / upgrade issues:
* Visual observations are now configured via Camera Sensor Component and RenderTexture Sensor Components. This will require changes to any scenes which were previously using visual observations.
* AgentParameters no longer contains Cameras and RenderTexture.
* BrainParameters no longer requires setting Camera Resolutions. The resolutions are now set on Camera Sensor Component and RenderTexture Sensor Components.


UnitySDK/Assets/ML-Agents/Examples/GridWorld/Scenes/GridWorld.unity
```diff
@@ -175,6 +175,16 @@ Transform:
   m_Father: {fileID: 0}
   m_RootOrder: 5
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
+--- !u!20 &57863226 stripped
+Camera:
+  m_PrefabParentObject: {fileID: 20743940359151984, guid: 5c2bd19e4bbda4991b74387ca5d28156,
```
chriselion: I'm probably guilty of crimes against prefabs. Will cleanup eventually.

UnitySDK/Assets/ML-Agents/Scripts/Sensor/RenderTextureSensor.cs
```diff
@@ -0,0 +1,46 @@
+using System;
+using MLAgents.InferenceBrain;
+using UnityEngine;
+
+namespace MLAgents.Sensor
+{
+    class RenderTextureSensor : SensorBase
```
chriselion: This is almost identical to the CameraSensor, not sure if it's worth trying to de-dedupe some of the code.

UnitySDK/Assets/ML-Agents/Scripts/Utilities.cs
```diff
@@ -28,31 +28,67 @@ public static class Utilities
             var numTextures = textures.Count;
             var width = textures[0].width;
             var height = textures[0].height;
-            var data = tensorProxy.data;
```
chriselion: Old code wrote all the Textures to the tensor at once. Now we'll do it one texture at a time.

UnitySDK/Assets/ML-Agents/Scripts/Sensor/CameraSensor.cs
```diff
@@ -0,0 +1,46 @@
+using System;
+using MLAgents.InferenceBrain;
+using UnityEngine;
+
+namespace MLAgents.Sensor
+{
+    public class CameraSensor : SensorBase
```
vincentpierre: Why is this implementing a SensorBase and not a ISensor?

UnitySDK/Assets/ML-Agents/Scripts/Sensor/RenderTextureSensor.cs
vincentpierre: They use the same Utility under the hood, I think it is fine.

UnitySDK/Assets/ML-Agents/Scripts/Sensor/CameraSensor.cs
vincentpierre: If it is to get the monobehavior, it could inherit `ISensor, MonoBehavior`

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -595,25 +611,18 @@ void SendInfoToBrain()
...
-            var camCount = agentParameters.agentCameras.Count;
-            for (var i = 0; i < agentParameters.agentRenderTextures.Count; i++)
-            {
-                var obsTexture = ObservationToTexture(
-                    agentParameters.agentRenderTextures[i],
-                    param.cameraResolutions[camCount + i].width,
-                    param.cameraResolutions[camCount + i].height);
-                m_Info.visualObservations.Add(obsTexture);
+                // TODO separate this from inference time
+                var sensor = m_Sensors[i];
```
surfnerd: that looks much nicer

UnitySDK/Assets/ML-Agents/Scripts/Sensor/CameraSensor.cs
chriselion: Yeah, it could do that. But my plan for Agent setup was to call `GetComponents<SensorBase>()` (and allow the user to provide their own list of ISensors if they have custom implementations)

UnitySDK/Assets/ML-Agents/Scripts/Sensor/RenderTextureSensor.cs
chriselion: Not quite, ObservationToTexture is overloaded :)

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -314,6 +323,12 @@ void OnEnableHelper(Academy academy)
         {
             m_Info = new AgentInfo();
             m_Action = new AgentAction();
+            m_Sensors = new List<ISensor>();
+
+            // TODO deterministic sorting
+            var attachedSensors = GetComponents<SensorBase>();
```
surfnerd: could init this with the capacity of of `atachedSensors` and add in a loop to avoid extra GC allocations.  But no big deal

UnitySDK/Assets/ML-Agents/Scripts/Grpc/GrpcExtensions.cs
```diff
@@ -38,15 +38,12 @@ public static AgentInfoProto ToProto(this AgentInfo ai)
...
+            foreach (var obs in ai.compressedObservations)
             {
-                using (TimerStack.Instance.Scoped("encodeVisualObs"))
-                {
-                    agentInfoProto.VisualObservations.Add(
-                        ByteString.CopyFrom(obs.EncodeToPNG())
-                    );
-                }
+                // TODO pass compression type and dimensions separately.
+                agentInfoProto.VisualObservations.Add(ByteString.CopyFrom(obs.Data));
```
surfnerd: Could you add a Timer/Profiler around this.  Would be interesting to keep track of this, or maybe one level up, for conversion of all Sensors --> Proto

UnitySDK/Assets/ML-Agents/Scripts/Sensor/CameraSensor.cs
```diff
@@ -0,0 +1,46 @@
...
+        public bool grayscale;
+
+        public override int[] GetFloatObservationShape()
+        {
+            return new [] {width, height, grayscale ? 1 : 3};
+        }
+
+        public override void WriteObservation(float[] observationsOut)
+        {
+            throw new NotImplementedException("Have to use compression");
```
surfnerd: what about inference?

UnitySDK/Assets/ML-Agents/Scripts/Utilities.cs
```diff
@@ -28,31 +28,67 @@ public static class Utilities
...
+            bool grayScale,
+            int textureOffset = 0)
+        {
+            var width = texture.width;
+            var height = texture.height;
+            var data = tensorProxy.data;
+
+            var t = textureOffset;
+            var texturePixels = texture.GetPixels32();
+            for (var h = height - 1; h >= 0; h--)
```
surfnerd: can we add a comment here about `h--`

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
chriselion: Yeah, I think you can also do `GetComponents<SensorBase>(m_Sensors)` too. I'll probably break this out to a virtual method so users can hook in their own ISensors too.

UnitySDK/Assets/ML-Agents/Scripts/Sensor/CameraSensor.cs
chriselion: I'm punting on that for now, partly because I'm still trying to decide what the best thing to do about the Y flip is, and how to keep it consistent between training and inference.

UnitySDK/Assets/ML-Agents/Scripts/Utilities.cs
```diff
@@ -28,31 +28,67 @@ public static class Utilities
...
 
             for (var t = 0; t < numTextures; t++)
             {
-                var texturePixels = textures[t].GetPixels32();
-                for (var h = height - 1; h >= 0; h--)
+                var texture = textures[t];
+                if (width != texture.width || height != texture.height)
                 {
-                    for (var w = 0; w < width; w++)
+                    // TODO throw? We didn't do anything about this before.
```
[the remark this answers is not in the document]
chriselion: Yeah, not a bad idea. I think we'll end up enforcing it sooner (check sensor sizes when connecting agents to the brain) but worth adding some defense-in-depth.

UnitySDK/Assets/ML-Agents/Scripts/Utilities.cs
chriselion: Yeah, this is the old code but probably unindented one level. Need to merge that other PR and update this.

UnitySDK/Assets/ML-Agents/Scripts/Sensor/CameraSensor.cs
vincentpierre: Ah, and that requires SensorBase to inherit MonoBehaviour, I see.

chriselion: I'm open to other suggestions, but that definitely seemed like the easiest way to allow setting up the sensors in the editor.

vincentpierre: Still I would have : 
```csharp
interface ISensor {}
abstract SensorComponent : ISensor, MonoBehavior  {}//(is a component and a ISensor and can use GetComponents<SensorComponent>())
abstract SensorBase : ISensor  {}// Not a component but is provides a simplified way to collect observations
```

chriselion: But then what about RayPerception, which we want to be a SensorComponent (it's already a MonoBehaviour) but want to produce floats for observations? It would be OK if we put the float observations in ISensor.

UnitySDK/Assets/ML-Agents/Examples/Template/Scripts/TemplateDecision.cs
```diff
@@ -16,7 +16,7 @@ public class TemplateDecision : Decision
 
     public override List<float> MakeMemory(
         List<float> vectorObs,
-        List<Texture2D> visualObs,
+        List<MLAgents.Sensor.ISensor> sensors,
```
surfnerd: Any reason not to import the Sensor namespace?

UnitySDK/Assets/ML-Agents/Scripts/InferenceBrain/GeneratorImpl.cs
```diff
@@ -312,11 +312,17 @@ public class VisualObservationInputGenerator : TensorGenerator.IGenerator
...
         {
-            var textures = agents.Select(
-                agent => agent.Info.visualObservations[m_Index]).ToList();
-
             TensorUtils.ResizeTensor(tensorProxy, batchSize, m_Allocator);
-            Utilities.TextureToTensorProxy(textures, tensorProxy, m_GrayScale);
+            var agentIndex = 0;
+            foreach (var agent in agents)
+            {
+                // TODO direct access to sensors list here - should we do it differently?
```
surfnerd: It could be beneficial to group like sensors by type, instead of relying on an index into a list on each agent.  So you can just iterate through all of the CameraSensors of all agents instead of `agent.m_Sensors[index]`.  Though I'd have to look at the code more to see if that's a feasible option.

UnitySDK/Assets/ML-Agents/Examples/Template/Scripts/TemplateDecision.cs
chriselion: No reason, I was just being lazy.

UnitySDK/Assets/ML-Agents/Scripts/InferenceBrain/GeneratorImpl.cs
vincentpierre: Ideally, it does not mater if the sensor is a camera or not. We should ustt make sure there is an ordered mapping between python and inference. We could all each sensor input `sensor_<n>` where n is the sensor index. Another way to go would be to name each sensor.

UnitySDK/Assets/ML-Agents/Scripts/Grpc/RpcCommunicator.cs
```diff
@@ -424,6 +425,53 @@ private void HandleOnPlayModeChanged(PlayModeStateChange state)
...
+        {
+            if (m_sentBrainKeys.Contains(brainKey))
+            {
+                return;
+            }
+
+            m_unsentBrainKeys[brainKey] = brainParameters;
+        }
+
+        private CommunicatorObjects.UnityRLInitializationOutputProto GetTempUnityRlInitializationOutput()
```
surfnerd: nit: `CommunicatorObjects` is already imported.

UnitySDK/Assets/ML-Agents/Examples/Basic/Scripts/BasicDecision.cs
```diff
@@ -1,12 +1,12 @@
...
+﻿using System.Collections.Generic;
 using UnityEngine;
 using MLAgents;
 
 public class BasicDecision : Decision
 {
     public override float[] Decide(
         List<float> vectorObs,
-        List<Texture2D> visualObs,
+        List<MLAgents.Sensor.ISensor> sensors,
```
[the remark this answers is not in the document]
chriselion: Wasn't sure what the convention was - I assumed subfolder => new namespace. I can combine them.

UnitySDK/Assets/ML-Agents/Scripts/Grpc/RpcCommunicator.cs
```diff
@@ -424,6 +425,53 @@ private void HandleOnPlayModeChanged(PlayModeStateChange state)
             }
         }
 
+        private void CacheBrainParameters(string brainKey, BrainParameters brainParameters)
+        {
+            if (m_sentBrainKeys.Contains(brainKey))
+            {
+                return;
```
vincentpierre: I would almost want to compare the brainParameters here to the cached ones to ensure they are the same (with assert). This might be too much work for something that will change soon.

UnitySDK/Assets/ML-Agents/Scripts/Grpc/RpcCommunicator.cs
```diff
@@ -424,6 +425,53 @@ private void HandleOnPlayModeChanged(PlayModeStateChange state)
...
+            }
+
+            foreach (var brainProto in output.BrainParameters)
+            {
+                m_sentBrainKeys.Add(brainProto.BrainName);
+                m_unsentBrainKeys.Remove(brainProto.BrainName);
+            }
+
```
vincentpierre: extra line

.circleci/config.yml
```diff
@@ -146,7 +146,7 @@ jobs:
             . venv/bin/activate
             pip install --upgrade pip
             pip install grpcio-tools==1.13.0  --progress-bar=off
-            pip install mypy-protobuf --progress-bar=off
+            pip install mypy-protobuf==1.16.0 --progress-bar=off
```
chriselion: This was causing discrepancies between locally generated files and the CI ones, due to https://github.com/dropbox/mypy-protobuf/pull/92. I can undo this and apply it in another PR to cutdown on noise (it touches all the .pyi files) if that's better.

UnitySDK/Assets/ML-Agents/Scripts/Grpc/RpcCommunicator.cs
```diff
@@ -258,6 +258,8 @@ public void PutObservations(string brainKey, ICollection<Agent> agents)
             {
                 foreach (var agent in m_CurrentAgents[brainKey])
                 {
+                    // Update the sensor data on the AgentInfo
+                    agent.GenerateSensorData();
```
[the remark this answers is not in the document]
chriselion: I think CollectObservation will eventually go away. This is currently here instead of the general agent step because we don't want it called during inferrence (we want to write directly to the tensors then).

UnitySDK/Assets/ML-Agents/Scripts/Grpc/RpcCommunicator.cs
chriselion: Sure, that's reasonable.

UnitySDK/Assets/ML-Agents/Scripts/Grpc/RpcCommunicator.cs
chriselion: Fixed (and cleaned up a few other places where it was being used).

UnitySDK/Assets/ML-Agents/Scripts/InferenceBrain/ModelRunner.cs
```diff
@@ -105,6 +107,19 @@ public void PutObservations(string key, ICollection<Agent> agents)
                 return;
             }
 
+            if (!m_visualObservationsInitialized)
+            {
+                // Just grab the first agent in the collection (any will do).
```
chriselion: Rewriting using FirstOrDefault()

UnitySDK/Assets/ML-Agents/Scripts/Sensor/CameraSensor.cs
surfnerd: We could have SensorComponent as a holder for an ISesnor if we want to avoid having ISensors all implanted as MonoBehaviours.

.circleci/config.yml
surfnerd: Yeah, maybe merge this protobuf change before the sensor change.

UnitySDK/Assets/ML-Agents/Examples/Basic/Scripts/BasicDecision.cs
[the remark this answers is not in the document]
surfnerd: I think the namespace is fine.  I'd prefer that if we are going to reference `MLAgents.Sensor.ISensor` that we import it and reference the type directly though.

UnitySDK/Assets/ML-Agents/Scripts/InferenceBrain/GeneratorImpl.cs
surfnerd: I think my suggestion was more around how we organize our data.  Out of the scope of this PR.

.circleci/config.yml
chriselion: https://github.com/Unity-Technologies/ml-agents/pull/2757

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -124,12 +121,14 @@ public class AgentParameters
         /// The list of the Camera GameObjects the agent uses for visual
         /// observations.
         /// </summary>
+        // TODO REMOVE!
```
chriselion: Still TODO

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -546,6 +549,23 @@ public virtual void InitializeAgent()
...
         }
 
+        /// <summary>
+        /// Set up the list of ISensors on the Agent. By default, this will select any
+        /// SensorBase's attached to the Agent. To add custom ISensor implementations that are not attached,
+        /// override this method and add them there.
+        /// </summary>
+        public virtual void InitializeSensors()
+        {
+            // TODO deterministic sorting
```
chriselion: Any thoughts on this? Require a name in the sensor (and sensor component)?

UnitySDK/Assets/ML-Agents/Scripts/Brain.cs
```diff
@@ -34,6 +37,30 @@ public abstract class Brain : ScriptableObject
...
+                // First agent, save the sensor sizes
+                foreach(var sensor in agent.m_Sensors)
+                {
+                    m_SensorShapes.Add(sensor.GetFloatObservationShape());
+                }
+            }
+            else
+            {
+                // Check for compatibility with the other Agents' sensors
+                // TODO make sure this only checks once per agent
```
chriselion: I added this as a sanity check. Not sure if we should disable it, only enable in Development, or always do it. Hopefully will get easier with Vince's changes.

UnitySDK/Assets/ML-Agents/Scripts/InferenceBrain/GeneratorImpl.cs
chriselion: @vincentpierre Do you think this is a blocker for this PR? I agree we need a more concrete mapping, but not sure if it can wait until I make a second pass on this.

UnitySDK/Assets/ML-Agents/Scripts/Sensor/SensorComponent.cs
```diff
@@ -0,0 +1,10 @@
+using System;
+using UnityEngine;
+
+namespace MLAgents.Sensor
```
chriselion: TODO docstring

UnitySDK/ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +1 @@
-m_EditorVersion: 2017.4.32f1
+m_EditorVersion: 2017.4.29f1
```
chriselion: Meant to revert this :/

ml-agents/mlagents/trainers/tests/mock_brain.py
```diff
@@ -235,3 +235,26 @@ def create_mock_banana_brain():
         vector_observation_space_size=0,
     )
     return mock_brain
+
+
+def make_brain_parameters(
```
chriselion: Need to try to combine this with the other methods in this file.

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
vincentpierre: We need this to be sorted because : 
 - we need all agents who have same sensors to be identifiable
 - we need that even when the user shuffles the sensors, it still works
I don't see another way besides naming them. We could use reflexion to do that (using the type of the sensor if we can ensure there will only ever be one type of sensor per agent...)

UnitySDK/Assets/ML-Agents/Scripts/Sensor/CameraSensorComponent.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+using UnityEngine;
+
+namespace MLAgents.Sensor
+{
+    class CameraSensorComponent : SensorComponent
```
vincentpierre: I think this needs to have a list of Sensors (Since currently we support multiple cameras per agents.

UnitySDK/Assets/ML-Agents/Scripts/Sensor/CameraSensor.cs
chriselion: Updated, SensorComponent now produces an ISensor.

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
chriselion: I now do `m_Sensors.Capacity += attachedSensorComponents.Length;` before appending the ISensors

UnitySDK/Assets/ML-Agents/Scripts/Grpc/GrpcExtensions.cs
chriselion: Added timers in GrpcCommunicator around the whole AgentInfo.ToProto loop, and also around the bodies of the ISensor methods.

UnitySDK/Assets/ML-Agents/Scripts/Sensor/CameraSensorComponent.cs
chriselion: Couldn't you just add multiple CameraSensorComponents in that case? Seems easier to keep them 1:1 on the Component.

vincentpierre: I do not think it is possible to attach multiple of the same component to a GameObject in Unity. If there is a way then yes, makes more sense.

chriselion: image

surfnerd: You can have multiple components with the same type.  And then access them with the `GetComponents` API.

UnitySDK/Assets/ML-Agents/Scripts/Grpc/RpcCommunicator.cs
chriselion: Punting on this for now

UnitySDK/Assets/ML-Agents/Scripts/Sensor/CameraSensorComponent.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public new Camera camera;
+        public string sensorName = "CameraSensor";
```
chriselion: I think we can revisit this and try to make a unique name if adding another sensor component of the same type. It's OK for now though.

UnitySDK/Assets/ML-Agents/Scripts/Sensor/RenderTextureSensor.cs
```diff
@@ -0,0 +1,97 @@
...
+        private int[] m_Shape;
+
+        public RenderTextureSensor(RenderTexture renderTexture, int width, int height, bool grayscale, string name)
+        {
+            m_RenderTexture = renderTexture;
+            m_Width = width;
+            m_Height = height;
+            m_Grayscale = grayscale;
+            m_Name = name;
+            m_Shape = new[] { width, height, grayscale ? 1 : 3 };
```
chriselion: Note: shape is cached now, so no extra allocations at runtime.

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
chriselion: Done - ISensor now has a GetName() method. Sensors are sorted by that and checked for uniqueness

UnitySDK/Assets/ML-Agents/Scripts/Sensor/RenderTextureSensor.cs
chriselion: (and same for CameraSensor)

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -546,6 +525,39 @@ public virtual void InitializeAgent()
...
+            for (var i = 0; i < m_Sensors.Count - 1; i++)
+            {
+                Debug.Assert(!m_Sensors[i].GetName().Equals(m_Sensors[i+1].GetName()), "Sensor names must be unique.");
+            }
+        }
+
+        /// <summary>
+        /// Create additional ISensors to be added to Agent.
+        /// </summary>
+        public virtual void AddCustomSensors()
```
vincentpierre: I would say this is a little premature. I would rather not have this additional virtual method. (especially if we don't show how to use them in the Examples. If anything, make this a : 
```csharp
public virtual IEnumerator<ISensor> AddCustomSensors(){ }
```

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -625,15 +605,37 @@ void SendInfoToBrain()
...
             }
 
             m_Info.textObservation = "";
         }
 
-        public void ClearVisualObservations()
+        /// <summary>
+        /// Generate data for each sensor and store it on the Agent's AgentInfo
+        /// </summary>
+        public void GenerateSensorData()
```
vincentpierre: Is there a reason the Compressed data is in the AgentInfo ? Would it be easier to generate it in the Communicator and the recorder? Maybe this is good for now.

vincentpierre: Can you document in the code what the future design should look like ?

UnitySDK/Assets/ML-Agents/Scripts/Brain.cs
```diff
@@ -34,6 +40,30 @@ public abstract class Brain : ScriptableObject
         public void SubscribeAgentForDecision(Agent agent)
         {
             LazyInitialize();
+            if (m_SensorShapes.Count == 0)
```
vincentpierre: Can you put this checking code into a separate method with a fancy name so It is easier to know what it does?

UnitySDK/Assets/ML-Agents/Scripts/InferenceBrain/BarracudaModelParamLoader.cs
```diff
@@ -241,18 +241,19 @@ private static ModelActionType GetActionType(int isContinuousInt)
...
-            {
-                if (!tensorsNames.Contains(
-                    TensorNames.VisualObservationPlaceholderPrefix + visObsIndex))
-                {
-                    failedModelChecks.Add(
-                        "The model does not contain a Visual Observation Placeholder Input " +
-                        "for visual observation " + visObsIndex + ".");
-                }
-            }
+            // TODO figure out how to do this if we move everything to agents
```
vincentpierre: Dead code

UnitySDK/Assets/ML-Agents/Scripts/InferenceBrain/BarracudaModelParamLoader.cs
```diff
@@ -343,12 +344,13 @@ private static IEnumerable<string> CheckOutputTensorPresence(Model model, int me
                 tensorTester[mem.input] = ((bp, tensor) => null);
             }
 
-            for (var obsIndex = 0; obsIndex < brainParameters.cameraResolutions.Length; obsIndex++)
-            {
-                var index = obsIndex;
-                tensorTester[TensorNames.VisualObservationPlaceholderPrefix + obsIndex] =
-                    (bp, tensor) => CheckVisualObsShape(bp, tensor, index);
-            }
+            // TODO figure out how to check this if we move everything to Agent
```
vincentpierre: dead code

UnitySDK/Assets/ML-Agents/Scripts/InferenceBrain/BarracudaModelParamLoader.cs
```diff
@@ -435,19 +437,19 @@ private static IEnumerable<string> CheckOutputTensorPresence(Model model, int me
...
-            var heightT = tensorProxy.shape[1];
-            var widthT = tensorProxy.shape[2];
-            var pixelT = tensorProxy.shape[3];
-            if ((widthBp != widthT) || (heightBp != heightT) || (pixelBp != pixelT))
-            {
-                return $"The visual Observation {visObsIndex} of the model does not match. " +
-                    $"Received TensorProxy of shape [?x{widthBp}x{heightBp}x{pixelBp}] but " +
-                    $"was expecting [?x{widthT}x{heightT}x{pixelT}].";
-            }
+//            var resolutionBp = brainParameters.cameraResolutions[visObsIndex];
```
vincentpierre: dead code

UnitySDK/Assets/ML-Agents/Scripts/InferenceBrain/ModelRunner.cs
```diff
@@ -106,6 +109,14 @@ public void PutObservations(string key, ICollection<Agent> agents)
                 return;
             }
 
+            if (!m_visualObservationsInitialized)
+            {
+                // Just grab the first agent in the collection (any will suffice, really).
+                var firstAgent = agents.FirstOrDefault();
```
vincentpierre: Is it possible for FirstOrDefault to return a default (an agent that does not exist) and mess up the Initialization of the TensorGenerator ?

chriselion: A few lines above, we early out if `agents.Count` is 0, so we should always have a valid Agent to grab here. I'll change this to `First()` and a comment though, so that it's clearer.

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
chriselion: Sure, will remove for now (I was mostly using it for a test)

vincentpierre:
The AreaRemderTexture Agent in the gridworld scene is using a Camera sensor but should be using a RenderTextureSensor.
You should also add the CameraSensor to all of the other Visual Environments such as Pyramids and Pushblock

UnitySDK/Assets/ML-Agents/Scripts/Brain.cs
chriselion: Moved to ValidateAgentSensorShapes()

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
chriselion: Maybe we should talk about it some more. I think the Brain (and maybe the presence of the recorder) should determine whether we're in compressed, uncompressed, or write-to-tensor mode; this is something that could be passed as a parameter to CollectObservations, which could call GenerateSensorData.

chriselion: > Would it be easier to generate it in the Communicator and the recorder? 

The recorder operates on AgentInfos, not Agents. 

We could pass the ISensors instead of the CompressedObservations in AgentInfo, and let the AgentInfo consumer decide what to do with them...

chriselion: Added some comments

chriselion:
@vincentpierre Updated scenes - everything uses CameraSensor Component (except for the RenderTexture one on GridWorld)

UnitySDK/Assets/ML-Agents/Scripts/Utilities.cs
chriselion: Will do this after merge

UnitySDK/Assets/ML-Agents/Scripts/InferenceBrain/BarracudaModelParamLoader.cs
```diff
@@ -239,20 +239,8 @@ private static ModelActionType GetActionType(int isContinuousInt)
...
-                if (!tensorsNames.Contains(
-                    TensorNames.VisualObservationPlaceholderPrefix + visObsIndex))
-                {
-                    failedModelChecks.Add(
-                        "The model does not contain a Visual Observation Placeholder Input " +
-                        "for visual observation " + visObsIndex + ".");
-                }
-            }
+
+            // TODO reenable checks there are enough Visual Observation Placeholder in the model.
```
vincentpierre: Do you have a way to do these checks ?

UnitySDK/Assets/ML-Agents/Scripts/Sensor/CameraSensorComponent.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+using UnityEngine;
+
+namespace MLAgents.Sensor
+{
+    class CameraSensorComponent : SensorComponent
+    {
+        public new Camera camera;
```
vincentpierre: Why the `new` keyword here ?

surfnerd: Don't this it's needed here

chriselion:
```
Assets/ML-Agents/Scripts/Sensor/CameraSensorComponent.cs(8,23): warning CS0108: `MLAgents.Sensor.CameraSensorComponent.camera' hides inherited member `UnityEngine.Component.camera'. Use the new keyword if hiding was intended
```
Should I just use the base member? Not sure how that's normally used.

chriselion:
Re: the warning: the field is assigned in the UI

vincentpierre:
> Re: the warning: the field is assigned in the UI

There has to be a way to suppress this warning, we have a lot of fields that are only assigned in the inspector. I think there is a way around it.

chriselion:
Looks like marking the classes as public gets rid of the warnings (and is probably something we want anyway): https://stackoverflow.com/a/12618676/224264

UnitySDK/Assets/ML-Agents/Scripts/Sensor/CameraSensorComponent.cs
vincentpierre: Make a new one. `cAmErA` ?

UnitySDK/Assets/ML-Agents/Scripts/Brain.cs
surfnerd: wrap in `#if DEBUG`

UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -546,6 +525,29 @@ public virtual void InitializeAgent()
...
+            m_Sensors.Capacity += attachedSensorComponents.Length;
+            foreach (var component in attachedSensorComponents)
+            {
+                m_Sensors.Add(component.CreateSensor());
+            }
+
+            // Sort the sensors by name to ensure determinism
+            m_Sensors.Sort((x, y) => x.GetName().CompareTo(y.GetName()));
+
+            // Make sure the names are actually unique
```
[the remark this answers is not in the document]
chriselion: Done (and the call to ValidateAgentSensorShapes too)

UnitySDK/Assets/ML-Agents/Scripts/InferenceBrain/BarracudaModelParamLoader.cs
surfnerd: ➕

chriselion: Not currently, at least not at model load time, since we don't have any way to get the observation sizes until there's an Agent.

UnitySDK/Assets/ML-Agents/Scripts/Sensor/CameraSensorComponent.cs
vincentpierre: - `SensingCamera`?
 - `SensorCamera`?
 - `AttachedCamera`?
 - `ObservationCamera`?

chriselion: I like ObservationCamera. @surfnerd any preferences?

This means that I either need to fix up the scenes again or use a `FormerlySerializedAs`, so want to make sure I only do it once.

chriselion: Also the base member is deprecated; I'd rather just leave the `new` here unless you feel strongly.

surfnerd: Oh wow, yeah that property was removed from `Component` in 2019.2.  `new` is fine.  I didn't realize there was a camera property.

PR #2740, opened 2019-10-16
Fix crash when SAC is used with Curiosity and Continuous Actions

ervteng:
Actually fixes two bugs:
- Bug in Curiosity `signal.py` that used the wrong placeholder and minibatch element
- Bug that causes reward signals to update multiple times in SAC if there are more than one


chriselion:
Do we have anything that tests PPO+curiosity+continuous? Just want to make sure that doesn't break too.

ervteng:
I tested PPO and it doesn't break (though there's no test). In fact it's more correct than the previous implementation. Let me add one though.

Issue #2751, opened 2019-10-17
Flag to force training with CPU


caioc2:
Hi,

When using visual observations and big batch size GPU's make a lot of sense and improve the training speed considerably, in the other hand, when using vector observations with small batch size (ie 32 and alike) using GPU can make it a lot slower than CPU. 

Currently, as far as I know, you cant select which device (CPU/GPU) you want to use for training, either you have to install the tensorflow with/out GPU, disable your GPU to use CPU only or create multiple environments each one with the desired configuration.

It could be made easier by adding for example a flag "--cpu" to the ml-agents for using CPU only.

So far I've been using my hackish solution on the "learn.py" file like the following:

`parser.add_argument(
        "--cpu",
        default=False,
        action="store_true",
        help="Run with CPU only",
    )
`

`if options.cpu:
        os.environ['CUDA_VISIBLE_DEVICES'] = '-1'`

It get the job done without needing seriously changes to code, if it can be useful for other people I would like to see something like this in the ml-agents.

Thanks.

chriselion:
Hi @caioc2,
Thanks for the feedback. I'll log it as a feature request, but if you want to submit a pull request with the change, I can try to get it merged in.

As a side note, does running with `CUDA_VISIBLE_DEVICES=-1 mlagents-learn ...` work too?

caioc2:
I could not run with `CUDA_VISIBLE_DEVICES=-1 mlagents-learn` but probably one can set in another way, the only drawback is this variable may not be known by many people (I didn't know about it until now), while the flag is easier and would be in the documentation about training.

Anyway, I made the pull request.

PR #2754, opened 2019-10-18
Added --cpu flag to run using only cpu for training

caioc2:
Discussion: 
[Flag to force training with CPU #2751 ](https://github.com/Unity-Technologies/ml-agents/issues/2751)


caioc2:
I can change the target branch to develop, but I cant change the branch "from". This would cause some conflicts. Then I think it may be better to close this one and open a new PR.

PR #2773, opened 2019-10-21
Develop yamato

surfnerd:
This adds a Yamato pipeline for testing our edit mode tests on non-fork PRs against develop.

### DIFF
--- a/.yamato/csharp-tests.yml
+++ b/.yamato/csharp-tests.yml
@@ -1,6 +1,6 @@
 name: Mac Edit Mode Tests
 triggers:
-    sources:
+    branches:
       only:
       - "/develop-.*/"
     targets:
--- a/.yamato/csharp-tests.yml
+++ b/.yamato/csharp-tests.yml
@@ -6,6 +6,12 @@ triggers:
     targets:
       only:
       - "develop"
+    pull_requests:
+      - targets:
+          only:
+            - "master"
+            - "/release-.*/"
+            - "/hotfix-.*/"
 
 agent:
   type: Unity::VM::osx
--- a/.yamato/csharp-tests.yml
+++ b/.yamato/csharp-tests.yml
@@ -9,4 +9,4 @@ variables:
   LC_ALL: "en_US.UTF-8"
   UNITY_PATH: "/Users/bokken/2018.7.33f1/Unity.app/Contents/MacOS/Unity"
 commands:
-  - ./run-tests-editmode-osx.sh
+  - ./run-tests-editmode-osx-editor.sh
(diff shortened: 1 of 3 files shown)


run-tests-editmode-osx-editor.sh
```diff
@@ -0,0 +1,50 @@
...
+if [[ -f ${BOKKEN_UNITY} ]]; then 
+    UNITY=${BOKKEN_UNITY}
+else
+    UNITY=${HUB_UNITY}
+fi 
+
+pushd $(dirname "${0}") > /dev/null
+BASETPATH=$(pwd -L)
+popd > /dev/null
+
```
xiaomaogy: What is the purpose of the pushd and popd here?

surfnerd: not sure. was copy pasta :).  I can remove it.  Seems redundant.

surfnerd: I think it's a safeguard against `pwd` directory not being `.` and also guarding against relative paths.  But may not be needed.

xiaomaogy: Then let's keep it. Looks good to me.

.yamato/csharp-tests.yml
```diff
@@ -0,0 +1,19 @@
+name: Mac Edit Mode Tests
+triggers:
+    branches:
```
chriselion: What about release and hotfix branches? What about merges into develop-* branches? (The latter is less common)

surfnerd: Could update it to support those branches as well.  Adding multiple `triggers` sections `OR`s them together.

run-tests-editmode-osx-editor.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
+set -eo pipefail
+
+EDITOR_VERSION="2017.4.33f1"
```
chriselion: Can we combine this with the `UNITY_PATH` in the yaml? Or at least define it there so that we only have to change in one file?

.yamato/csharp-tests.yml
chriselion: Cool, shouldn't block this PR now, but probably something we'll want soon-ish (like whenever the next release is).

run-tests-editmode-osx-editor.sh
surfnerd: Going to remove it from the yml for now.

.yamato/csharp-tests.yml
surfnerd: fixed

PR #2789, opened 2019-10-24
Deploy new versions to pypi

chriselion:
Deploy to pypi when new releases are made. Follows the instructions from https://circleci.com/blog/continuously-deploying-python-packages-to-pypi-with-circleci/ but with a few modifications to handle 3 packages, and make it easier to switch to test.pypi.org

### DIFF
--- a/gym-unity/setup.py
+++ b/gym-unity/setup.py
@@ -5,7 +5,7 @@ import sys
 from setuptools import setup, find_packages
 from setuptools.command.install import install
 
-VERSION = "0.11.0.dev"
+VERSION = "0.10.1"
 
 
 class VerifyVersionCommand(install):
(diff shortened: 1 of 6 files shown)


gym-unity/setup.py
```diff
@@ -1,15 +1,40 @@
...
 
+import os
+import sys
 from setuptools import setup, find_packages
+from setuptools.command.install import install
+
+VERSION = "0.10.1"
+
+
+class VerifyVersionCommand(install):
```
chriselion: I don't have a good way to share this across the libraries :/

harperj: :'(

harperj: New package `ml-agents-setup-common` :troll:

chriselion: It crossed my mind.

PR #2808, opened 2019-10-28
build fixes for 2018+

chriselion:
* BuildPipeline.BuildPlayer return type change from a string to a BuildReport in 2018.1. Handles both types now.
* CompressionType is also in UnityEngine namespace, which causes conflicts. We could be explicit about the namespace where we use it, but it would be annoying for users, so just be verbose with the name instead.


UnitySDK/Assets/ML-Agents/Scripts/Agent.cs
```diff
@@ -160,7 +160,7 @@ public class AgentParameters
     /// of the environment extracts its current observation, sends them to its
     /// policy and in return receives an action. In practice,
     /// however, an agent need not send its observation at every step since very
-    /// little may have changed between successive steps.
```
chriselion: IDE must have done this. I'll revert if it's too noisy for this PR.

PR #2826, opened 2019-10-30
Fix serialization for m_TrainingConfiguration

GProulx:
Bug was introduce on #2519 and is already on `master` too.


surfnerd:
Hey @GProulx,
Thanks for finding this!

PR #2842, opened 2019-11-02
more pylint fixes

chriselion:
Uncomment and fix some more pylint warnings, namely:
* Parameters differ from overridden
* Attribute '...' defined outside `__init__`

Depending on some precommit settings, the latter sometimes fires on MultiGpuPPOPolicy, but I had a hard time fixing that one up because MultiGpuPPOPolicy.create_model will be called from the parent's init.

### DIFF
--- a/ml-agents-envs/mlagents/envs/base_unity_environment.py
+++ b/ml-agents-envs/mlagents/envs/base_unity_environment.py
@@ -9,9 +9,9 @@ class BaseUnityEnvironment(ABC):
     def step(
         self,
         vector_action: Optional[Dict] = None,
-        memory: Optional[Dict] = None,
         text_action: Optional[Dict] = None,
         value: Optional[Dict] = None,
+        custom_action: Dict[str, Any] = None,
     ) -> AllBrainInfo:
         pass
 
--- a/ml-agents-envs/mlagents/envs/env_manager.py
+++ b/ml-agents-envs/mlagents/envs/env_manager.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from typing import List, Dict, NamedTuple, Optional
+from typing import Any, List, Dict, NamedTuple, Optional
 from mlagents.envs.brain import AllBrainInfo, BrainParameters
 from mlagents.envs.policy import Policy
 from mlagents.envs.action_info import ActionInfo
@@ -24,7 +24,10 @@ class EnvManager(ABC):
 
     @abstractmethod
     def reset(
-        self, config: Dict = None, train_mode: bool = True
+        self,
+        config: Dict = None,
+        train_mode: bool = True,
+        custom_reset_parameters: Any = None,
     ) -> List[EnvironmentStep]:
         pass
 
--- a/ml-agents/mlagents/trainers/bc/trainer.py
+++ b/ml-agents/mlagents/trainers/bc/trainer.py
@@ -45,20 +45,20 @@ class BCTrainer(Trainer):
 
     def add_experiences(
         self,
-        curr_info: AllBrainInfo,
-        next_info: AllBrainInfo,
+        curr_all_info: AllBrainInfo,
+        next_all_info: AllBrainInfo,
         take_action_outputs: ActionInfoOutputs,
     ) -> None:
         """
         Adds experiences to each agent's experience history.
-        :param curr_info: Current AllBrainInfo (Dictionary of all current brains and corresponding BrainInfo).
-        :param next_info: Next AllBrainInfo (Dictionary of all current brains and corresponding BrainInfo).
+        :param curr_all_info: Current AllBrainInfo (Dictionary of all current brains and corresponding BrainInfo).
+        :param next_all_info: Next AllBrainInfo (Dictionary of all current brains and corresponding BrainInfo).
         :param take_action_outputs: The outputs of the take action method.
         """
 
         # Used to collect information about student performance.
-        info_student = curr_info[self.brain_name]
-        next_info_student = next_info[self.brain_name]
+        info_student = curr_all_info[self.brain_name]
+        next_info_student = next_all_info[self.brain_name]
         for agent_id in info_student.agents:
             self.evaluation_buffer[agent_id].last_brain_info = info_student
 
--- a/ml-agents/mlagents/trainers/components/reward_signals/gail/signal.py
+++ b/ml-agents/mlagents/trainers/components/reward_signals/gail/signal.py
@@ -126,7 +126,7 @@ class GAILRewardSignal(RewardSignal):
     def prepare_update(
         self,
         policy_model: LearningModel,
-        mini_batch_policy: Dict[str, np.ndarray],
+        mini_batch: Dict[str, np.ndarray],
         num_sequences: int,
     ) -> Dict[tf.Tensor, Any]:
         """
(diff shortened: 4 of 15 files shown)


ml-agents-envs/mlagents/envs/base_unity_environment.py
```diff
@@ -9,9 +9,9 @@ class BaseUnityEnvironment(ABC):
     def step(
         self,
         vector_action: Optional[Dict] = None,
-        memory: Optional[Dict] = None,
```
chriselion: This was missed in the memory cleanup

ml-agents-envs/mlagents/envs/base_unity_environment.py
```diff
@@ -9,9 +9,9 @@ class BaseUnityEnvironment(ABC):
...
         text_action: Optional[Dict] = None,
         value: Optional[Dict] = None,
+        custom_action: Dict[str, Any] = None,
```
chriselion: These are being passed (for now) by the implementations, but will likely go away soon.

ml-agents-envs/mlagents/envs/env_manager.py
```diff
@@ -24,7 +24,10 @@ def step(self) -> List[EnvironmentStep]:
 
     @abstractmethod
     def reset(
-        self, config: Dict = None, train_mode: bool = True
+        self,
+        config: Dict = None,
+        train_mode: bool = True,
+        custom_reset_parameters: Any = None,
```
chriselion: Being passed by implementations.

ml-agents/mlagents/trainers/bc/trainer.py
```diff
@@ -45,20 +45,20 @@ def __init__(self, brain, trainer_parameters, training, load, seed, run_id):
 
     def add_experiences(
         self,
-        curr_info: AllBrainInfo,
-        next_info: AllBrainInfo,
+        curr_all_info: AllBrainInfo,
```
chriselion: Renamed these to match parent

ml-agents/mlagents/trainers/components/reward_signals/gail/signal.py
```diff
@@ -126,7 +126,7 @@ def check_config(
     def prepare_update(
         self,
         policy_model: LearningModel,
-        mini_batch_policy: Dict[str, np.ndarray],
+        mini_batch: Dict[str, np.ndarray],
```
chriselion: Renamed to match parent

ml-agents/mlagents/trainers/ppo/trainer.py
```diff
@@ -79,15 +79,15 @@ def __init__(
...
     def process_experiences(
-        self, current_info: AllBrainInfo, new_info: AllBrainInfo
+        self, current_info: AllBrainInfo, next_info: AllBrainInfo
     ) -> None:
         """
         Checks agent histories for processing condition, and processes them as necessary.
         Processing involves calculating value and advantage targets for model updating step.
         :param current_info: Dictionary of all current brains and corresponding BrainInfo.
-        :param new_info: Dictionary of all next brains and corresponding BrainInfo.
+        :param next_info: Dictionary of all next brains and corresponding BrainInfo.
```
chriselion: We seemed to be 50/50 on new vs next for the names here. I went with next_info because it seems more consistent with current_info

ml-agents/mlagents/trainers/sac/models.py
```diff
@@ -46,9 +46,41 @@ def __init__(
         self.activ_fn = self.swish
 
         self.policy_memory_in: Optional[tf.Tensor] = None
+        self.policy_memory_out: Optional[tf.Tensor] = None
```
chriselion: Not sure if there's a better way to group all of these.

ml-agents/mlagents/trainers/sac/policy.py
```diff
@@ -188,7 +188,12 @@ def evaluate(self, brain_info: BrainInfo) -> Dict[str, np.ndarray]:
 
     @timed
     def update(
-        self, mini_batch: Dict[str, Any], num_sequences: int, update_target: bool = True
+        # pylint: disable=arguments-differ
+        # TODO ervteng FIX ME
+        self,
```
chriselion: @ervteng Can you take a look at this? `update_target` is not in the base class signature. I only did a quick search, but it looked like we only ever call this with `update_target=True` but wasn't 100% sure if it was safe to remove

ml-agents/mlagents/trainers/sac/models.py
ervteng: Ahh, this is ugly (one more reason to move to TF 2.0). I can take a stab at grouping by comment (e.g. divvy up policy vs. critic vs. output tensors) but they are all needed. 

I'm assuming we'll have a similar block in the PPO model?

ml-agents/mlagents/trainers/sac/policy.py
ervteng: Hmm, this is a subtlety of the SAC algorithm. There are two value networks, a target and a current one. The target is used to bootstrap the current one, and is updated periodically towards the weights of the current. 

Most implementations of SAC let you vary when this update happens. Initially, I had an additional hyperparameter that did just that, but removed it b/c realistically you can get away with just adjusting tau (and less hyperparameters are better for most users). 

I think it's safe to remove it for now. A researcher might want to vary this, but then again a researcher can edit the Python code. Perhaps we can add a comment -> I'll add it below.

ervteng: Line 213: `# Update target network. By default, target update happens at every policy update.`

ml-agents/mlagents/trainers/sac/models.py
chriselion: it's OK as it is. One way to DRY it up a bit would be something like
```
class QInfo
    self.q: Optional[tf.Tensor] = None
    self.q_p: Optional[tf.Tensor] = None
    self.q_memory_in: Optional[tf.Tensor] = None
    self.q_memory_out: Optional[tf.Tensor] = None
    self.q_heads: Optional[Dict[str, tf.Tensor]] = None
    self.q_pheads: Optional[Dict[str, tf.Tensor]] = None
    ....
class SACNetwork(LearningModel):
    self.q1 = QInfo()
    self.q2 = QInfo()
```
but that might not be a good abstraction (and probably not a good name).

Surprisingly it's not complaining about PPO right now.

ml-agents/mlagents/trainers/sac/policy.py
chriselion: Sounds good, I'll apply that. If you want to keep the flexibility, you could so something like move all the current `update()` to `_update()`, and make `update()` call `_update()` with `update_target=True` (or some other logic)

ervteng: 👍 I'll probably do something like this when we rehaul the Policy and Trainer APIs, but will hold off for this particular version of the code.

ml-agents/mlagents/trainers/sac/models.py
ervteng: Hmm the class nesting doesn't seem very obvious. I'd almost want a group that is model inputs and one that is model outputs - but that seems to require more code changes, so let's leave it for now.

PR #2847, opened 2019-11-04
speed up slow unit test

chriselion:
Takes the test time for 
```
pytest ml-agents/mlagents/trainers/tests/test_simple_rl.py::test_simple_sac
```
from 17.63 seconds down to  11.64 seconds on my laptop.


chriselion:
That helped a bit, along with dropping the number of steps

PR #2852, opened 2019-11-05
Add ISensor.Update()

chriselion:
This add an ISensor.Update() method and implements it for the cases where it makes sense. This prevents observations from being cleared before they can be sent to the DemonstrationRecorder.

### DIFF
--- a/UnitySDK/Assets/ML-Agents/Editor/Tests/DemonstrationTests.cs
+++ b/UnitySDK/Assets/ML-Agents/Editor/Tests/DemonstrationTests.cs
@@ -102,24 +102,20 @@ namespace MLAgents.Tests
                 BindingFlags.Instance | BindingFlags.NonPublic);
             var agentEnableMethod = typeof(Agent).GetMethod("OnEnable",
                 BindingFlags.Instance | BindingFlags.NonPublic);
+            var agentSendInfo = typeof(Agent).GetMethod("SendInfo",
+                BindingFlags.Instance | BindingFlags.NonPublic);
 
             agentEnableMethod?.Invoke(agent1, new object[] { });
             academyInitializeMethod?.Invoke(aca, new object[] { });
 
             // Step the agent
             agent1.RequestDecision();
-            // TODO grab with reflection?
-            agent1.SendInfo();
+            agentSendInfo?.Invoke(agent1, new object[] { });
+
             demoRecorder.Close();
 
             // Read back the demo file and make sure observations were written
-
-            var filename = DemonstrationStore.k_DemoDirecory + demoRecorder.demonstrationName + DemonstrationStore.k_ExtensionType;
-            Assert.AreEqual(filename, "Assets/Demonstrations/TestBrain.demo");
-            var reader = fileSystem.File.OpenRead(filename);
-
-            DemonstrationMetaProto.Parser.ParseDelimitedFrom(reader);
-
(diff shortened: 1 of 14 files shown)


UnitySDK/Assets/ML-Agents/Editor/Tests/DemonstrationTests.cs
```diff
@@ -57,5 +60,73 @@ public void TestStoreInitalize()
...
+            var agentEnableMethod = typeof(Agent).GetMethod("OnEnable",
+                BindingFlags.Instance | BindingFlags.NonPublic);
+            var agentSendInfo = typeof(Agent).GetMethod("SendInfo",
+                BindingFlags.Instance | BindingFlags.NonPublic);
+
+            agentEnableMethod?.Invoke(agent1, new object[] { });
+            academyInitializeMethod?.Invoke(aca, new object[] { });
+
+            // Step the agent
+            agent1.RequestDecision();
```
chriselion: Unfortunately this doesn't fail without the fix in place. Might need to discuss a better way to get the test Agent to act like it's in inference mode.

UnitySDK/Assets/ML-Agents/Scripts/DemonstrationRecorder.cs
```diff
@@ -35,15 +36,16 @@ void Update()
         /// <summary>
         /// Creates demonstration store for use in recording.
         /// </summary>
-        void InitializeDemoStore()
+        public void InitializeDemoStore(IFileSystem fileSystem = null)
```
chriselion: Allow for the mock file system to be injected. Could keep this private and do it by reflection, but not sure if that's necessary.

UnitySDK/Assets/ML-Agents/Scripts/DemonstrationRecorder.cs
```diff
@@ -71,14 +73,22 @@ public void WriteExperience(AgentInfo info)
             m_DemoStore.Record(info);
         }
 
+        public void Close()
```
chriselion: Should this happen on disable instead? Need it for testing to make sure the write stream is flushed,

UnitySDK/Assets/ML-Agents/Scripts/DemonstrationStore.cs
```diff
@@ -21,12 +22,14 @@ public class DemonstrationStore
 
         public DemonstrationStore(IFileSystem fileSystem)
         {
-            m_FileSystem = fileSystem;
-        }
-
-        public DemonstrationStore()
-        {
-            m_FileSystem = new FileSystem();
+            if (fileSystem != null)
```
chriselion: Combined these.

UnitySDK/Assets/ML-Agents/Scripts/DemonstrationRecorder.cs
surfnerd: would you move the initialization to OnEnable?

chriselion: It's already done lazily in Update()

surfnerd: should it be set to `null` after it's closed?  Is there a use case for stopping recording during play mode and resuming?

PR #2876, opened 2019-11-08
ingore attribute-defined-outside-init in multi_gpu_policy

chriselion:
pylint seems to ignore some files randomly, see https://github.com/PyCQA/pylint/issues/2893

It was currently ignoring the errors on multi_gpu_policy.py, but this can change if the list of files changes (as was happening on a branch).

I think these warnings aren't worth fixing for this file right now.

### DIFF
--- a/ml-agents/mlagents/trainers/ppo/multi_gpu_policy.py
+++ b/ml-agents/mlagents/trainers/ppo/multi_gpu_policy.py
@@ -48,7 +48,6 @@ class MultiGpuPPOPolicy(PPOPolicy):
         :param seed: Random seed.
         """
         self.devices = get_devices()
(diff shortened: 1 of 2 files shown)


ml-agents/mlagents/trainers/ppo/multi_gpu_policy.py
```diff
@@ -29,6 +49,7 @@ def create_model(
         """
         self.devices = get_devices()
         self.towers = []
```
ervteng: Do we still need to declare here?

chriselion: Don't need to, I'll remove

PR #2900, opened 2019-11-12
Migration - add a note about updating both python and c#

chriselion:


chriselion:
@anupambhatnagar Updated with links to the API version strings.

anupambhatnagar:
looks good. Approved.

chriselion:
FWIW this will get better in the next release with the `--version` flag that @vincentpierre added. But can't do that just yet.

PR #2902, opened 2019-11-13
Move mlagents.envs

chriselion:
Rename mlagents.envs to mlagentsenvs.envs.

The old naming scheme relied on[ PEP420](https://www.python.org/dev/peps/pep-0420/) which was causing problems with pylint (had to disable a warning) and mypy (couldn't enable --namespace-packages without hitting https://github.com/python/mypy/issues/7510, but this missed some errors).

Open to suggestions on the name. We could remove the `envs` directory so the imports are e.g. `mlagentsenvs.environment` instead of `mlagents.envs.environment`

Currently this just moves the files and updates the imports and the proto files. I'll do another pass to fix any doc changes if we're OK with the move and the name. Expecting mypy to fail until the newly uncovered errors are fixed.


harperj:
What about `mlagents_envs.environment`?

I'm also beginning to think the right thing may be to just re-combine the two packages.  Have you thought about that option?

surfnerd:
I think removing the `envs` directory makes sense here.  If there were any sibling directories, or code maybe it would make less sense.  Alas, there is nothing under the `mlagents` folder other than `envs`.

surfnerd:
does the windows protobuf generation script also need to be updated?

chriselion:
Cool, will probably redo this PR instead of fixing the merge conflicts.

I'll make sure to update the windows script, although it's probably less important now that we removed the Custom* protos.

chriselion:
Closing this for now, will revisit soon

PR #2909, opened 2019-11-14
Update Training-ML-Agents.md

njustesen:
<<<<<<< HEAD removed

PR #2910, opened 2019-11-14
Check for merge conflict strings (and fix)

chriselion:
Pointed out in https://github.com/Unity-Technologies/ml-agents/pull/2909

This also adds a pre-commit check to make sure we don't commit new ones.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,6 +49,8 @@ repos:
     -   id: trailing-whitespace
         name: trailing-whitespace-markdown
         types: [markdown]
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -49,6 +49,8 @@ repos:
     -   id: trailing-whitespace
         name: trailing-whitespace-markdown
         types: [markdown]
+    -   id: check-merge-conflict
+        args: [--assume-in-merge]
```
chriselion: Normally this only runs if it's in a `git merge` situation. That meant it missed the error until I added this flag.

PR #2912, opened 2019-11-15
Bubbled up indexing of AllBrainInfo to trainer controller from trainers

andrewcoh:
Currently, AllBrainInfo dicts are passed to trainers in process_experiences and add_experiences.  Within those functions, the AllBrainInfos are immediately indexed by brain_name for the brain info corresponding to that trainer and the remaining brain_infos are never touched.  It's unnecessary for the trainers to have this data since it's never used.  I had to modify some of the tests which call these functions to handle the changes.  All pytest tests pass.

### DIFF
--- a/ml-agents/mlagents/trainers/tests/test_bc.py
+++ b/ml-agents/mlagents/trainers/tests/test_bc.py
@@ -77,19 +77,20 @@ def test_bc_trainer_add_proc_experiences(dummy_config):
     trainer, env = create_bc_trainer(dummy_config)
     # Test add_experiences
     returned_braininfo = env.step()
+    brain_name = "Ball3DBrain"
     trainer.add_experiences(
-        returned_braininfo["Ball3DBrain"], returned_braininfo["Ball3DBrain"], {}
+        returned_braininfo[brain_name], returned_braininfo[brain_name], {}
     )  # Take action outputs is not used
-    for agent_id in returned_braininfo["Ball3DBrain"].agents:
+    for agent_id in returned_braininfo[brain_name].agents:
         assert trainer.evaluation_buffer[agent_id].last_brain_info is not None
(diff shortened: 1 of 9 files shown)


ml-agents/mlagents/trainers/tests/test_bc.py
```diff
@@ -78,15 +78,17 @@ def test_bc_trainer_add_proc_experiences(dummy_config):
     # Test add_experiences
     returned_braininfo = env.step()
     trainer.add_experiences(
-        returned_braininfo, returned_braininfo, {}
+        returned_braininfo["Ball3DBrain"], returned_braininfo["Ball3DBrain"], {}
```
chriselion: I know it's test code, so the quality bar is lower. but maybe do e.g. `brain_name = "Ball3DBrain"` and use that instead of the string everywhere?

ml-agents/mlagents/trainers/tests/test_trainer_controller.py
```diff
@@ -159,11 +159,11 @@ def trainer_controller_with_take_step_mocks():
 
 def test_take_step_adds_experiences_to_trainer_and_trains():
     tc, trainer_mock = trainer_controller_with_take_step_mocks()
-
     action_info_dict = {"testbrain": MagicMock()}
 
-    old_step_info = EnvironmentStep(Mock(), Mock(), action_info_dict)
-    new_step_info = EnvironmentStep(Mock(), Mock(), action_info_dict)
+    brain_info_dict = {"testbrain": Mock()}
```
chriselion: Same comment for the brain_name

ml-agents/mlagents/trainers/tests/test_bc.py
andrewcoh: I'll fix this now. Thanks.

PR #2916, opened 2019-11-15
Convert most other scenes to RayPerception sensor

chriselion:


UnitySDK/Assets/ML-Agents/Examples/Soccer/Scripts/AgentSoccer.cs
```diff
@@ -3,6 +3,13 @@
 
 public class AgentSoccer : Agent
 {
+    // Note that that the detectable tags are different for the blue and purple teams. The order is
```
chriselion: Not sure where else to document this. It's a bit non-intuitive if you just look at the tags in the sensor.

surfnerd: seems pretty self documenting in the editor.  Having it here only makes it more clear.

chriselion:
Need to retrain the models for these - will kick off a cloud run over the weekend.

chriselion:
Having trouble getting Pyramid cloud training to run. Backing that out for now.

PR #2948, opened 2019-11-21
check for numpy float64

chriselion:
The goal is to make sure we keep as much as possible in float32 arrays to avoid extra memory usage.

This works by monkey-patching some numpy methods. This runs by setting an environment variable, which happens on CI runs.

I want to run some performance tests on this, but the overall approach is ready for review.


chriselion:
Timing experiments were disappointing. On GridWorld and Hallway, no noticeable difference. I also modified 3DBall to use 50000 observations, and that showed about a 10% speedup.

ml-agents/mlagents/trainers/models.py
```diff
@@ -112,7 +112,11 @@ def create_learning_rate(
             learning_rate = tf.Variable(lr)
         elif lr_schedule == LearningRateSchedule.LINEAR:
             learning_rate = tf.train.polynomial_decay(
-                lr, global_step, max_step, 1e-10, power=1.0
+                np.float32(lr),
```
chriselion: I should change this back - I was originally trying to prevent tensorflow from creating float64 arrays internally, but gave up on that fight.

PR #2952, opened 2019-11-22
Develop barracuda 0.3.x

mantasp:
Switching from Barracuda as drop-in library to Barracuda as a package.
Note: Protobuf got removed as it is coming as part of Barracuda package.


surfnerd:
sweet

PR #3013, opened 2019-12-02
[timer] Add gauges to csharp timers to reflect python gauges.

surfnerd:


UnitySDK/Assets/ML-Agents/Scripts/Timer.cs
```diff
@@ -250,6 +294,7 @@ public void Reset(string name = "root")
         {
             m_Stack = new Stack<TimerNode>();
             m_RootNode = new TimerNode(name, true);
+            m_RootNode.Gauges = new Dictionary<string, GaugeNode>();
```
chriselion: nit: maybe do this in TimerNode.TimerNode if `isRoot` is true? Could also create lazily or store in TimerStack instead, if you're worried about memory.

UnitySDK/Assets/ML-Agents/Scripts/Timer.cs
```diff
@@ -75,6 +82,12 @@ public double TotalSeconds
             set {}  // Serialization needs this, but unused.
         }
 
+        public Dictionary<string, GaugeNode> Gauges
+        {
+            get { return m_Gauges; }
+            set { m_Gauges = value; }
```
chriselion: I left the setter empty in some of the other properties; unfortunately serialization seems to need them, but we don't really want them settable, and Rider complains about the unused `value` parameter if they're unimplemented.

I don't really care either way, but we probably should be consistent.

surfnerd: took your suggestion.

UnitySDK/Assets/ML-Agents/Scripts/Timer.cs
surfnerd: took your suggestion

PR #3018, opened 2019-12-03
Add note about installing Barracuda

chriselion:


ervteng:
Should we also add it to the Basic guide? https://github.com/Unity-Technologies/ml-agents/blob/master/docs/Basic-Guide.md

chriselion:
@ervteng good call, will do

chriselion:
Actually, let's talk about that tomorrow - @surfnerd and I need to figure out if we're specifying dependencies wrong.

PR #3034, opened 2019-12-05
handle mismatch between brain and metacurriculum

chriselion:
* Having a brain that doesn't have a curriculum results in a KeyError in trainer_util.py (amongst other places). This is _probably_ a setup error, but we should at least be more forgiving.
* Having a curriculum that doesn't have a brain results in KeyErrors in trainer_controller.py (amongst other places). Seems like this could happen pretty easily for a stale config, or a brain that hasn't connected yet.

Tested manually by renaming a WallJump curriculum file, also added unit tests.


ml-agents/mlagents/trainers/meta_curriculum.py
```diff
@@ -32,9 +32,9 @@ def __init__(self, curriculum_folder: str):
         try:
             for curriculum_filename in os.listdir(curriculum_folder):
                 # This process requires JSON files
-                if not curriculum_filename.lower().endswith(".json"):
+                brain_name, extension = os.path.splitext(curriculum_filename)
```
chriselion: So far I don't think this has tripped anybody up. But the old behavior would mean that a brain with a `.` in the name wouldn't be able to match up with the metacurriculum, because `curriculum_filename.split(".")[0]` would take the only the filename up to the first `.` instead of the last one.

PR #3043, opened 2019-12-06
Close environment if step raises an exception.

chriselion:
In https://github.com/Unity-Technologies/ml-agents/pull/2988, that was raising with this callstack:
```
  File "/Users/chris.elion/code/ml-agents/venv/bin/mlagents-learn", line 11, in <module>
    load_entry_point('mlagents', 'console_scripts', 'mlagents-learn')()
  File "/Users/chris.elion/code/ml-agents/ml-agents/mlagents/trainers/learn.py", line 422, in main
    run_training(0, run_seed, options, Queue())
  File "/Users/chris.elion/code/ml-agents/ml-agents/mlagents/trainers/learn.py", line 266, in run_training
    tc.start_learning(env)
  File "/Users/chris.elion/code/ml-agents/ml-agents/mlagents/trainers/trainer_controller.py", line 209, in start_learning
    n_steps = self.advance(env_manager)
  File "/Users/chris.elion/code/ml-agents/ml-agents-envs/mlagents/envs/timers.py", line 262, in wrapped
    return func(*args, **kwargs)
  File "/Users/chris.elion/code/ml-agents/ml-agents/mlagents/trainers/trainer_controller.py", line 283, in advance
    step_info.brain_name_to_action_info[brain_name].outputs,
  File "/Users/chris.elion/code/ml-agents/ml-agents/mlagents/trainers/rl_trainer.py", line 131, in add_experiences
    curr_to_use, take_action_outputs["action"], next_info
TypeError: 'NoneType' object is not subscriptable
```
The actual error isn't that important, but the thing to note is that the worker process stuck was waiting  for the main process :
https://github.com/Unity-Technologies/ml-agents/blob/2a0bb710672e978dd35bc2689958edb8291c8510/ml-agents-envs/mlagents/envs/subprocess_env_manager.py#L104
(I think, may have been somewhere different).
This means that the process never exited and basically deadlocked.

This change tries to call EnvManager.close() even if training raised an exception. I confirmed by undoing the fix from #2988 that this at least lets python exit and Unity stops playing (instead of having a spinning beach ball).

### DIFF
--- a/ml-agents/mlagents/trainers/tests/test_trainer_controller.py
+++ b/ml-agents/mlagents/trainers/tests/test_trainer_controller.py
@@ -120,7 +120,6 @@ def test_start_learning_trains_forever_if_no_train_model(tf_reset_graph):
     assert tc.advance.call_count == 11
     tc._export_graph.assert_not_called()
     tc._save_model.assert_not_called()
-    env_mock.close.assert_called_once()
 
 
 @patch.object(tf, "reset_default_graph")
@@ -138,7 +137,6 @@ def test_start_learning_trains_until_max_steps_then_saves(tf_reset_graph):
     tf_reset_graph.assert_called_once()
     env_mock.reset.assert_called_once()
     assert tc.advance.call_count == trainer_mock.get_max_steps + 1
-    env_mock.close.assert_called_once()
     tc._save_model.assert_called_once()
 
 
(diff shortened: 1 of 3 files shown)


ml-agents/mlagents/trainers/tests/test_trainer_controller.py
```diff
@@ -120,7 +120,6 @@ def test_start_learning_trains_forever_if_no_train_model(tf_reset_graph):
     assert tc.advance.call_count == 11
     tc._export_graph.assert_not_called()
     tc._save_model.assert_not_called()
-    env_mock.close.assert_called_once()
```
chriselion: close() isn't being called here anymore. I think it's cleaner to have it in the trainer_controller since that's what creates it in the first place.

PR #3048, opened 2019-12-07
better logging for ports and versions

chriselion:
* Clean up (IMHO) the port decision logic on C# side
* Always print the version information on python
* Print the port number when waiting for editor connection
* If editor doesn't connect, print the port and API version.

### DIFF
--- a/UnitySDK/Assets/ML-Agents/Scripts/Academy.cs
+++ b/UnitySDK/Assets/ML-Agents/Scripts/Academy.cs
@@ -42,6 +42,7 @@ namespace MLAgents
     public abstract class Academy : MonoBehaviour
     {
         const string k_ApiVersion = "API-12";
+        const int k_EditorTrainingPort = 5004;
 
         /// Temporary storage for global gravity value
         /// Used to restore oringal value when deriving Academy modifies it
@@ -149,7 +150,7 @@ namespace MLAgents
         }
 
         // Used to read Python-provided environment parameters
-        static int ReadArgs()
+        static int ReadPortFromArgs()
         {
             var args = System.Environment.GetCommandLineArgs();
             var inputPort = "";
@@ -161,7 +162,22 @@ namespace MLAgents
                 }
             }
 
-            return int.Parse(inputPort);
+            try
+            {
+                return int.Parse(inputPort);
+            }
+            catch
+            {
+                // No arg passed, or malformed port number.
+#if UNITY_EDITOR
+                // Try connecting on the default editor port
+                return k_EditorTrainingPort;
+#else
+                // This is an executable, so we don't try to connect.
+                return -1;
+#endif
+            }
+
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


chriselion:
One thing I didn't realize earlier is that we lost some of the exception information coming from here
https://github.com/Unity-Technologies/ml-agents/blob/22520fe4724c7c592afcc37069e723036b5521f5/UnitySDK/Assets/ML-Agents/Scripts/Grpc/RpcCommunicator.cs#L92-L103
because this is the expected path when doing inference. Not the end of the world though.

UnitySDK/Assets/ML-Agents/Scripts/Academy.cs
```diff
@@ -217,6 +225,10 @@ void InitializeEnvironment()
                 }
                 catch
                 {
+                    Debug.Log($"" +
+                        $"Couldn't connect to trainer on port {port} using API version {k_ApiVersion}. " +
```
chriselion: Hope this is OK, I want some information to make sure we're using the same port as python, but also want to make it clear that this is expected sometimes (when doing inference).

UnitySDK/Assets/ML-Agents/Scripts/Academy.cs
```diff
@@ -179,23 +195,15 @@ void InitializeEnvironment()
...
 
             // Try to launch the communicator by using the arguments passed at launch
-            try
+            var port = ReadPortFromArgs();
+            if (port > 0)
             {
                 Communicator = new RpcCommunicator(
                     new CommunicatorInitParameters
                     {
-                        port = ReadArgs()
```
chriselion: The try/catch here was really for the ReadArgs(); RpcCommunicator never throws.

ml-agents/mlagents/trainers/learn.py
```diff
@@ -59,12 +59,11 @@ def from_argparse(args: Any) -> "CommandLineOptions":
...
 
 def get_version_string() -> str:
-    return f""" Version information:\n
-    ml-agents: {mlagents.trainers.__version__},
-    ml-agents-envs: {mlagents.envs.__version__},
-    Communicator API: {UnityEnvironment.API_VERSION},
-    TensorFlow: {tf_utils.tf.__version__}
-"""
+    return f""" Version information:
+  ml-agents: {mlagents.trainers.__version__},
```
chriselion: Just trimmed up the whitespace here.

ml-agents/mlagents/trainers/learn.py
```diff
@@ -170,7 +169,7 @@ def parse_command_line(argv: Optional[List[str]] = None) -> CommandLineOptions:
         "--cpu", default=False, action="store_true", help="Run with CPU only"
     )
 
-    parser.add_argument("--version", action="version", version=get_version_string())
+    parser.add_argument("--version", action="version", version="")
```
chriselion: `--version` still works, but we always print the version anyway. I think it's worth keeping since that's a standard argument for a CLI.

PR #3055, opened 2019-12-09
Updated import format for example with openai gym

albertoxamin:


chriselion:
Thanks @albertoxamin. We need some better checks to make sure these docs stay up-to-date; hopefully we'll get to that soon.

PR #3059, opened 2019-12-09
Update Installation.md

unityjeffrey:


chriselion:
This is already linked from https://github.com/Unity-Technologies/ml-agents/tree/master/docs#deprecated-docs - do we really need to add another place where we say "we don't really support this but..."?

unityjeffrey:
i believe so.  many of the first time users walk through the installation doc as a first step.  Since its unlikely they will see the depricated docs sections, its important we call it out there as we don't want to an abandonment of the funnel

unityjeffrey:
i made this change as I was walking some folks at Neurips through the workflow and couldn't find it.

docs/Installation.md
```diff
@@ -21,6 +21,9 @@ We now support a single mechanism for installing ML-Agents on Mac/Windows/Linux
 Environments. For more information on Virtual Environments and installation instructions,
 follow this [guide](Using-Virtual-Environment.md).
 
+Although we don't support Anaconda installation path of ML-Agents for Windows, the previous guide
+is still in the docs folder.  Please refer to [Windows Installation (Depricated)](Installation-Windows.md).
```
chriselion:
```suggestion
is still in the docs folder.  Please refer to [Windows Installation (Deprecated)](Installation-Windows.md).
```

PR #3065, opened 2019-12-10
Disable blank issues

harperj:
The goal of our issue templates was that all users should choose one
of these templates and fill it out to begin a discussion or report a
bug / feature request.  The "blank issues" option is a way around this,
and this change disables it as an option.


chriselion:
Let's make an Installation template too (can be separate pr)

PR #3076, opened 2019-12-12
Add stats reporter class and re-enable missing stats

ervteng:
This PR adds back in certain statistics missing from the AgentProcessor PR (entropy, learning rate) that were only known to the AgentProcessor and not the Trainer. 

We do this by creating a global class called a StatsReporter, that takes in a category, key, and float value. This StatsReporter can then write the mean of these values out on command. Currently this is still handled by the Trainer. 

The StatsReporter also keeps a list of Writer classes - currently, we only have a Tensorboard writer but we can imagine adding more in the future (e.g. REST API writer, CSV writer).  

Note: Why is this a PR to the AgentProcessor PR and not to Master? The AgentProcessor marks the first time we have multiple sources for stats, and thus requires that certain stats (related to Policy inference, e.g. reward, entropy, episode steps) come from a different place than others (related to Policy training, e.g. loss).


ml-agents/mlagents/trainers/agent_processor.py
```diff
@@ -16,27 +17,35 @@ class AgentProcessor:
...
             ActionInfoOutputs
         )
-        self.stats: Dict[str, List[float]] = defaultdict(list)
         # Note: this is needed until we switch to AgentExperiences as the data input type.
         # We still need some info from the policy (memories, previous actions)
         # that really should be gathered by the env-manager.
         self.policy = policy
-        self.episode_steps: Dict[str, int] = {}
+        self.episode_steps: Counter = Counter()
+        self.episode_rewards: Dict[str, float] = defaultdict(lambda: 0.0)
```
chriselion: nit: `defaultdict(float)` is more common I think.

ml-agents/mlagents/trainers/agent_processor.py
```diff
@@ -51,12 +60,16 @@ def add_experiences(
         :param take_action_outputs: The outputs of the Policy's get_action method.
         """
         if take_action_outputs:
-            self.stats["Policy/Entropy"].append(take_action_outputs["entropy"].mean())
-            self.stats["Policy/Learning Rate"].append(
-                take_action_outputs["learning_rate"]
+            stats.stats_reporter.add_stat(
+                self.stats_category,
+                "Policy/Entropy",
```
chriselion: Is it worth making an enum for these stat keys?

chriselion: Eh, I guess not because of reward signals.

ml-agents/mlagents/trainers/stats.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+        Write out some text.
+        :param category: The highest categorization of the statistic, e.g. behavior name.
+        :param text: The text to write out.
+        :param step: Training step which to write these stats as.
+        """
+        for writer in self.writers:
+            writer.write_text(category, text, step)
+
+    def get_mean_stat(self, category: str, key: str) -> float:
```
chriselion: What do you think about combining get_mean_stat, get_std_stat, and get_num_stats into something like `get_summary_stats()` that returns a NamedTuple with count, mean, and stddev. I think that would clean up the usage in write_summary() a bit.

ervteng: Done

ml-agents/mlagents/trainers/agent_processor.py
ervteng: Changed

ml-agents/mlagents/trainers/stats.py
```diff
@@ -0,0 +1,109 @@
...
+class StatsSummary(NamedTuple):
+    mean: float
+    std: float
+    num: int
+
+
+class StatsReporter:
+    def __init__(self, writers: List[StatsWriter]):
+        """
+        Generic StatsReporter. A category is the broadest type of storage (would
```
harperj: Rather than having a global StatsReporter and passing in `category` for every `add_stat`, why not have a StatsReporter for every category?

ml-agents/mlagents/trainers/ppo/trainer.py
```diff
@@ -99,7 +100,11 @@ def process_trajectory(self, trajectory: Trajectory) -> None:
         )
         for name, v in value_estimates.items():
             agent_buffer_trajectory["{}_value_estimates".format(name)].extend(v)
-            self.stats[self.policy.reward_signals[name].value_name].append(np.mean(v))
+            stats.stats_reporter.add_stat(
+                self.summary_path,
+                self.policy.reward_signals[name].value_name,
+                np.mean(v),
```
chriselion: Is v always non-empty? Do you need to guard against NaNs anywhere?

chriselion: Guess it's same behavior as before...

ml-agents/mlagents/trainers/rl_trainer.py
```diff
@@ -46,23 +47,19 @@ def end_episode(self) -> None:
...
-                self.stats["Environment/Cumulative Reward"].append(
-                    rewards.get(agent_id, 0)
-                )
                 self.reward_buffer.appendleft(rewards.get(agent_id, 0))
                 rewards[agent_id] = 0
             else:
-                self.stats[self.policy.reward_signals[name].stat_name].append(
-                    rewards.get(agent_id, 0)
+                stats.stats_reporter.add_stat(
+                    self.summary_path,
```
harperj: Feels a little weird that the "category" here is a filepath.  The "category" seems like something that should just be the filename / behavior name, whereas the base bath should be something used to configure the Tensorboard writer.

ml-agents/mlagents/trainers/ppo/trainer.py
ervteng: v shouldn't be NaN unless there are NaNs in the network (dun dun dun)

ml-agents/mlagents/trainers/stats.py
ervteng: That's possible, but there still needs to be a way to access commonly used StatsReporters, e.g. the AgentProcessor and the Trainer will record in the same category. What about if we had the global class store a Dict of StatsReporters, and each module would grab one and use it?

ml-agents/mlagents/trainers/rl_trainer.py
ervteng: Seems to me like the summary_path is exactly that - just the name of the behavior and not the full path

ml-agents/mlagents/trainers/stats.py
harperj: How about making `StatsReporter.writers` and `StatsReporter.stats_dict` static members, and accepting a "category" as the only argument of the constructor?

ml-agents/mlagents/trainers/stats.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        Write out all stored statistics that fall under the category specified.
+        The currently stored values will be averaged, written out as a single value,
+        and the buffer cleared.
+        :param category: The category which to write out the stats.
+        :param step: Training step which to write these stats as.
+        """
+        for key in self.stats_dict[category]:
+            if len(self.stats_dict[category][key]) > 0:
+                stat_mean = float(np.mean(self.stats_dict[category][key]))
```
harperj: I don't love that we won't have any method for logging min/max values via this interface.  Not sure I have a great solution for this at the moment, though.

ml-agents/mlagents/trainers/stats.py
```diff
@@ -0,0 +1,109 @@
+from collections import defaultdict
```
harperj: Seems like while you have updated tests to reflect the usage of the classes in this module, we aren't explicitly testing the functionality captured here.  Could you add some tests (you should just be able to mock out the `tf.summary.FileWriter`)?

ervteng: Added test

ml-agents/mlagents/trainers/stats.py
ervteng: Hmm, that's a good idea, I think it might be worth adding. 

I'm inclined to get the StatsReporter interface installed into the code and then work on what goes behind it in future PRs - currently it's blocking a larger trainer refactor that's in turn blocking another trainer refactor :P

ml-agents/mlagents/trainers/agent_processor.py
```diff
@@ -16,31 +17,39 @@ class AgentProcessor:
     One AgentProcessor should be created per agent group.
     """
 
-    def __init__(self, trainer: Trainer, policy: TFPolicy, max_trajectory_length: int):
+    def __init__(
+        self,
+        trainer: Trainer,
+        policy: TFPolicy,
+        max_trajectory_length: int,
+        stats_category: str,
```
harperj: If this is coming from the trainer, why not pass a `StatsReporter`?

ml-agents/mlagents/trainers/trainer.py
```diff
@@ -71,6 +69,29 @@ def check_param_keys(self):
...
                 )
 
+    def write_tensorboard_text(self, key: str, input_dict: Dict[str, Any]) -> None:
+        """
+        Saves text to Tensorboard.
+        Note: Only works on tensorflow r1.2 or above.
+        :param key: The name of the text.
+        :param input_dict: A dictionary that will be displayed in a table on Tensorboard.
+        """
+        #        try:
```
harperj: Remove commented out code if it won't be used.

ervteng: Good catch. This shouldn't be here at all.

ml-agents/mlagents/trainers/agent_processor.py
ervteng: Done

PR #3077, opened 2019-12-12
WIP Develop magic string

andrewcoh:
This is the PR for spawning multiple brain names as outlined in this design [doc.](https://docs.google.com/document/d/1qg0CSVWljfAprhE8sk2XFiydvmFXfRMtJRr1jRJhw48/edit#) As is, it passes all pytest/C# tests and trains 3DBall.  I'm calling it a WIP because I think there are bits that I can clean up but I'd really appreciate your feedback. 

The major changes:
- behavior name is now the root behavior name + team id (team id field added to the editor)
- python side has some parsing in trainer controller to handle name_behavior_id
- policies are created in a separate function within the trainer
- getter/setter for policies by name_behavior_id
-modified tests for new trainer functions

The minor changes:
- trainer doesn't take brain_parameters as an argument, just brain name.

This is the starting point for the ghost trainer.


ml-agents/mlagents/trainers/ppo/trainer.py
```diff
@@ -277,6 +270,29 @@ def update_policy(self):
...
+        else:
+            self.ppo_policy = PPOPolicy(
+                self.seed,
+                brain_parameters,
+                self.trainer_parameters,
+                self.is_training,
+                self.load,
+            )
+
+        for _reward_signal in self.ppo_policy.reward_signals.keys():
```
chriselion:
```suggestion
        self.policy = self.ppo_policy
        for _reward_signal in self.ppo_policy.reward_signals.keys():
```

chriselion: Never mind, handled in rltrainer

ml-agents/mlagents/trainers/trainer.py
```diff
@@ -283,5 +287,17 @@ def update_policy(self):
         """
         raise UnityTrainerException("The update_model method was not implemented.")
 
+    def create_policy(self, brain_parameters: BrainParameters) -> TFPolicy:
+        """
+        Creates policy
+        """
+        raise UnityTrainerException("The update_model method was not implemented.")
```
chriselion:
```suggestion
        raise UnityTrainerException("The create_policy method was not implemented.")
```

ml-agents/mlagents/trainers/trainer.py
```diff
@@ -60,7 +60,8 @@ def __init__(
         )
         self.summary_writer = tf.summary.FileWriter(self.summary_path)
         self._reward_buffer: Deque[float] = deque(maxlen=reward_buff_cap)
-        self.policy: TFPolicy = None  # type: ignore  # this will always get set
+        self.policy: TFPolicy
```
chriselion: This looks really awkward. What's the difference between self.policy and self.policies?

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -49,6 +50,8 @@ def __init__(
         :param resampling_interval: Specifies number of simulation steps after which reset parameters are resampled.
         """
         self.trainers: Dict[str, Trainer] = {}
+        self.multi_trainers: Dict[str, Trainer] = {}
```
ervteng: Do some trainers go in `trainers` and some in `multi_trainers`?

ml-agents/mlagents/trainers/trainer_util.py
```diff
@@ -37,10 +36,10 @@ def __init__(
         self.meta_curriculum = meta_curriculum
         self.multi_gpu = multi_gpu
 
-    def generate(self, brain_parameters: BrainParameters) -> Trainer:
+    def generate(self, brain_name: str) -> Trainer:
         return initialize_trainer(
             self.trainer_config,
-            brain_parameters,
+            brain_name,
```
ervteng: Do we even need the brain_name here? Isn't it passed when the policy is created?

ml-agents/mlagents/trainers/trainer_controller.py
andrewcoh: This actually doesn't get used.  All trainers go into `self.trainers` corresponding to root behavior name.  Then, I use a separate `dict[str, set]` which maps root behavior name to `name behavior id`. Removing this.

ml-agents/mlagents/trainers/trainer_util.py
andrewcoh: Brain parameters are. The brain name there is used to grab config params etc.

ml-agents/mlagents/trainers/rl_trainer.py
```diff
@@ -265,6 +269,14 @@ def add_rewards_outputs(
             "The add_rewards_outputs method was not implemented."
         )
 
+    def add_policy(self, brain_parameters: BrainParameters) -> None:
+        """
+        Adds policy to trainers list of policies
+        """
+        policy = self.create_policy(brain_parameters)
+        self.policy = policy
```
chriselion: This overwrites the old policy?

ml-agents/mlagents/trainers/trainer.py
andrewcoh: self.policies is for when one trainer needs to track multiple policies (ghost trainer). But you are right, PPO/SAC never need to do this so I should bubble up this field just for multiagent/self-play algorithms.  Essentially, the get_policy and add_policy methods will access self.policies in multiagent algorithms, but ppo and sac don't ever need to. Going to fix this.

ml-agents/mlagents/trainers/rl_trainer.py
andrewcoh: In ppo/sac this should only be called once.  In the ghost trainer this will be overridden to add to self.policies.  We'll still need to set self.policy in the wrapped trainer though.

ml-agents/mlagents/trainers/trainer.py
chriselion: Sounds good.

chriselion: Should we separate Trainer into SinglePolicyTrainer and MultiPolicyTrainer? Seems like they have fundamentally different interfaces.

One thing we could do is make `SinglePolicyTrainer.policy` an abstract property instead of a field. Then SACTrainer can return self.sac_policy etc.

ml-agents/mlagents/trainers/trainer_util.py
ervteng: I might be crazy but I don't see `brain_name` being used in the Trainer

ml-agents/mlagents/trainers/trainer.py
ervteng: IMO a single policy trainer is just a special case of a multi policy trainer, with only one entry in the dict. But not sure if this PR goes that far.

ml-agents/mlagents/trainers/trainer_util.py
andrewcoh: The logger uses it to generate a path/label and __str__ uses it.  We can probably cut it out though.

andrewcoh: It gets used by the logger in Trainer as a label/path and also in the __str__ method

ervteng: Oh IC. OK makes sense

ml-agents/mlagents/trainers/rl_trainer.py
chriselion: If it's only meant to be called once, is that worth enforcing with an exception if self.policy is already set?

ml-agents/mlagents/trainers/rl_trainer.py
```diff
@@ -265,6 +269,14 @@ def add_rewards_outputs(
...
+        # self.policies[brain_parameters.brain_name] = policy
```
chriselion: nit: dead code

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -185,30 +187,42 @@ def write_to_tensorboard(self, global_step: int) -> None:
...
-                        trainer = self.trainer_factory.generate(
-                            env_manager.external_brains[name]
-                        )
-                        self.start_trainer(trainer, env_manager)
-                    last_brain_names = external_brains
+                external_brain_behavior_ids = set(env_manager.external_brains.keys())
+                new_behavior_ids = external_brain_behavior_ids - last_brain_behavior_ids
+                for name_behavior_id in new_behavior_ids:
+                    try:
+                        brain_name, _ = name_behavior_id.split("?")
```
chriselion: OK for now, but if we ever parse the name_behavior_id in more than one place, let's move it to a utility function.

chriselion:
(summarizing offline discussion)
Subsequent PR will remove `policy` from `Trainer`, and move these to SACTrainer and PPOTrainer respectively (this will get rid of the need for their `sac_policy` and `ppo_policy`). Where RLTrainer uses `self.policy`, it will swap to using get_policy() with the behavior_name_id.

PR #3083, opened 2019-12-12
Rename mlagents.envs to mlagents_envs

chriselion:
https://github.com/Unity-Technologies/ml-agents/pull/2902, take two. Motivation from the original PR:

> The old naming scheme relied on[ PEP420](https://www.python.org/dev/peps/pep-0420/) which was causing problems with pylint (had to disable a warning) and mypy (couldn't enable --namespace-packages without hitting https://github.com/python/mypy/issues/7510, but this missed some errors).

I've since enabled mypy --namespace-packages and fixed the errors.

~One thing I need to check is that the we can still build packages with this.~ ✅ 

~Will also drop a note in the migration guide.~ ✅


ml-agents-envs/setup.py
```diff
@@ -41,7 +41,10 @@ def run(self):
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
     ],
-    packages=["mlagents.envs", "mlagents.envs.communicator_objects"],  # Required
+    packages=[
+        "mlagents_envs",
+        "mlagents_envs.communicator_objects",
+    ],  # Required  # TODO SIDE CHANNELS
```
chriselion: This was a bug waiting to happen - side_channels directory wasn't in the built packages. I'm switching it to automatically find the directories in another PR.

PR #3084, opened 2019-12-12
Develop mla 426 test pypi

chriselion:
DO NOT MERGE


chriselion:
Initial packages didn't contain the side_channel directory.

After switching to find_packages(), the installed packages from https://test.pypi.org/project/mlagents/0.12.1.dev1/ worked for me.

PR #3094, opened 2019-12-16
Run precommit in its own job, cache the data

chriselion:
* Move pre-commit to it's own job. Currently it runs on all 3 python testing jobs.
* Cache the pre-commit cache (and venv). Timings below.

Should we also make the python unit test jobs dependent on precommit job?

Resolves:
https://jira.unity3d.com/browse/MLA-17
https://jira.unity3d.com/browse/MLA-394

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -56,6 +56,7 @@ repos:
     -   id: check-merge-conflict
         args: [--assume-in-merge]
     -   id: check-yaml
+        # Won't handle the templating in yamato
         exclude: \.yamato/*
 
 -   repo: https://github.com/pre-commit/pygrep-hooks
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -63,8 +63,6 @@ repos:
     hooks:
     -   id: python-check-mock-methods
 
-
-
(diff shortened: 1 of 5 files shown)


chriselion:
Cache miss:
image

Cache hit (precommit only)
image

Cache hit (precommit + venv)
image

Total time for the run goes from 1:44 uncached to 0:42 cached.

.pre-commit-config.yaml
```diff
@@ -55,14 +55,14 @@ repos:
         types: [markdown]
     -   id: check-merge-conflict
         args: [--assume-in-merge]
+    -   id: check-yaml
```
chriselion: I can back this out, but I it saved me at least one bad `config.yml` push.

.pre-commit-config.yaml
```diff
@@ -55,14 +55,14 @@ repos:
...
+        exclude: \.yamato/*
```
surfnerd: were the Yamato configs failing?

chriselion: (discussed offline, templating breaks the checks)

chriselion: (discussed offline, templating breaks the checks)

PR #3095, opened 2019-12-16
Enable child GameObjects to have a SensorComponent as well

ervteng:
In many cases, we want to add Sensors to child GameObjects of the Agent - this could be when the Agent is made up of many GameObjects, or if we want to apply a transform to e.g. a RayPerception component. 

However, in the case of hierarchical agents, we don't necessarily want this. This PR makes whether or not we search children for Sensors optional.

### DIFF
--- a/UnitySDK/Assets/ML-Agents/Editor/BehaviorParametersEditor.cs
+++ b/UnitySDK/Assets/ML-Agents/Editor/BehaviorParametersEditor.cs
@@ -33,6 +33,7 @@ namespace MLAgents
             EditorGUILayout.PropertyField(so.FindProperty("m_InferenceDevice"), true);
             EditorGUI.indentLevel--;
             EditorGUILayout.PropertyField(so.FindProperty("m_BehaviorType"));
+            EditorGUILayout.PropertyField(so.FindProperty("m_useChildSensors"), true);
             // EditorGUILayout.PropertyField(serializedObject.FindProperty("m_Heuristic"), true);
             EditorGUI.indentLevel--;
             if (EditorGUI.EndChangeCheck())
@@ -58,7 +59,15 @@ namespace MLAgents
             Model barracudaModel = null;
             var model = (NNModel)serializedObject.FindProperty("m_Model").objectReferenceValue;
             var behaviorParameters = (BehaviorParameters)target;
-            var sensorComponents = behaviorParameters.GetComponents<SensorComponent>();
+            SensorComponent[] sensorComponents;
+            if(behaviorParameters.useChildSensors)
+            {
+                sensorComponents = behaviorParameters.GetComponentsInChildren<SensorComponent>();
+            }
+            else
+            {
+                sensorComponents = behaviorParameters.GetComponents<SensorComponent>();
+            }
             var brainParameters = behaviorParameters.brainParameters;
             if (model != null)
             {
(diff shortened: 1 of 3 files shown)


Blueeyesjt1:
I copied and pasted the modified files into the code in my ml-agents folder, but it still **doesn't seem to work** for me. Is the debug ray visual still working? (Awfully new to github) Thanks.

ervteng:
Hi @Blueeyesjt1, are you using the UnitySDK folder from the `master` branch of ML-Agents? What's the hierarchy of GameObjects under your Agent look like?

Blueeyesjt1:
image

I'm (now) using the master branch version, but it sadly seemed to have not helped my case.

@ervteng

ervteng:
Interesting - and you have "UseChildSensors" checked? I've been able to add such a sensor to an empty child gameobject and it does work. What's the error you're getting?

Blueeyesjt1:
@ervteng And this is where the issue probably is: What do you mean by "is 'UseChildSensors' checked?" Where would I check that?

Blueeyesjt1:
@ervteng Nevermind, I was making the mistake of updating the MLAgents folder outside of unity. I was fixing the wrong set of scripts! Silly mistake. It works now. Thank you for your time and help!

ervteng:
Awesome! Glad you figured it out.

Blueeyesjt1:
image
image
image

(I deleted my other question because it was unrelated to the topic.)

But now, I'm having an issue when try to test my agent AFTER training and adding the model to the inspector. Adding more observations does not fix this. What should I do? If I add an extra 33, then the error also adds another 33. It keeps asking for +33 over and over again the more I add on!

ervteng:
Hi @Blueeyesjt1, it's a little confusing but you shouldn't add the RayPerceptionSensor obs to the VectorObservationSpaceSize. 

When you quit Python training, the terminal should print out something like this:
```
Converting ./models/booger-0/PushBlock/frozen_graph_def.pb to ./models/booger-0/PushBlock.nn
GLOBALS: 'is_continuous_control', 'version_number', 'memory_size', 'action_output_shape'
IN: 'vector_observation': [-1, 1, 1, 224] => 'main_graph_0/hidden_0/BiasAdd'
IN: 'action_masks': [-1, 1, 1, 7] => 'strided_slice_1'
OUT: 'action_probs', 'concat_2', 'action'
DONE: wrote ./models/booger-0/PushBlock.nn file.
INFO:mlagents.trainers:Exported ./models/booger-0/PushBlock.nn file
```

Can you post your output? It'll help a lot in debugging. Thanks!

Blueeyesjt1:
image

ervteng:
This looks correct - are you absolutely sure that you're using the right model file?

Blueeyesjt1:
@ervteng Yes, I do. Now, I have the rayPerceptionSensorComp3D set as an observation. Is that a bad idea? I have heard it works on its own.

ervteng:
Yes it does. If you're using only RayPerceptionSensor, the VectorObservations parameter in the agent should be 0. What're the other 295 vector obs?

Blueeyesjt1:
I've now removed the ray observer. I'm gonna see if that works. Besides that, basic gameObject arrays keeping traction of physics things like positions and rotations. 
image

Blueeyesjt1:
Update: Removing the raySensor3D component from the observation list fixed it. Thanks for the help. (Again!)

PR #3097, opened 2019-12-17
Update tests to support pytest 5.x

harperj:
Our tests were using pytest fixtures by actually calling the fixture
methods, but in newer 5.x versions of pytest this causes test failures.
The recommended method for using fixtures is dependency injection.

This change updates the relevant test fixtures to either not use
`pytest.fixture` or to use dependency injection to pass the fixture.
The version range requirements in `test_requirements.txt` were also
updated accordingly.

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -1,3 +1,3 @@
 # Test-only dependencies should go here, not in setup.py
-pytest>=3.2.2,<4.0.0
+pytest>=3.2.2,<6.0.0
 pytest-cov==2.6.1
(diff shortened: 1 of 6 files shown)


test_requirements.txt
```diff
@@ -1,3 +1,3 @@
 # Test-only dependencies should go here, not in setup.py
-pytest>=3.2.2,<4.0.0
```
chriselion: Should we increase the min version too?

chriselion: 4.0.0 was out last year, seems like a good min version

harperj: Sure thing, will fix

PR #3101, opened 2019-12-18
Support for ONNX export

chriselion:
Adds support for exporting to ONNX format. This will be disabled by default, and enabled if the `tf2onnx` package is installed. A recent version (>=1.5.5) is recommended. Note that tf2onnx is not currently compatible with tensorflow 2.0.0 or higher.

ONNX files generated from ML-Agents can only be imported with Barracuda v0.6.0 or later.

Other notable changes:
* the min supported version of numpy has increased to 1.14.1


UnitySDK/Assets/ML-Agents/Scripts/InferenceBrain/BarracudaModelParamLoader.cs
```diff
@@ -138,46 +156,71 @@ public static IEnumerable<string> CheckModel(Model model, BrainParameters brainP
...
-            }
 
-            failedModelChecks.AddRange(
-                CheckIntScalarPresenceHelper(new Dictionary<string, int>()
+//                var modelApiVersion = (int)model.GetTensorByName(TensorNames.VersionNumber)[0];
+//                var memorySize = (int)model.GetTensorByName(TensorNames.MemorySize)[0];
+//                var isContinuousInt = (int)model.GetTensorByName(TensorNames.IsContinuousControl)[0];
+//                var isContinuous = GetActionType(isContinuousInt);
+//                var actionSize = (int)model.GetTensorByName(TensorNames.ActionOutputShape)[0];
+                var modelApiVersion = 2;
```
chriselion: Hacked this in for now to see what else broke in the importing. Currently the desired magic strings (i.e. TensorNames.VectorObservationPlacholder) are also not found (because the model has them suffixed with `:0`

ml-agents/mlagents/trainers/tf_policy.py
```diff
@@ -290,13 +326,158 @@ def export_model(self):
...
+            tensor_value = TensorProto(
+                data_type=TensorProto.INT32, name=name, int32_data=[value], dims=[1, 1, 1, 1]
+            )
+            value_attribute = AttributeProto(
+                name="value", t=tensor_value, type=AttributeProto.TENSOR
+            )
+            return NodeProto(
+                output=f"{name}:0",
+                name=name,
+                op_type="Constant",
```
chriselion: Though this was correct based on https://github.cds.internal.unity3d.com/unity/UnityInferenceEngine/blob/0378a31b135bdb1604b59d582a54f8c67869124c/UnityProject/Assets/Barracuda/Plugins/Editor/BarracudaEditor/ONNXModelImporter.cs#L58 but not seeing anything in the import

chriselion:
https://github.com/onnx/tensorflow-onnx/issues/760 for converting RNNs

chriselion:
I cancelled the yamato tests. Will add changelog entries after the release branch is merged to master.

ml-agents/mlagents/serialization.py
```diff
@@ -0,0 +1,205 @@
+import logging
```
surfnerd: would it make sense to call this file `onnx_serialization.py`?

chriselion: It also handles the tensorflow_to_barracuda conversion (those methods were migrated out of `tf_policy.py`). Maybe`model_serialization.py` instead?

surfnerd: eh, guess not. looks like a lot more is going on here.

surfnerd: i like `model_serialization.py`

ervteng:
This doesn't yet do ONNX -> Barracuda right? The export part LGTM

ml-agents/mlagents/trainers/trainer.py
```diff
@@ -192,7 +193,9 @@ def export_model(self, name_behavior_id: str) -> None:
         """
         Exports the model
         """
-        self.get_policy(name_behavior_id).export_model()
+        policy = self.get_policy(name_behavior_id)
+        settings = SerializationSettings(policy.model_path, policy.brain.brain_name)
+        export_policy_model(settings, policy.graph, policy.sess)
```
ervteng: 👍 Do like moving exporting out of Policy

chriselion:
@ervteng Barracuda importer will handle everything, we just need to export the ONNX.

docs/Unity-Inference-Engine.md
```diff
@@ -27,10 +27,17 @@ but we only tested for the following platforms :
 * iOS
 * Android
 
+## Supported formats
+There are currently two supported model formats:
```
chriselion: @surfnerd @ervteng - I made this change to the docs too. Can you take a look when you get a chance? (I might merge first and revise if needed).

PR #3108, opened 2019-12-19
Remove TrainerMetrics and add CSVWriter using new StatsWriter API

ervteng:
The old TrainerMetrics class provided a subset of the functionality of the Timers and the Tensorboard writing. 

This PR uses the new StatsWriter API to implement a new CSV writer, and removes the TrainerMetrics class. 

- The StatsWriter API was changed slightly to write batches of stats, and to pass std and num as well (not used in TensorboardWriter and CSVWriter). 
- Limitation: The CSV writer only writes fields found during the first summary write, and ignores those after. For instance, it is possible that loss stats are not added until after the first summary step. If not, these will be ignored. I chose this behavior over, e.g., overwriting the title row as it will still include the specified required fields (inc. reward).


ml-agents/mlagents/trainers/stats.py
```diff
@@ -47,10 +59,43 @@ def write_text(self, category: str, text: str, step: int) -> None:
...
+        if category not in self.csv_fields:
+            summary_dir = self.base_dir
+            os.makedirs(summary_dir, exist_ok=True)
+            self.csv_fields[category] = keys
+            with open(self._get_filepath(category), "w") as file:
+                title_row = ["Steps"]
+                title_row.extend(keys)
+                writer = csv.writer(file)
+                writer.writerow(title_row)
+                self.csv_fields[category] = keys
```
chriselion: nit: redundant? You do the same before the `with...`

ervteng: Good catch  - removed

PR #3112, opened 2019-12-20
check min size for visual encoders

chriselion:
The visual encoders have a minimum usable size due to the strides. Currently, if you go below this size, we get an ugly looking exception in keras.

This PR adds a friendlier exception when you go under the min size, and adds tests to make sure the min is actually the min.

### DIFF
--- a/ml-agents/mlagents/trainers/models.py
+++ b/ml-agents/mlagents/trainers/models.py
@@ -490,6 +490,17 @@ class LearningModel:
             ),
         )
 
+    @staticmethod
+    def _check_resolution_for_encoder(
+        camera_res: CameraResolution, vis_encoder_type: EncoderType
+    ) -> None:
+        min_res = LearningModel.MIS_RESOLUTION_FOR_ENCODER[vis_encoder_type]
+        if camera_res.height < min_res or camera_res.width < min_res:
+            raise UnityTrainerException(
+                f"Visual observation resolution ({camera_res.width}x{camera_res.height}) is too small for"
+                f"the provided EncoderType ({vis_encoder_type.value}). The min dimension is {min_res}"
+            )
+
     def create_observation_streams(
         self,
         num_streams: int,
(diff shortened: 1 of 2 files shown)


ml-agents/mlagents/trainers/models.py
```diff
@@ -427,6 +438,17 @@ def create_resnet_visual_observation_encoder(
             )
         return hidden_flat
 
+    @staticmethod
+    def get_encoder_for_type(encoder_type: EncoderType) -> EncoderFunction:
+        ENCODER_FUNCTION_BY_TYPE = {
```
chriselion: Don't think I can make this dict a class member :/

ml-agents/mlagents/trainers/models.py
```diff
@@ -26,9 +29,17 @@ class LearningRateSchedule(Enum):
     LINEAR = "linear"
 
 
-class LearningModel(object):
+class LearningModel:
     _version_number_ = 2
 
+    # Minimum supported side for each encoder type. If refactoring an encoder, please
+    # adjust these also.
+    MIS_RESOLUTION_FOR_ENCODER = {
```
ervteng: nit: should this be `MIN` and not `MIS`?

chriselion: 🤦‍♂ yup, will fix

ml-agents/mlagents/trainers/tests/test_ppo.py
```diff
@@ -499,5 +501,41 @@ def test_normalization(dummy_config):
...
+            )
+            LearningModel._check_resolution_for_encoder(good_res, encoder_type)
+            vis_input = LearningModel.create_visual_input(
+                good_res, "test_min_visual_size"
+            )
+            enc_func = LearningModel.get_encoder_for_type(encoder_type)
+            enc_func(vis_input, 32, LearningModel.swish, 1, "test", False)
+
+        # Anything under the min size should raise an exception. If not, decrease the min size!
+        with pytest.raises(Exception):
```
ervteng: Should we except the specific exception (and check that it's the negative dimension issue) just in case the model has a different error?

chriselion: I don't think we need the exact type; the test is sufficient to make sure that size N works and N-1 doesn't. Besides, it's possible that different encoders would fail for different reasons.

PR #3113, opened 2019-12-21
Add Trajectory and Policy Queues

ervteng:
This PR adds Trajectory and Policy Queues between the env_manager, AgentProcessor, and Trainers. Getting one step closer to decoupling the environment stepping and the trainer training. 

However there are still two points of coupling that remain:

- ~~Summary and model writing. When the summaries are written is driven by the TrainerController currently - the TC tells the Trainers when to write out to Tensorboard. It would be better to have the trainer write out when appropriate during the `trainer.advance()` call -> Addressed in PR: https://github.com/Unity-Technologies/ml-agents/pull/3124~~

- Curriculum. Somehow, the TC still needs to know what the individual trainers' rewards are to update reset parameters in the environment. I think we might be able to get these stats from the AgentProcessor instead. - **Postponed to future PR**


ml-agents/mlagents/trainers/agent_processor.py
```diff
@@ -134,8 +133,8 @@ def add_experiences(
                         agent_id=agent_id,
                         next_obs=next_obs,
                     )
-                    # This will eventually be replaced with a queue
-                    self.trainer.process_trajectory(trajectory)
+                    for _traj_queue in self.trajectory_queues:
```
chriselion: nit: don't use leading underscores for local variables.

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -185,6 +188,25 @@ def start_trainer(self, trainer: Trainer, env_manager: EnvManager) -> None:
...
+        agent_manager = AgentManager(
+            processor=AgentProcessor(
+                trainer.policy,
+                trainer.stats_reporter,
+                trainer.parameters.get("time_horizon", sys.maxsize),
+            ),
+            trajectory_queue=Queue(),
+            policy_queue=Queue(),
+        )
+        agent_manager.processor.publish_trajectory_queue(agent_manager.trajectory_queue)
```
chriselion: This feels strange that the AgentManager doesn't subscribe to its own trajectory_queue. I'd recommend making AgentManager its own class and set this up in `AgentManager.__init__` (you can maybe do this as a namedtuple still, but I think you have to do it in `__new__` instead)

ml-agents/mlagents/trainers/trainer.py
```diff
@@ -5,13 +5,15 @@
 from mlagents.tf_utils import tf
 
 from collections import deque
+from queue import Queue
```
chriselion: Why do we need `queue.Queue` here instead of something lighter-weight like `collections.deque`? I don't think we need the synchronization that Queue provides (https://docs.python.org/3/library/queue.html)

ervteng: Not yet :) The intention was to enable the environment stepping and trainer advancing to run in separate threads (or even processes) for algorithms like APE-X. Eventually we'd want the trainer to wait until a Trajectory is available before doing any computation. I went with Queue since the method calls between Queue (and multiprocessing.Queue) are different from `deque`. 

We could also just use `deque` until the time comes to change over.

chriselion: Sounds good.

ervteng: It appears that `deque` is actually thread-safe as long as you're using just `get` and `put`, so I'll switch to that. Even if we go multi-thread we can still use `deque`, and it's faster. 

https://stackoverflow.com/questions/717148/queue-queue-vs-collections-deque

ml-agents/mlagents/trainers/trainer_controller.py
ervteng: I made AgentManager a subclass of AgentProcessor, that contains 2 queues. Didn't integrate it with AgentProcessor since in the future we might have processors that subscribe to multiple queues.

ml-agents/mlagents/trainers/agent_processor.py
ervteng: Fixed

ml-agents/mlagents/trainers/agent_processor.py
```diff
@@ -156,3 +156,49 @@ def add_experiences(
...
+    out this implementation.
+    """
+
+    def __init__(self, behavior_id: str = None):
+        """
+        Initializes an AgentManagerQueue. Note that we can give it a behavior_id so that it can be identified
+        separately from an AgentManager.
+        """
+        self.queue: Deque[Union[Trajectory, Policy]] = deque()
+        self.behavior_id = behavior_id
```
ervteng: Giving the queue a behavior_id is understandably a bit weird. I did it this way so that the Trainer can retrieve which behavior id the policy queue belongs to and publish the right policy to it. 

The alternative is to give the Trainer access to the AgentManager itself, but I wanted to keep the AgentManager abstraction separate from the Trainer. I'd imagine in the future, we'd want to assemble queues and Trainers in a different way.

docs/Migrating.md
```diff
@@ -19,12 +19,14 @@ The versions can be found in
 * Offline Behavioral Cloning has been removed. To learn from demonstrations, use the GAIL and
 Behavioral Cloning features with either PPO or SAC. See [Imitation Learning](Training-Imitation-Learning.md) for more information.
 * `mlagents.envs` was renamed to `mlagents_envs`. The previous repo layout depended on [PEP420](https://www.python.org/dev/peps/pep-0420/), which caused problems with some of our tooling such as mypy and pylint.
+* Trainer steps are now counted per-Agent, not per-environment as in previous versions. For instance, if you have 10 Agents in the scene, 20 environment steps now corresponds to 200 steps as printed in the terminal and in Tensorboard.
```
chriselion: Unless you're going to merge this into the release branch, we should make a "Migrating from 0.13 to latest" section and put these there.

ml-agents/mlagents/trainers/trainer.py
chriselion: Should anything still be using queue.Queue()? If not, you should remove the imports and see what's breaking.

ml-agents/mlagents/trainers/trainer.py
```diff
@@ -165,97 +189,129 @@ def export_model(self, name_behavior_id: str) -> None:
...
+        :param queue: Policy queue to publish to.
+        """
+        self.policy_queues.append(policy_queue)
+
+    def subscribe_trajectory_queue(self, trajectory_queue: Queue) -> None:
+        """
+        Adds a trajectory queue to the list of queues for the trainer injest Trajectories from.
+        :param queue: Trajectory queue to publish to.
+        """
+        self.trajectory_queues.append(trajectory_queue)
```
chriselion: Wrong types here - without https://github.com/Unity-Technologies/ml-agents/pull/3151 mypy doesn't understand what an AgentManagerQueue is. With it, you'll get an error like
```
Argument 1 to "append" of "list" has incompatible type "Queue[Any]"; expected "AgentManagerQueue"
```

ml-agents/mlagents/trainers/agent_processor.py
```diff
@@ -156,3 +156,49 @@ def add_experiences(
...
+
+
+class AgentManager(AgentProcessor):
+    """
+    An AgentManager is an AgentProcessor that also holds a single trajectory and policy queue.
+    Note: this leaves room for adding AgentProcessors that publish multiple trajectory queues.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
chriselion: I'd prefer to be explicit about the arguments here, but if you think it's too tedious, ignore me.

ml-agents/mlagents/trainers/trainer.py
```diff
@@ -165,97 +189,129 @@ def export_model(self, name_behavior_id: str) -> None:
...
+        """
         pass
+
+    def advance(self) -> None:
+        """
+        Steps the trainer, taking in trajectories and updates if ready.
+        """
+        with hierarchical_timer("process_trajectory"):
+            for traj_queue in self.trajectory_queues:
+                if not traj_queue.empty():
```
chriselion: I think if there are ever multiple consumers here, you could run into problems - if there's one element in the queue, both could think it's non-empty, and one would try to grab it and the other would fail and raise an exception. Might be better to just to a try/except around the `get_nowait()` call, and avoid the `not empty()` check.

ml-agents/mlagents/trainers/trainer.py
ervteng: I think some of the tests still do - will switch them out

ml-agents/mlagents/trainers/agent_processor.py
ervteng: Explicit it is

ml-agents/mlagents/trainers/trainer.py
ervteng: Added an AgentManagerQueue.Empty exception and now using try/except here and in TC

docs/Migrating.md
ervteng: Done

PR #3121, opened 2019-12-23
Bump the official supported Unity version to 18.4

surfnerd:


UnitySDK/Packages/manifest.json
```diff
@@ -0,0 +1,41 @@
...
+    "com.unity.modules.unityanalytics": "1.0.0",
+    "com.unity.modules.unitywebrequest": "1.0.0",
+    "com.unity.modules.unitywebrequestassetbundle": "1.0.0",
+    "com.unity.modules.unitywebrequestaudio": "1.0.0",
+    "com.unity.modules.unitywebrequesttexture": "1.0.0",
+    "com.unity.modules.unitywebrequestwww": "1.0.0",
+    "com.unity.modules.vehicles": "1.0.0",
+    "com.unity.modules.video": "1.0.0",
+    "com.unity.modules.vr": "1.0.0",
+    "com.unity.modules.wind": "1.0.0",
```
chriselion: Do we really need all these?

surfnerd: It failed without them

surfnerd: I can try to trim it down manually if you want me to.  This is what the editor put in by default.  Aren't you on PTO? :)

surfnerd:
> Guess we can get rid of the 2017.4 yamato tests now :)

Doing that now :)

UnitySDK/Packages/manifest.json
chriselion: Just seems weird that we need wind and particles. But maybe they're transitive dependencies from other packages that we do need.

PR #3122, opened 2019-12-23
Develop magic string + trajectory

andrewcoh:
This is a new PR for spawning multiple brain names with the new trajectory-centric structure of the trainers as outlined in this [design doc](https://docs.google.com/document/d/1qg0CSVWljfAprhE8sk2XFiydvmFXfRMtJRr1jRJhw48/edit#heading=h.v7t1ynb2u4tr). As is, it passes all pytest/C# tests and trains 3DBall. 

The major changes:

- behavior name is now the root behavior name + team id (team id field added to the editor)
- python side has some parsing in trainer controller to handle name_behavior_id
- policies are created in a separate function within the trainer
- getter/setter for policies by name_behavior_id
- self.policy is removed from trainer.py. Multiagent trainers can now sensibly inherit from this hierarchy. save_model/export_model now expect a name_behavior_id identifier for a policy.
- modified tests for new trainer functions

The minor changes:

- trainer doesn't take brain_parameters as an argument, just brain name.


ml-agents/mlagents/trainers/ppo/trainer.py
```diff
@@ -161,7 +154,9 @@ def process_trajectory(self, trajectory: Trajectory) -> None:
 
         # If this was a terminal trajectory, append stats and reset reward collection
         if trajectory.done_reached:
-            self._update_end_episode_stats(agent_id)
+            self._update_end_episode_stats(
+                agent_id, self.get_policy(trajectory.behavior_id)
```
ervteng: Clever

ml-agents/mlagents/trainers/trainer.py
```diff
@@ -152,19 +151,19 @@ def increment_step(self, n_steps: int) -> None:
 
         :param n_steps: number of steps to increment the step count by
         """
-        self.step = self.policy.increment_step(n_steps)
+        self.step += n_steps
```
ervteng: Are we still incrementing the policy's steps somewhere? I believe the steps are needed to update the nodes in the TF graph that hold the global steps, which are used for annealing the different params

andrewcoh: Yes, the policy steps get updated right after this is called in trainer_controller.

ervteng: Is there a reason why we don't just have `increment_step(self, n_steps, behavior_name_id)` and do the get_policy inside the method? So we don't have to keep track of two step increments

ervteng: Basically in my latest code (https://github.com/Unity-Technologies/ml-agents/commit/28ad94a733e2dc79cea3a59521804e2e52350a6e) I've replaced all of these calls on the trainer with a single `advance()` call that takes care of the incrementing, trajectory ingestion, and policy updates.

ml-agents/mlagents/trainers/sac/trainer.py
```diff
@@ -257,7 +283,7 @@ def update_sac_policy(self) -> None:
         for stat, stat_list in batch_update_stats.items():
             self.stats_reporter.add_stat(stat, np.mean(stat_list))
 
-        bc_module = self.sac_policy.bc_module
+        bc_module = self.policy.bc_module
```
chriselion: I'm surprised this works since self.policy should be a TFPolicy not a SACPolicy

ml-agents/mlagents/trainers/sac/trainer.py
```diff
@@ -27,7 +29,14 @@ class SACTrainer(RLTrainer):
     """
 
     def __init__(
-        self, brain, reward_buff_cap, trainer_parameters, training, load, seed, run_id
+        self,
```
chriselion: Could you add type annotations on these?

ml-agents/mlagents/trainers/ppo/trainer.py
```diff
@@ -20,7 +22,7 @@ class PPOTrainer(RLTrainer):
 
     def __init__(
         self,
-        brain,
+        brain_name,
```
chriselion: Ditto for type annotations

ml-agents/mlagents/trainers/sac/trainer.py
andrewcoh: Is it because I removed self.policy from trainer?  self.policy now just exist at the level of ppo/sac

ml-agents/mlagents/trainers/sac/trainer.py
andrewcoh: will do

ml-agents/mlagents/trainers/ppo/trainer.py
andrewcoh: it's odd that the trainers have escaped type annotations for this long.

PR #3135, opened 2019-12-30
Fix mypy errors in trainer code.

chriselion:


.pre-commit-config.yaml
```diff
@@ -10,11 +10,7 @@ repos:
             )$
 
 -   repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.750
-    # Currently mypy may assert after logging one message. To get all the messages at once, change repo and rev to
-    #   repo: https://github.com/chriselion/mypy
-    #   rev: 3d0b6164a9487a6c5cf9d144110b86600fd85e25
-    # This is a fork with the assert disabled, although precommit has trouble installing it sometimes.
+    rev: v0.761
```
chriselion: Tried bumping this to get the errors to show, but no luck. Removed the obsolete comment.

ml-agents/mlagents/trainers/sac/policy.py
```diff
@@ -206,7 +206,7 @@ def update(
         return update_stats
 
     def update_reward_signals(
-        self, reward_signal_minibatches: Dict[str, Dict], num_sequences: int
+        self, reward_signal_minibatches: Mapping[str, Dict], num_sequences: int
```
chriselion: This was getting Dict[str, AgentBuffer] passed to it, which causes the helpful errors:
```
ml-agents/mlagents/trainers/sac/trainer.py:322: note: "Dict" is invariant -- see http://mypy.readthedocs.io/en/latest/common_issues.html#variance
ml-agents/mlagents/trainers/sac/trainer.py:322: note: Consider using "Mapping" instead, which is covariant in the value type
```

ml-agents/mlagents/trainers/ppo/trainer.py
```diff
@@ -70,7 +70,7 @@ def __init__(
         self.load = load
         self.multi_gpu = multi_gpu
         self.seed = seed
-        self.policy: TFPolicy = None
+        self.policy: PPOPolicy = None  # type: ignore
```
chriselion: Ignore the error of setting this to None, since we'll always set it to an instance before using.

Issue #3133, opened 2019-12-29
BUG?  gym wrapper's step() returns "BatchedStepResult" instead of "dict" for the fourth param.


weiziqian:
the 4th. return parameter ("info") for function step() should be "dict";
however in a recent update (2019 dec 10) it was changed into "BatchedStepResult".

which at least caused troubles when training with baseline,
e.g.:
File "/xxx/stable-baselines/stable_baselines/td3/td3.py", line 343, in learn
    maybe_ep_info = info.get('episode')
AttributeError: 'BatchedStepResult' object has no attribute 'get'

chriselion:
Thanks @weiziqian - this does look like a bug. I've logged it as MLA-486 in our internal tracker, and we should have a fix for it soon.

chriselion:
I believe https://github.com/Unity-Technologies/ml-agents/pull/3136 will fix this, but I want to run it by some other folks that are more familiar with the gym interface.

chriselion:
@weiziqian - this is fixed on `master` branch now, and will be in the next release. Thanks for the report!

chriselion:
I also logged a longer-term issue for us to add better testing around this; we have a few unit tests but no integration tests that run training with the gym wrapper.

weiziqian:
hi Chris, thank you so much for such a prompt response!

PR #3136, opened 2019-12-30
[MLA-486] type annotations on gym interface, return dict from step

chriselion:
Fixes https://github.com/Unity-Technologies/ml-agents/issues/3133

I need someone more familiar with gym to review this. From what I can tell, we're returning different types based on whether we're in single or multi agent mode, which is ugly at best. I need confirmation this is actually the intended behavior.

### DIFF
--- a/gym-unity/gym_unity/envs/__init__.py
+++ b/gym-unity/gym_unity/envs/__init__.py
@@ -1,9 +1,13 @@
 import logging
 import itertools
-import gym
 import numpy as np
-from mlagents_envs.environment import UnityEnvironment
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+import gym
 from gym import error, spaces
+
+from mlagents_envs.environment import UnityEnvironment
+from mlagents_envs.base_env import BatchedStepResult
 
 
 class UnityGymException(error.Error):
@@ -16,6 +20,11 @@ class UnityGymException(error.Error):
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger("gym_unity")
+
+
+GymSingleStepResult = Tuple[np.ndarray, float, bool, Dict]
+GymMultiStepResult = Tuple[List[np.ndarray], List[float], List[bool], Dict]
+GymStepResult = Union[GymSingleStepResult, GymMultiStepResult]
 
 
 class UnityEnv(gym.Env):
@@ -138,7 +147,7 @@ class UnityEnv(gym.Env):
         else:
             self._observation_space = spaces.Box(-high, high, dtype=np.float32)
 
-    def reset(self):
+    def reset(self) -> Union[List[np.ndarray], np.ndarray]:
         """Resets the state of the environment and returns an initial observation.
         In the case of multi-agent environments, this is a list.
         Returns: observation (object/list): the initial observation of the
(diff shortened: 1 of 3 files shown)


ml-agents-envs/mlagents_envs/rpc_utils.py
```diff
@@ -123,13 +123,13 @@ def batched_step_result_from_proto(
         is_visual = len(obs_shape) == 3
         if is_visual:
             obs_shape = cast(Tuple[int, int, int], obs_shape)
-            obs_list += [
+            obs_list.append(
```
chriselion: Found this when looking for other things like what `_get_vis_obs_list()` did. These have the same effect, but don't require creating an intermediate list object, and are clearer (IMHO).

gym-unity/gym_unity/envs/__init__.py
```diff
@@ -151,12 +160,12 @@ def reset(self):
         self.game_over = False
 
         if not self._multiagent:
-            obs, reward, done, info = self._single_step(info)
+            res: GymStepResult = self._single_step(info)
         else:
-            obs, reward, done, info = self._multi_step(info)
-        return obs
+            res = self._multi_step(info)
```
ervteng: Is the type annotation only required for the 1st clause in the if statement?

chriselion: yeah, it's kind of annoying because mypy complains if you put it on both (even if it's the same type). 

The idea here is that _single_step and _multi_step return different types, but GymStepResult is the union of those two.

ervteng:
> From what I can tell, we're returning different types based on whether we're in single or multi agent mode, which is ugly at best. I need confirmation this is actually the intended behavior.

Yep, it's unfortunately intended behavior. The issue is that there is no multi-agent interface for Gym (at least, no official one). What we do is just return a List for each of the obs, reward, dones as returned by `step`. 

Returning a dict as info is fine. Ideally we'd convert the BatchedStepResult tuple into a dict of the values, but the Gym interface spec (http://gym.openai.com/docs/) doesn't specify a type for the values of the info dict, so we're technically in spec.

gym-unity/gym_unity/envs/__init__.py
```diff
@@ -226,21 +236,31 @@ def _single_step(self, info):
         else:
             default_observation = self._get_vector_obs(info)[0, :]
 
-        return (default_observation, info.reward[0], info.done[0], info)
+        return (
+            default_observation,
+            info.reward[0],
+            info.done[0],
+            {"batched_step_result": info},
```
ervteng: Ideally we'd turn the batched_step_result NamedTuple into a dict of its fields so it can be easily accessed without knowing anything about the `BatchedStepResult` object, but this is technically in spec with Gym.

chriselion: Previously, we were using a BrainInfo as the value, so I think I'm OK with this for now. 

I'm not sure what other keys are generally used, but here's where it was failing in the original issue: https://github.com/hill-a/stable-baselines/blob/6bdb7ce8ed03224a8c03d3197ff59b86cbcb0b94/stable_baselines/td3/td3.py#L336-L343

chriselion: Previously, we were using a BrainInfo as the value, so I think I'm OK with this for now.

I'm not sure what other keys are generally used, but here's where it was failing in the original issue: https://github.com/hill-a/stable-baselines/blob/6bdb7ce8ed03224a8c03d3197ff59b86cbcb0b94/stable_baselines/td3/td3.py#L336-L343

ervteng: The keys available are completely arbitrary :P I suppose the ATARI envs have some standard ones that are available.

PR #3137, opened 2019-12-30
add flake8-bugbear

chriselion:
Adds another flake8 plugin (https://pypi.org/project/flake8-bugbear/) that provides some useful extra checks (in particular, warning about mutable default args).

### DIFF
--- a/ml-agents/mlagents/trainers/demo_loader.py
+++ b/ml-agents/mlagents/trainers/demo_loader.py
@@ -31,7 +31,7 @@ def make_demo_buffer(
     for idx, experience in enumerate(pair_infos):
         if idx > len(pair_infos) - 2:
             break
-        current_pair_info = pair_infos[idx]
+        current_pair_info = experience
         next_pair_info = pair_infos[idx + 1]
         current_brain_info = BrainInfo.from_agent_proto(
             0, [current_pair_info.agent_info], brain_params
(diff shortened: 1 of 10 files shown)


ml-agents/mlagents/trainers/demo_loader.py
```diff
@@ -31,7 +31,7 @@ def make_demo_buffer(
     for idx, experience in enumerate(pair_infos):
         if idx > len(pair_infos) - 2:
             break
-        current_pair_info = pair_infos[idx]
+        current_pair_info = experience
```
harperj: minor, but this should probably just be named `current_pair_info` in the `for` statement

chriselion: Good call, that's better.

ml-agents/mlagents/trainers/demo_loader.py
```diff
@@ -28,10 +28,9 @@ def make_demo_buffer(
     # Create and populate buffer using experiences
     demo_raw_buffer = AgentBuffer()
     demo_processed_buffer = AgentBuffer()
-    for idx, experience in enumerate(pair_infos):
+    for idx, current_pair_info in enumerate(pair_infos):
```
chriselion: @kevtan Heads up on this, might cause merge conflicts with what you're working on.

PR #3149, opened 2020-01-03
Refactor and cleaning up demonstration file finding logic + associated test

kevtan:
Cleaned up the logic.
---
Figuring out how to contribute more sizable code.

### DIFF
--- a/ml-agents/mlagents/trainers/demo_loader.py
+++ b/ml-agents/mlagents/trainers/demo_loader.py
@@ -115,6 +115,8 @@ def load_demonstration(
     :return: BrainParameter and list of AgentInfoActionPairProto containing demonstration data.
     """
 
+    # First 32 bytes of file dedicated to meta-data.
+    INITIAL_POS = 33
     file_paths = get_demo_files(file_path)
 
     brain_params = None
@@ -132,13 +134,12 @@ def load_demonstration(
                     meta_data_proto = DemonstrationMetaProto()
                     meta_data_proto.ParseFromString(data[pos : pos + next_pos])
                     total_expected += meta_data_proto.number_steps
-                    # first 32 bytes of file dedicated to meta-data.
-                    pos = 33
-                elif obs_decoded == 1:
+                    pos = INITIAL_POS
+                if obs_decoded == 1:
                     brain_param_proto = BrainParametersProto()
                     brain_param_proto.ParseFromString(data[pos : pos + next_pos])
                     pos += next_pos
-                else:
+                if obs_decoded > 1:
                     agent_info_action = AgentInfoActionPairProto()
                     agent_info_action.ParseFromString(data[pos : pos + next_pos])
                     if brain_params is None:
@@ -147,7 +148,6 @@ def load_demonstration(
                         )
                     info_action_pairs.append(agent_info_action)
                     if len(info_action_pairs) == total_expected:
-                        breakpoint()
                         break
                     pos += next_pos
                 obs_decoded += 1
--- a/ml-agents/mlagents/trainers/demo_loader.py
+++ b/ml-agents/mlagents/trainers/demo_loader.py
@@ -27,16 +27,12 @@ def make_demo_buffer(
     # Create and populate buffer using experiences
     demo_raw_buffer = AgentBuffer()
     demo_processed_buffer = AgentBuffer()
-    for idx, current_pair_info in enumerate(pair_infos):
-        if idx > len(pair_infos) - 2:
-            break
-        next_pair_info = pair_infos[idx + 1]
-        current_brain_info = BrainInfo.from_agent_proto(
-            0, [current_pair_info.agent_info], brain_params
-        )
-        next_brain_info = BrainInfo.from_agent_proto(
-            0, [next_pair_info.agent_info], brain_params
-        )
+    brain_infos = [
+        BrainInfo.from_agent_proto(0, [pair_info.agent_info], brain_params)
+        for pair_info in pair_infos
+    ]
+    for idx in range(len(brain_infos) - 1):
+        current_brain_info, next_brain_info = brain_infos[idx:idx+2]
         previous_action = (
             np.array(pair_infos[idx].action_info.vector_actions, dtype=np.float32) * 0
         )
@@ -54,7 +50,7 @@ def make_demo_buffer(
             demo_raw_buffer["vector_obs"].append(
                 current_brain_info.vector_observations[0]
             )
-        demo_raw_buffer["actions"].append(current_pair_info.action_info.vector_actions)
+        demo_raw_buffer["actions"].append(pair_infos[idx].action_info.vector_actions)
         demo_raw_buffer["prev_action"].append(previous_action)
         if next_brain_info.local_done[0]:
             demo_raw_buffer.resequence_and_append(
--- a/ml-agents/mlagents/trainers/demo_loader.py
+++ b/ml-agents/mlagents/trainers/demo_loader.py
@@ -135,7 +135,8 @@ def load_demonstration(
                     meta_data_proto.ParseFromString(data[pos : pos + next_pos])
                     total_expected += meta_data_proto.number_steps
                     # first 32 bytes of file dedicated to metadata
-                    pos = 33
+                    INITIAL_POS = 33
+                    pos = INITIAL_POS
                 elif obs_decoded == 1:
                     brain_param_proto = BrainParametersProto()
                     brain_param_proto.ParseFromString(data[pos : pos + next_pos])
--- a/ml-agents/mlagents/trainers/demo_loader.py
+++ b/ml-agents/mlagents/trainers/demo_loader.py
@@ -1,4 +1,3 @@
-import pathlib
 import logging
 import os
 from typing import List, Tuple
@@ -102,7 +101,9 @@ def get_demo_files(path: str) -> List[str]:
             raise ValueError("There are no '.demo' files in the provided directory.")
         return fpaths
     else:
-        raise FileNotFoundError(f"The demonstration file or directory {path} does not exist.")
+        raise FileNotFoundError(
+            f"The demonstration file or directory {path} does not exist."
+        )
 
 
 @timed
--- a/ml-agents/mlagents/trainers/demo_loader.py
+++ b/ml-agents/mlagents/trainers/demo_loader.py
@@ -32,7 +32,7 @@ def make_demo_buffer(
         for pair_info in pair_infos
     ]
     for idx in range(len(brain_infos) - 1):
-        current_brain_info, next_brain_info = brain_infos[idx:idx+2]
+        current_brain_info, next_brain_info = brain_infos[idx : idx + 2]
         previous_action = (
             np.array(pair_infos[idx].action_info.vector_actions, dtype=np.float32) * 0
         )
--- a/ml-agents/mlagents/trainers/demo_loader.py
+++ b/ml-agents/mlagents/trainers/demo_loader.py
@@ -115,8 +115,6 @@ def load_demonstration(
     :return: BrainParameter and list of AgentInfoActionPairProto containing demonstration data.
     """
 
-    # First 32 bytes of file dedicated to meta-data.
-    INITIAL_POS = 33
     file_paths = get_demo_files(file_path)
 
     brain_params = None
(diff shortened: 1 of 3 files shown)


ml-agents/mlagents/trainers/demo_loader.py
```diff
@@ -83,39 +82,40 @@ def demo_to_buffer(
...
-                file_paths.append(os.path.join(file_path, _file))
-        if not all_files:
+    Raises errors if |path| is invalid.
+    """
+    if os.path.isfile(path):
+        if not path.endswith(".demo"):
+            raise ValueError("The path provided is not a '.demo' file.")
+        return [path]
+    elif os.path.isdir(path):
+        fnames = filter(lambda fname: fname.endswith(".demo"), os.listdir(path))
```
chriselion: I generally dislike maps and filters (and the author of python agrees https://www.artima.com/weblogs/viewpost.jsp?thread=98196). I think the old code for this was OK, but if you want to make a change, can you use list comprehensions instead? E.g.
```
fnames = [fname for fname in os.listdir(path) if fname.endswith(".demo")]
```

ml-agents/mlagents/trainers/demo_loader.py
```diff
@@ -83,39 +82,40 @@ def demo_to_buffer(
...
+        fpaths = list(map(lambda fname: os.path.join(path, fname), fnames))
+        if len(fpaths) == 0:
```
chriselion: It's more pythonic to just write `if not fpaths:` - empty lists are "falsey".

ml-agents/mlagents/trainers/demo_loader.py
```diff
@@ -131,12 +131,13 @@ def load_demonstration(
                     meta_data_proto = DemonstrationMetaProto()
                     meta_data_proto.ParseFromString(data[pos : pos + next_pos])
                     total_expected += meta_data_proto.number_steps
-                    pos = INITIAL_POS
-                if obs_decoded == 1:
+                    # first 32 bytes of file dedicated to metadata
+                    pos = 33
```
chriselion: I think this was better as a constant - it's a bit less "magic" that way.
https://stackoverflow.com/questions/47882/what-is-a-magic-number-and-why-is-it-bad

ml-agents/mlagents/trainers/demo_loader.py
```diff
@@ -83,39 +82,40 @@ def demo_to_buffer(
...
+@timed
+def load_demonstration(
+    path: str,
+) -> Tuple[BrainParameters, List[AgentInfoActionPairProto], int]:
+    """
+    Loads and parses a demonstration file or directory.
+    :param path: File or directory.
+    :return: BrainParameter and list of AgentInfoActionPairProto containing demonstration data.
+    """
+    file_paths = get_demo_files(path)
```
chriselion: 👍 I like moving the file logic to its own function.

ml-agents/mlagents/trainers/tests/test_demo_loader.py
```diff
@@ -27,3 +32,11 @@ def test_load_demo_dir():
...
     _, demo_buffer = demo_to_buffer(path_prefix + "/test_demo_dir", 1)
     assert len(demo_buffer["actions"]) == total_expected - 1
+
+
+def test_nonexisting_paths():
+    path_prefix = os.path.dirname(os.path.abspath(__file__))
+    with pytest.raises(FileNotFoundError):
+        get_demo_files(os.path.join(path_prefix, "idontexistfile.demo"))
+    with pytest.raises(FileNotFoundError):
+        get_demo_files(os.path.join(path_prefix, "idontexistdirectory"))
```
chriselion: If you want to test some of the other logic that raises ValueError (not a .demo file, or no demo files in the directory), you can use the python tempfile library (https://docs.python.org/3/library/tempfile.html) to create directories and files that will automatically be cleaned up. So something like

```python
with tempfile.TemporaryDirectory() as tmpdirname:
    # get_demo_files(tmpdirname) should raise a ValueError
    with open(tmpdirname + "/mydemo.notademo", 'w') as f:
        f.write("I'm not a demo")
    # get_demo_files(tmpdirname) should raise a ValueError
    with open(tmpdirname + "/mydemo.demo", 'w') as f:
        f.write("I'm a demo file.")
    # get_demo_files(tmpdirname) should return a list with mydemo.demo in it
```
(that's just a rough outline, you get to fill in the commented parts yourself)

ml-agents/mlagents/trainers/demo_loader.py
kevtan: Thanks for the great suggestion! I'd gotten hooked on using `map` and `filter` lately and I had my own "aha" moment that there were clearer list/generator-comprehension alternatives.

ml-agents/mlagents/trainers/demo_loader.py
kevtan: Makes sense! It just seemed weird to me to declare the constant at the beginning of the function, when it wasn't clear what it's not clear what it's going to be used for. Let me know what you think about the latest change.

chriselion:
Are you going to add the tests for the ValueError conditions that I suggested? it's optional but would give a bit more coverage.

kevtan:
@chriselion Thank you for taking the time to write all the detailed comments. They were super helpful!

kevtan:
Completed the original task and local benchmarks were showing slightly better performance. Instead of taking up around 76% of the total demonstration loading time, the `BrainInfo.from_agent_proto` calls now only take up about 72% of the time. Also, trial runs that used to take around `0.016` seconds now take around `0.010` seconds.

ml-agents/mlagents/trainers/tests/test_demo_loader.py
```diff
@@ -27,3 +33,28 @@ def test_load_demo_dir():
...
+        with pytest.raises(ValueError):
+            get_demo_files(tmpdirname)
+        # invalid file
+        invalid_fname = tmpdirname + "/mydemo.notademo"
+        with open(invalid_fname, "w") as f:
+            f.write("I'm not a demo")
+        with pytest.raises(ValueError):
+            get_demo_files(invalid_fname)
+        # valid file
+        valid_fname = tmpdirname + "/mydemo.demo"
```
chriselion: nit: you use `os.path.join` above but string concat here. I'd recommend using os.path.join for consistency (I was being lazy in my sample code).

ml-agents/mlagents/trainers/tests/test_demo_loader.py
```diff
@@ -27,3 +33,28 @@ def test_load_demo_dir():
...
+        get_demo_files(os.path.join(path_prefix, "nonexistent_directory"))
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        # empty directory
+        with pytest.raises(ValueError):
+            get_demo_files(tmpdirname)
+        # invalid file
+        invalid_fname = tmpdirname + "/mydemo.notademo"
+        with open(invalid_fname, "w") as f:
+            f.write("I'm not a demo")
+        with pytest.raises(ValueError):
```
chriselion: You should also confirm that `get_demo_files(tmpdirname)` fails here too (since there is a file but it gets filtered out).

ml-agents/mlagents/trainers/demo_loader.py
```diff
@@ -28,16 +27,12 @@ def make_demo_buffer(
...
-        if idx > len(pair_infos) - 2:
-            break
-        next_pair_info = pair_infos[idx + 1]
-        current_brain_info = BrainInfo.from_agent_proto(
-            0, [current_pair_info.agent_info], brain_params
-        )
-        next_brain_info = BrainInfo.from_agent_proto(
-            0, [next_pair_info.agent_info], brain_params
-        )
+    brain_infos = [
```
chriselion: I'm a little worried that this will eat up a lot of memory for large demo files, since it will decompress everything at once (in the original issue, demos were 1GB and that includes compressed visual observations). I think a better approach would be something like (very rough sketch):

``` python
current_brain_info = BrainInfo.from_agent_proto(0, [pair_infos[0].agent_info], brain_params)  # TODO handle empty pair_infos
for idx, current_pair_info in enumerate(pair_infos):
  # (dont calculate current_brain_info)
  # (rest of the loop stays the same)
  current_brain_info = next_brain_info
```
Does that sounds reasonable?

kevtan: Yes, great suggestion! This way, the entire demonstration file doesn't need to be in memory at once. I want to write a test for the demonstration loading that verifies that the loading is still being done correctly. My idea is to take the old implementation, load the demo files in `test_demo_dir` with `demo_to_buffer` and then pickle the returned objects and save them somewhere as the "correct answer." Then, in `test_demo_loader.py`, load in the pickled objects and compare them against what the current implementation of `demo_to_buffer` is returning.

The problem is that the `AgentBuffer` and `BrainParameters` classes don't have defined `__eq__` methods yet. Any ideas? Should I go through with the test or not? If so, what should my next step be?

chriselion: > This way, the entire demonstration file doesn't need to be in memory at once.

The way I proposed still has the protobuf objects loaded all at once (this is the current behavior) but it at least doesn't increase the peak. Adding things to the buffer as soon as they're read from the file is another option, but could probably wait until another PR.

Instead of saving pickled data, how about writing new demonstration files and comparing the values? We don't currently have a python implementation of this, but I think it would be useful to have anyway.

You can see how this is done in an old gist that I wrote: https://gist.github.com/chriselion/3714d05255eea2f9132b96a182fbdcaa#file-convert_demo-py-L101-L115
(the structures being written might have changed a bit, but the overall format is the same).

So the test would be something like
1) Create new BrainParameters and AgentInfo protos with specific values
2) Save as a new demonstration to a temp directory (or temp file)
3) Load and compare to the original values

Adding `__eq__` to BrainParameters sounds fine if you think it would help. I'm not sure it's necessary for AgentBuffer since that inherits from dict (whether or not it should is another discussion).

kevtan: I don't think I quite get what you mean. Every iteration through the for loop (in your proposed implementation), don't we lose an object reference? In particular, the object that `current_brain_info` referred to before `current_brain_info` and `next_brain_info`  got updated. After the reference to that object is lost, wouldn't Python garbage collect that data, and, so not all the demonstration file is in memory at once?

And, just so I'm clear, you're proposing to reconvert the loaded demonstration file contents back into a demonstration file (in a temporary directory) and then directly compare the file contents? I was just thinking about if we wanted to have tests for different `sequence_length`s for the `make_demo_buffer` function. Wouldn't this approach only work for `sequence_length` = 1?

chriselion: Sorry for the delay.

By "still has the protobuf objects loaded all at once", I meant the `pair_infos` list. If we wanted to reduce the total memory, we could combine `load_demonstration` and `make_demo_buffer` so that the `make_demo_buffer()` loop logic happens right after `agent_info_action.ParseFromString()`. But I don't think we should worry about this for now; I just don't want to _increase_ the peak usage, but we don't need to try to decrease it.

For a test, I'm suggesting something like
1. Make temporary directory
2. Create BrainParameters and AgentInfos
3. Write demo file using BrainParameters and AgentInfos to temp directory
4. Run load_demonstration() and make_demo_buffer() to get the a buffer
5. Make sure the result from 4 agrees with what should be produced from 2 (I'm not exactly sure what this looks like off the top of my head, and not quite sure about the sequence length)

chriselion: A logistical note - there are some more changes to this surrounding code in the next week or so. To minimize the number of merge conflicts you're going to have to deal with, let's try to get the get_demo_files() and load_demonstration() changed merged, and come back to the make_demo_buffer / proto conversion change after. 

Can you either undo the changes to make_demo_buffer(), or move them to another PR and I'll approve that? (the former is probably easier)

kevtan: @chriselion Sorry for the delay. School just started for me and it took a while to get myself settled in! I'll undo the changes now so that you can merge the `get_demo_files()` and `load_demonstration()` changes!

chriselion: No worries. Let me know if you want to keep working on this, or want to try something else.

kevtan: @chriselion I really want to complete this first assignment, and I'll try to do so by the end of this week! But, that being said, I'd really appreciate it if you could start to onboard me to the next thing I could be working on at the same time. Much appreciated!

PR #3152, opened 2020-01-03
queue type cleanup

chriselion:


ml-agents/mlagents/trainers/agent_processor.py
```diff
@@ -178,16 +181,16 @@ def __init__(self, behavior_id: str = None):
         Initializes an AgentManagerQueue. Note that we can give it a behavior_id so that it can be identified
         separately from an AgentManager.
         """
-        self.queue: Deque[Union[Trajectory, Policy]] = deque()
+        self.queue: Deque[T] = deque()
```
chriselion: I assume this is what you were going for with the Union, this makes it a bit more general

PR #3171, opened 2020-01-07
use TensorShape for index calc

chriselion:
Followup from the last PR - save a TensorShape instead of the int[] in WriteAdapter to reuse the index calculation code from there

This also undoes a signature change, since we don't need the shape when passing a TensorProxy to write (since the tensor already knows its shape).

### DIFF
--- a/UnitySDK/Assets/ML-Agents/Scripts/Sensor/WriteAdapter.cs
+++ b/UnitySDK/Assets/ML-Agents/Scripts/Sensor/WriteAdapter.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using Barracuda;
 using MLAgents.InferenceBrain;
 
 namespace MLAgents.Sensor
@@ -15,7 +16,7 @@ namespace MLAgents.Sensor
         TensorProxy m_Proxy;
         int m_Batch;
 
-        int[] m_Shape;
+        TensorShape m_TensorShape;
 
         /// <summary>
         /// Set the adapter to write to an IList at the given channelOffset.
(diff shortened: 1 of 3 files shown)


UnitySDK/Assets/ML-Agents/Scripts/Sensor/WriteAdapter.cs
```diff
@@ -39,13 +48,13 @@ public void SetTarget(IList<float> data, int[] shape, int offset)
...
         /// <param name="channelOffset">Offset from the start of the channel to write to.</param>
-        public void SetTarget(TensorProxy tensorProxy, int[] shape, int batchIndex, int channelOffset)
+        public void SetTarget(TensorProxy tensorProxy, int batchIndex, int channelOffset)
         {
             m_Proxy = tensorProxy;
             m_Batch = batchIndex;
             m_Offset = channelOffset;
             m_Data = null;
-            m_Shape = shape;
+            m_TensorShape = m_Proxy.data.shape;
```
chriselion: This isn't used in the tensor case, but just set it for consistency.

UnitySDK/Assets/ML-Agents/Scripts/Sensor/WriteAdapter.cs
```diff
@@ -80,26 +88,21 @@ public void SetTarget(TensorProxy tensorProxy, int[] shape, int batchIndex, int
...
                     {
-                        throw new IndexOutOfRangeException($"channel value {ch} must be in range [0, {channels-1}]");
+                        throw new IndexOutOfRangeException($"channel value {ch} must be in range [0, {m_TensorShape.channels-1}]");
                     }
 
-                    // Math copied from TensorShape.Index(). Note that m_Batch should always be 0
-                    var index = m_Batch * height * width * channels + h * width * channels + w * channels + ch;
-                    m_Data[index + m_Offset] = value;
+                    var index = m_TensorShape.Index(m_Batch, h, w, ch + m_Offset);
+                    m_Data[index] = value;
```
chriselion: Could compute the index for both TensorProxy and float array cases, and use `m_Proxy.data[index] = value;` but I don't see a strong reason for doing that. Same for the vector case.

PR #3181, opened 2020-01-08
Fix issue with different decision intervals for different brains

ervteng:
If there is more than one brain each with a different decision interval, only one of the brains would be updated. This was because we were checking if we had any actions present for each EnvironmentStep - and in the error case there would be no actions but still a braininfo present. 

We now process all the steps, but don't withdraw previous step unless there is an action.

### DIFF
--- a/ml-agents/mlagents/trainers/trainer_controller.py
+++ b/ml-agents/mlagents/trainers/trainer_controller.py
@@ -294,14 +294,15 @@ class TrainerController(object):
         with hierarchical_timer("env_step"):
             new_step_infos = env.step()
         for step_info in new_step_infos:
-            for brain_name in self.trainers.keys():
-                for name_behavior_id in self.brain_name_to_identifier[brain_name]:
-                    _processor = self.managers[name_behavior_id].processor
-                    _processor.add_experiences(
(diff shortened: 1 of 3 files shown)


ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -294,17 +294,15 @@ def advance(self, env: EnvManager) -> int:
...
-                        _processor.add_experiences(
-                            step_info.previous_all_brain_info[name_behavior_id],
-                            step_info.current_all_brain_info[name_behavior_id],
-                            step_info.brain_name_to_action_info[
-                                name_behavior_id
-                            ].outputs,
-                        )
+            for name_behavior_id in step_info.name_behavior_ids:
+                if name_behavior_id not in self.managers:
+                    continue
```
chriselion: Should we warn and/or throw an exception here?

ervteng: Added a warning

ervteng:
> Think it's good, let's just get another pair of eyes on it.

Cool, let's wait for Andrew to take a look before merging

PR #3183, opened 2020-01-08
run tensorflow 2.1 in CI

chriselion:
Tensorflow 2.1 was just released: https://github.com/tensorflow/tensorflow/releases/tag/v2.1.0

This is the first pass at checking support for it - if this passes, I'll retrain some scenes with it and make sure inference is OK.


chriselion:
Jumped the gun a bit, no pypi packages yet.

PR #3194, opened 2020-01-10
Self-play for symmetric games

andrewcoh:
Ghost trainer for symmetric games and single agent trainers and Tennis Refactor.


ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -0,0 +1,231 @@
...
+            ] = AgentManagerQueue(trajectory_queue.behavior_id)
+            self.trainer.subscribe_trajectory_queue(internal_trajectory_queue)
+
+
+# Taken from https://github.com/Unity-Technologies/ml-agents/pull/1975
+# ELO calculation
+K = 1  # Constant for rating changes, higher is less stable but converges faster
+
+
+def compute_elo_rating_changes(rating1, rating2, result):
```
ervteng: Maybe make a class method rather than having it down here somewhere

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def _update_policy(self) -> None:
+        pass
+
+    def advance(self) -> None:
+        """
+        Steps the trainer, taking in trajectories and updates if ready.
+        """
+        for traj_queue, internal_traj_queue in zip(
+            self.trajectory_queues, self.trainer.trajectory_queues
```
ervteng: Store reference to internal trajectory queues

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -0,0 +1,231 @@
...
+                internal_traj_queue.put(t)
+                self._process_trajectory(t)
+            except AgentManagerQueue.Empty:
+                pass
+
+        self.next_update_step = self.trainer.next_update_step
+        self.trainer.advance()
+        self._maybe_write_summary(self.get_step)
+
+        for q in self.trainer.policy_queues:
```
ervteng: Need to check if not q.empty()

ml-agents/mlagents/tf_utils/tf.py
```diff
@@ -1,3 +1,10 @@
+from __future__ import absolute_import
```
chriselion: This _is_ the future, you shouldn't need these.

ml-agents/mlagents/tf_utils/tf.py
```diff
@@ -28,3 +35,179 @@ def set_warnings_enabled(is_enabled: bool) -> None:
...
+    for shape in shapes:
+        size = np.prod(shape, dtype=np.int)
+        array = vector[i : (i + size)].reshape(shape)
+        arrays.append(array)
+        i += size
+    assert len(vector) == i, "Passed weight does not have the correct shape."
+    return arrays
+
+
+class TensorFlowVariables(object):
```
chriselion: Taken from https://github.com/ray-project/ray/blob/master/python/ray/experimental/tf_utils.py? Is there a better way to include this? Ray is probably pretty heavy weight.

ml-agents/mlagents/trainers/behavior_id_utils.py
```diff
@@ -0,0 +1,32 @@
...
+        Parses a name_behavior_id into a BehaviorIdentifiers NamedTuple.
+        This allows you to access the brain name and distinguishing identifiers
+        without parsing more than once.
+        :param name_behavior_id: String of behavior params in HTTP format.
+        :returns: A BehaviorIdentifiers object.
+        """
+
+        ids: Dict[str, int] = {}
+        if "?" in name_behavior_id:
+            name, identifiers = name_behavior_id.split("?")
```
chriselion: Can you add some unit tests for this? 

In particular, I think this will choke on names that have a "?" in them e.g. "brainName?whatsThat?team=0"
``` python
>>> name_behavior_id = "brainName?whatsThat?team=0"
>>> name_behavior_id.split("?")
['brainName', 'whatsThat', 'team=0']
>>> name_behavior_id.rsplit("?", 1)
['brainName?whatsThat', 'team=0']
```
`rsplit` will start splitting from the right, and the second arg will limit the number of splits to 1. (sorry for not thinking of this in the original PR)

ml-agents/mlagents/trainers/ghost/trainer.py
andrewcoh: fixed

ml-agents/mlagents/trainers/ghost/trainer.py
andrewcoh: fixed

ml-agents/mlagents/tf_utils/tf.py
andrewcoh: embarrassing. fixed

ml-agents/mlagents/tf_utils/tf.py
andrewcoh: I included this from @harperj's https://github.com/Unity-Technologies/mlagents-distributed. Open to suggestions.

ml-agents/mlagents/trainers/behavior_id_utils.py
andrewcoh: Should I just disallow the use of '?' in names on the C# side?

chriselion: Better to support them if you can - just changing to `.rsplit("?", 1)` should be enough

ml-agents/mlagents/tf_utils/tf.py
chriselion: They have an Apache 2.0 license https://github.com/ray-project/ray/blob/master/LICENSE; we can probably do this but will need to add attribution to our license too. You need to run this by Jeff and probably legal too.

harperj: I'd recommend we not re-distribute this code from Ray.  It's something that is easy enough to re-implement.  Essentially we just want to capture all of `tf.trainable_variables()`, create an assignment op for each, and then capture them in a dictionary so we have the values to reassign later.

Here's a previous version from when I was working on parallel environments:
```python
class Policy:
    # ... other methods here ...
    def serialize_in_memory(self):
        with self.graph.as_default():
            trainable_vars = tf.trainable_variables()
            values = [v.eval(session=self.sess) for v in trainable_vars]
            return values

    def load_from_memory(self, values):
        with self.graph.as_default():
            assign_ops = []
            feed_dict = {}
            trainable_vars = tf.trainable_variables()
            weights = zip(trainable_vars, values)
            for var, value in weights:
                value = np.asarray(value)
                assign_placeholder = tf.placeholder(var.dtype, shape=value.shape)
                assign_op = tf.assign(var, assign_placeholder)
                assign_ops.append(assign_op)
                feed_dict[assign_placeholder] = value
            self.sess.run(assign_ops, feed_dict=feed_dict)
```

This implementation is missing some things it should have -- should re-use`tf.trainable_variables()` and not create new placeholder nodes each time since memory will grow unbounded.

ml-agents/mlagents/trainers/behavior_id_utils.py
andrewcoh: fixed

ml-agents/mlagents/tf_utils/tf.py
harperj: ^- PS this is why I don't want anyone to make me delete my old branches :)

ml-agents/mlagents/tf_utils/tf.py
harperj: The future is now

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -0,0 +1,256 @@
...
+        # assign ghost's stats collection to wrapped trainer's
+        self.stats_reporter = self.trainer.stats_reporter
+
+        self_play_parameters = trainer_parameters["ghost"]
+        self.window = self_play_parameters["window"]
+        self.current_prob = self_play_parameters["current_prob"]
+        self.steps_between_snapshots = self_play_parameters["snapshot_per"]
+
+        self.policies: Dict[str, TFPolicy] = {}
+        self.policy_snapshots: List[Any] = [None for _ in range(self.window)]
```
chriselion: nit: `[None] * self.window` is equivalent for repeating a list

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -0,0 +1,256 @@
...
+        self.snapshot_counter: int = 0
+        self.learning_behavior_name: str = None
+        self.current_policy_snapshot = None
+        self.last_step = 0
+
+        self.initial_elo: float = 1200.0
```
chriselion: Is there ever a reason to want to specify another initial value? Same with "K"

chriselion: If it's arbitrary, may want to at least add some comments for us non chess players.

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -0,0 +1,256 @@
...
+        """
+        return self.trainer.get_step
+
+    def _write_summary(self, step: int) -> None:
+        """
+        Saves training statistics to Tensorboard.
+        """
+        opponents = np.array(self.policy_elos)
+        LOGGER.info(
+            " ELO: {}\n"
```
chriselion: nit: leading whitespace in string?

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -0,0 +1,256 @@
...
+            if final_reward < 0:
+                result = "loss"
+            change = compute_elo_rating_changes(
+                self.current_elo, self.policy_elos[self.current_opponent], result
+            )
+            self.current_elo += change
+            self.policy_elos[self.current_opponent] -= change
+
+    def _is_ready_update(self) -> bool:
+        pass
```
chriselion: return False? Probably equivalent but it's not doing what the type hint says.

ml-agents/mlagents/trainers/ghost/trainer.py
chriselion: Sounds like they could go in the "ghost" section of the trainer config.

ml-agents/mlagents/trainers/behavior_id_utils.py
```diff
@@ -0,0 +1,32 @@
...
+    brain_name: str
+    behavior_ids: Dict[str, int]
+
+    @staticmethod
+    def from_name_behavior_id(name_behavior_id: str) -> "BehaviorIdentifiers":
+        """
+        Parses a name_behavior_id into a BehaviorIdentifiers NamedTuple.
+        This allows you to access the brain name and distinguishing identifiers
+        without parsing more than once.
+        :param name_behavior_id: String of behavior params in HTTP format.
```
harperj: The format is the "query string".  Maybe it would be good to give an example of what this is parsing?  It wasn't clear to me and I'm familiar with HTTP query strings.

Even just `behavior_ids` -- It wasn't immediately clear which what the `str` is versus the `int` value.  I would expect a member called `behavior_ids` to be a list.

UnitySDK/Assets/ML-Agents/Examples/Tennis/Scripts/HitWall.cs
```diff
@@ -17,109 +18,111 @@ void Start()
         m_AgentB = m_Area.agentB.GetComponent<TennisAgent>();
     }
 
-    void OnTriggerExit(Collider other)
+    private void Reset()
```
surfnerd: nit: the `private` modifier is redundant in c#.

UnitySDK/Assets/ML-Agents/Examples/Tennis/Scripts/HitWall.cs
```diff
@@ -17,109 +18,111 @@ void Start()
...
                 }
                 else
                 {
-                    m_AgentA.AddReward(-0.01f);
-                    m_AgentB.SetReward(0);
-                    m_AgentB.score += 1;
+                    lastFloorHit = 0;
                 }
-            }
+            }
```
surfnerd: nit: remove whitespace

UnitySDK/Assets/ML-Agents/Examples/Tennis/Scripts/HitWall.cs
```diff
@@ -17,109 +18,111 @@ void Start()
...
-                    m_AgentA.SetReward(0);
-                    m_AgentB.AddReward(-0.01f);
-                    m_AgentA.score += 1;
+                    AgentAWins();
                 }
             }
             else if (collision.gameObject.name == "wallB")
             {
-                if (lastAgentHit == 0)
+                //Agent B hits into wall or agent A hit a winner
```
surfnerd: very nit: add a space after your comment `// Agent...`

UnitySDK/Assets/ML-Agents/Examples/Tennis/Scripts/TennisAgent.cs
```diff
@@ -85,6 +90,7 @@ public override float[] Heuristic()
         var action = new float[2];
 
         action[0] = Input.GetAxis("Horizontal");
+
```
surfnerd: nit: remove extra line

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -0,0 +1,256 @@
...
+        self.next_update_step = self.trainer.next_update_step
+        self.trainer.advance()
+        self._maybe_write_summary(self.get_step)
+
+        for q, internal_q in zip(self.policy_queues, self.internal_policy_queues):
+            # Get policies that correspond to the policy queue in question
+            try:
+                policy = cast(TFPolicy, internal_q.get_nowait())
+                with policy.graph.as_default():
+                    weights = policy.tfvars.get_weights()
```
chriselion: Regardless of how you implement the internals, I think the policy snapshot save/reload should go in TFPolicy (or base Policy?). Something like

``` python
class TFPolicySnapshot:
  def __init__(self, weights)
    # save weights or whatever else you need

class TFPolicy(Policy):
  def __init__(...):
    self.tfvars: Optional[TensorFlowVariables] = None # lazily set up if we need to save/load weights

  def save_snapshot(...) -> TFPolicySnapshot:
    if not self.tfvars:
      # set up self.tfvars
    weights = self.tfvars.get_weights()
    return TFPolicySnapshot(weights)

  def set_from_snapshot(snap: TFPolicySnapshot):
    if not self.tfvars:
      # set up tfvars, or raise?
    self.tfvars.set_weights(snap.weights)
```

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -0,0 +1,256 @@
...
+            if name_behavior_id == self.learning_behavior_name:
+                continue
+            elif np.random.uniform() < (1 - self.current_prob):
+                x = np.random.randint(len(self.policy_snapshots))
+                snapshot = self.policy_snapshots[x]
+            else:
+                snapshot = self.current_policy_snapshot
+                x = "current"
+            self.current_opponent = -1 if x == "current" else x
+            print(
```
chriselion: Use either LOGGER.info or LOGGER.debug

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -0,0 +1,256 @@
...
+
+        self.internal_policy_queues: List[AgentManagerQueue[Policy]] = []
+        self.internal_trajectory_queues: List[AgentManagerQueue[Trajectory]] = []
+
+        # assign ghost's stats collection to wrapped trainer's
+        self.stats_reporter = self.trainer.stats_reporter
+
+        self_play_parameters = trainer_parameters["ghost"]
+        self.window = self_play_parameters["window"]
+        self.current_prob = self_play_parameters["current_prob"]
```
harperj: I think this could use a different name, especially since it's surfaced as a parameter in the trainer config. "ghost_ratio" or something?

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -0,0 +1,256 @@
...
+
+        self.trainer = trainer
+
+        self.internal_policy_queues: List[AgentManagerQueue[Policy]] = []
+        self.internal_trajectory_queues: List[AgentManagerQueue[Trajectory]] = []
+
+        # assign ghost's stats collection to wrapped trainer's
+        self.stats_reporter = self.trainer.stats_reporter
+
+        self_play_parameters = trainer_parameters["ghost"]
```
harperj: These could (should?) have defaults in case they aren't specified.

ml-agents/mlagents/trainers/ghost/trainer.py
andrewcoh: True enough. Fixed.  It's never actually called since I override the advance() function.

ml-agents/mlagents/trainers/ghost/trainer.py
andrewcoh: changed here and a few lines down.

ml-agents/mlagents/trainers/ghost/trainer.py
andrewcoh: Sorry, that was for my own debugging etc. Switching to LOGGER.debug

ml-agents/mlagents/trainers/ghost/trainer.py
ervteng: Would make an optional parameter with default 1200 (e.g. use trainer_parameters.get("initial_elo", 1200)

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -0,0 +1,256 @@
...
+                internal_traj_queue.put(t)
+                self._process_trajectory(t)
+            except AgentManagerQueue.Empty:
+                pass
+
+        self.next_update_step = self.trainer.next_update_step
+        self.trainer.advance()
+        self._maybe_write_summary(self.get_step)
+
+        for q, internal_q in zip(self.policy_queues, self.internal_policy_queues):
```
ervteng: Is it possible for these two lists to be different lengths? It seems like you only create the internal policy queue if the behavior name matches right? So not all policy_queues will have a corresponding internal_policy_queue?

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -0,0 +1,256 @@
...
+        Responsible for collecting experiences and training trainer model via self_play.
+        :param trainer: The trainer of the policy/policies being trained with self_play
+        :param brain_name: The name of the brain associated with trainer config
+        :param reward_buff_cap: Max reward history to track in the reward buffer
+        :param trainer_parameters: The parameters for the trainer (dictionary).
+        :param training: Whether the trainer is set for training.
+        :param run_id: The identifier of the current run
+        """
+
+        super(GhostTrainer, self).__init__(
```
ervteng: Might be a bit late for this one - but I feel like we should rename "GhostTrainer" to "SelfPlayTrainer" and the config can have a category called `self_play`, which is more interpretable to the user than `ghost`

ml-agents/mlagents/trainers/ghost/trainer.py
ervteng: what about "play_against_current_self_ratio" ?

UnitySDK/Assets/ML-Agents/Examples/Tennis/Scripts/HitWall.cs
andrewcoh: removed.

ml-agents/mlagents/trainers/ghost/trainer.py
andrewcoh: I was hesitant to make it part of the policy itself since this is really only functionality we need because of self-play.  Most of the time, it will be code that is unused.  Is that an issue?

UnitySDK/Assets/ML-Agents/Examples/Tennis/Scripts/HitWall.cs
```diff
@@ -4,122 +4,125 @@ public class HitWall : MonoBehaviour
 {
     public GameObject areaObject;
     public int lastAgentHit;
+    public int lastFloorHit;
```
surfnerd: Could this be made into an enum to make it more clear what these numbers mean?
Perhaps something like this:
```suggestion
    public enum FloorHit
        {
            FloorAHit,
            FloorBHit,
            FloorHitUnset
        }
        
    public FloorHit lastFloorHit;
```

ml-agents/mlagents/trainers/ghost/trainer.py
andrewcoh: good catch

ml-agents/mlagents/trainers/ghost/trainer.py
andrewcoh: ELO is a weird one because it's relative to the population.  Initial ELO is supposed to be the average player strength.

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -0,0 +1,259 @@
...
+        self.policy_elos[self.snapshot_counter] = self.current_elo
+        self.snapshot_counter = (self.snapshot_counter + 1) % self.window
+
+    def swap_snapshots(self) -> None:
+        for q in self.policy_queues:
+            name_behavior_id = q.behavior_id
+            # here is the place for a sampling protocol
+            if name_behavior_id == self.learning_behavior_name:
+                continue
+            elif np.random.uniform() < (1 - self.play_against_current_self_ratio):
```
surfnerd: Is this seeded at all?  Does it matter if it is or isn't?

ml-agents/mlagents/tf_utils/tf.py
andrewcoh: Fixed this dependence.

ml-agents/mlagents/trainers/ghost/trainer.py
andrewcoh: For reproducibility, seeding would matter but for general training it shouldn't. Numpy gets seeded if a seed is specified at the beginning of a run.

ml-agents/mlagents/trainers/ghost/trainer.py
andrewcoh: @ervteng are you happy with the fix?

ml-agents/mlagents/trainers/ghost/trainer.py
chriselion: numpy gets seeded here https://github.com/Unity-Technologies/ml-agents/blob/48e21adcfed79ead46a95c37f96034abedd3328f/ml-agents/mlagents/trainers/trainer_controller.py#L67

ervteng:
Code LGTM - will approve when we have docs

docs/Training-Self-Play.md
```diff
@@ -0,0 +1,51 @@
+# Training with Self-Play
+
+ML-Agents provides the functionality to train symmetric, adversarial games with [Self-Play](https://openai.com/blog/competitive-self-play/).
```
ervteng: I think we should tell the reader (game-dev) what symmetric implies about their game and the observation/action space

docs/Training-Self-Play.md
```diff
@@ -0,0 +1,51 @@
...
+## Training with self-play
+
+Training with self-play adds additional confounding factors to the usual issues faced by reinforcement learning in general. This guide contains some discussion of the self-play hyperparameters and intuitions for tuning them.
+
+The general tradeoff is between speed/generality/stability of learning
+## Hyperparameters
+
+### Reward Signals
+
+The reward signal should still be used as described in the documentation for the other trainers and [reward signals.](Reward-Signals.md) However, we encourage users to be a bit more conservative when shaping reward functions due to the instability and non-stationarity of learning in adversarial games. Specifically, we encouraging users to begin with the simplest possible reward function (+1 winning, -1 losing) and to allow for more iterations of training to compensate for the sparsity.
```
ervteng: Specifically, we _encourage_

docs/Training-Self-Play.md
```diff
@@ -0,0 +1,51 @@
...
+## Training Statistics
+
+To view training statistics, use TensorBoard. For information on launching and
+using TensorBoard, see
+[here](./Getting-Started-with-Balance-Ball.md#observing-training-progress).
+
+### ELO
+In adversarial games, the cumulative environment reward may not be a meaningful metric by which to track learning progress.  This is because cumulative reward is entirely dependent on the skill of the opponent. An agent at a particular skill level will get more or less reward against a worse or better agent, respectively.
+
+We provide an implementation of the ELO rating system, a method for calculating the relative skill level between two players from a given population in a zero-sum game. For more informtion on ELO, please see [the ELO wiki](https://en.wikipedia.org/wiki/Elo_rating_system).
```
ervteng: We should also mention (3rd paragraph) that ELO should increase during a good training session, and that the absolute value is meaningless (or sth similar).

docs/Training-Self-Play.md
```diff
@@ -0,0 +1,51 @@
...
+Training with self-play adds additional confounding factors to the usual issues faced by reinforcement learning in general. This guide contains some discussion of the self-play hyperparameters and intuitions for tuning them.
+
+The general tradeoff is between speed/generality/stability of learning
+## Hyperparameters
+
+### Reward Signals
+
+The reward signal should still be used as described in the documentation for the other trainers and [reward signals.](Reward-Signals.md) However, we encourage users to be a bit more conservative when shaping reward functions due to the instability and non-stationarity of learning in adversarial games. Specifically, we encouraging users to begin with the simplest possible reward function (+1 winning, -1 losing) and to allow for more iterations of training to compensate for the sparsity.
+
+### Play against current self ratio
```
ervteng: I think for these hyperparams we should mention when to increase/decrease (e.g. decrease this if the ELO is unstable).

docs/Training-Self-Play.md
```diff
@@ -0,0 +1,51 @@
+# Training with Self-Play
+
+ML-Agents provides the functionality to train symmetric, adversarial games with [Self-Play](https://openai.com/blog/competitive-self-play/).
+With self-play, agents learn in adversarial games by competing against past versions of itself. Competing
```
chriselion: nit: Either "an agent learns" or "past versions of themselves"

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -0,0 +1,253 @@
...
+        self.stats_reporter.add_stat("ELO", self.current_elo)
+
+    def _process_trajectory(self, trajectory: Trajectory) -> None:
+        if (
+            trajectory.done_reached
+            and not trajectory.max_step_reached
+            and self.current_opponent > -1
+        ):
+            result = "win"
+            final_reward = trajectory.steps[-1].reward
```
chriselion: Do we ever want to factor in the magnitude of the reward and/or the opponent's reward? If not, it might be worth calling out in the docs that only the sign of the reward determines the winner.

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -0,0 +1,253 @@
...
+
+        self.internal_policy_queues: List[AgentManagerQueue[Policy]] = []
+        self.learning_policy_queues: Dict[str, AgentManagerQueue[Policy]] = {}
+        self.internal_trajectory_queues: List[AgentManagerQueue[Trajectory]] = []
+
+        # assign ghost's stats collection to wrapped trainer's
+        self.stats_reporter = self.trainer.stats_reporter
+
+        self_play_parameters = trainer_parameters["self_play"]
+        self.window = self_play_parameters["window"]
```
chriselion: Default values for these? E.g. `self.window = self_play_parameters.get("window", 10)`

Otherwise, you need some nicer error handling for when these aren't set.

ervteng: Am generally in support of default values over throwing an error

andrewcoh: Fixed in the next push.

docs/Training-Self-Play.md
```diff
@@ -15,29 +17,46 @@ Self-play is triggered by including the self-play hyperparameter hierarchy in th
...
+
+### Save Steps
+
+The `save_steps` parameter corresponds to the number of *trainer steps* between snapshots.  For example, if `save_steps`=10000 then a snapshot of the current policy will be saved every 10000 trainer steps. Note, trainer steps are counted per agent. For more information, please see the [migration doc](Migrating.md) after v0.13.
+
+A larger value of `save_steps` will yield a set of opponents that cover a wider range of skill levels and possibly playstyles since the policy receives more training. As a result, the agent will train against a variety of opponents
+### Swap Steps
+
+The `swap_steps` parameter corresponds to the number of *trainer steps* between swapping the opponents policy with a different snapshot. As in the `save_steps` discussion, notethat trainer steps are counted per agent. For more information, please see the [migration doc](Migrating.md) after v0.13.
```
ervteng: space between note and that

ml-agents/mlagents/trainers/ghost/trainer.py
andrewcoh: I called this out in the doc (still under construction).

PR #3219, opened 2020-01-14
Prevent tf.Session() from eating up all the GPU memory

ervteng:
By default, a tf.Session will take up all GPU memory. In the Policy, we combat this by enabling allow_growth = True and enabling soft placement (for multi-GPU). 

But we introduced another session to write hyperparameters. In the past this session was created after the policy, and respected those configs. However, now it is created before and needs its own config options.

### DIFF
--- a/ml-agents/mlagents/trainers/trainer.py
+++ b/ml-agents/mlagents/trainers/trainer.py
@@ -3,6 +3,7 @@ import logging
 from typing import Dict, List, Deque, Any
 
 from mlagents.tf_utils import tf
+from mlagents import tf_utils
 
 from collections import deque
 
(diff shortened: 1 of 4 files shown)


ml-agents/mlagents/trainers/trainer.py
```diff
@@ -70,7 +70,11 @@ def write_tensorboard_text(self, key: str, input_dict: Dict[str, Any]) -> None:
         :param input_dict: A dictionary that will be displayed in a table on Tensorboard.
         """
         try:
-            with tf.Session() as sess:
+            # Prevent GPU memory from being eaten up by this small writing session
+            config = tf.ConfigProto()
```
chriselion: Should we move the ConfigProto setup to a utility function? I'm ok with not doing it now, but if we need it in a 3rd place, I think we should.

ervteng: We can put it in tf_utils, that seems like the natural place for it

ervteng:
> Looks good, do we need to consider this for a hotfix release?

I think so - this and the Tensorboard fix should probably be a hotfix for 0.13. What do you think?

ml-agents/mlagents/trainers/trainer.py
ervteng: I moved it to tf in tf_utils - lmk how it looks.

Issue #3214, opened 2020-01-14
NaN in SAC training


mnsmuts:
After 92k steps an agent training to copy a demonstration starts to report:
WARNING:mlagents_envs:An agent had a NaN observation in the environment
and show this in tensorboard:
![image](https://user-images.githubusercontent.com/22680369/72335699-54edb100-36b7-11ea-8988-f31df1d6f92e.png)

Environment has 1 agent with 47 observations and 1 stacked vector, space type is continuous and action space is 4.

mlagents v13, windows 10, config file designed to imitate only:
![image](https://user-images.githubusercontent.com/22680369/72335968-c3cb0a00-36b7-11ea-89a4-489aeed11e5a.png)

I have not seen this before, could it be something to do with the upgrade to v13?

chriselion:
@mnsmuts 
If there is a NaN in your observations, you should definitely try to find out where it's coming from and fix it. I just added some better NaN checks on the C# side (https://github.com/Unity-Technologies/ml-agents/pull/3221), so if you apply those changes, you should get the callstack for where the observations are being added.

mnsmuts:
Great thanks

PR #3221, opened 2020-01-15
add NaN checks to reward and observation in C#

chriselion:
Currently we have NaN checks (and set the values to 0) on the python side. However, I think throwing sooner would help the user track down the source of the NaN.

As an example of where this would help make debugging easier: https://github.com/Unity-Technologies/ml-agents/issues/3214

### DIFF
--- a/UnitySDK/Assets/ML-Agents/Scripts/Sensor/VectorSensor.cs
+++ b/UnitySDK/Assets/ML-Agents/Scripts/Sensor/VectorSensor.cs
@@ -83,6 +83,12 @@ namespace MLAgents.Sensor
 
         void AddFloatObs(float obs)
         {
+#if DEBUG
+            if (float.IsNaN(obs))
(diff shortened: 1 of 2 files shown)


UnitySDK/Assets/ML-Agents/Scripts/Sensor/VectorSensor.cs
```diff
@@ -83,6 +83,12 @@ void Clear()
 
         void AddFloatObs(float obs)
         {
+#if DEBUG
```
chriselion: All of the Agent.AddVectorObs methods wind up here.

chriselion: We could add a flag to bypass the check if the source was an int, bool, or one-hot if that's worth optimizing for...

Issue #3201, opened 2020-01-10
What should I do when I want to continue the training ?


Thomas-6freedom:
Hello.

I have trained my AI 50k steps (the default max-steps) with this command:
`mlagents-learn config/config.yaml --run-id=first-run --train`

 And I think it hasn't learn yet. Should I relaunch the training with the same command ? Should I only change the "run-id" ? Should I erase everything and modify the max-steps to the number of steps I think is enough ?

Thank you

MarkTension:
If it's learned anything then you'd probably want to add --load to that run-id to continue where you ended, and increase max-steps. 

You say you think it hasn't learned yet after 50k steps? Maybe you want to check if it learned anything at all by checking the progress on tensorboard. If it learned to increase its cumulative reward to some degree then you might want to continue training. If nothing at all happened after 50k steps then you might have some error and you'd probably have to make some changes to your hyperparameter settings or environment if it's a custom one.

Thomas-6freedom:
Thank you. I tried to add --load on my second training. But in tensorboard, the graph restart from the left so it's really messy to read anything

MarkTension:
yeah that's pretty messy isn't it? It also messes up the relative time. You'll get all the info, but i've also not found a way to deal with that yet

dlindmark:
If I recall correctly, the tensorboard graph continued for mlagents version < 0.13.0. I prefered it that way, since it makes it for cleaner graphs.

PR #3223, opened 2020-01-15
Get step from policy

ervteng:
Fix for #3201 

Get current trainer steps from the policy checkpoint when loaded. Note: ideally we'd move the function into a common file (e.g. RLTrainer) but this is a minimal fix for hotfix.


Dastyn:
Hi,
Is there a way to change (increase) the learning rate between 2 --load ?
Thanks!
D.

andrewcoh:
Do you mean to stop training, change the learning rate and then resume training from the same point using --load?

Dastyn:
Yes, exactely! 
Besides, this might be useful to be able to increase this learning rate between two following curriculum lessons (without hacking Python API ;) ).

PR #3232, opened 2020-01-15
inference - fill 0s for done Agents

chriselion:


UnitySDK/Assets/ML-Agents/Scripts/InferenceBrain/TensorProxy.cs
```diff
@@ -90,6 +90,23 @@ public static TensorProxy TensorProxyFromBarracuda(Tensor src, string nameOverri
             };
         }
 
+        public static void FillTensorBatch(TensorProxy tensorProxy, int batch, float fillValue)
```
chriselion: Could do this as a method on TensorProxy, but it follows the convention in FillTensorWithRandomNormal (below)

PR #3233, opened 2020-01-15
Bugfix: Fix lost trajectories when they are produced faster than they are consumed

ervteng:
We were only consuming one trajectory at most during each `trainer.advance()` call. Some environments with a very short time horizon could produce many trajectories per step (e.g. GridWorld). This meant many trajectories were going unused (and, in fact, caused a memory leak). 

This change forces all trajectories to be consumed each time `trainer.advance()` is called. ~In a future multithreaded implementation, we will most likely need to cap the length of the AgentManagerQueues (and scale the # of environments respectively) to avoid this issue, but this fix is suitable for the single-threaded case.~

### DIFF
--- a/ml-agents/mlagents/trainers/agent_processor.py
+++ b/ml-agents/mlagents/trainers/agent_processor.py
@@ -192,12 +192,13 @@ class AgentManagerQueue(Generic[T]):
 
         pass
 
-    def __init__(self, behavior_id: str):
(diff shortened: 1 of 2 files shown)


ervteng:
> Might want to eventually expose the queue parameter, but looks good.

If we ever build a multithreaded trainer_controller, we'll pass the queue parameter on creation of the AgentManager

PR #3239, opened 2020-01-16
convert yamato tests to python

chriselion:
No immediate benefit, but should be easier to modify when we want to do more complicated tests like training and swapping in models.


ml-agents/tests/yamato/editmode_tests.py
```diff
@@ -0,0 +1,114 @@
...
+    failed: str
+    duration: str
+
+
+def parse_results(results_xml):
+    """
+    Extract the test results from the xml file.
+    """
+    stats = {}
+    dom_tree = xml.dom.minidom.parse(results_xml)
```
chriselion: Adapted from the beginning of https://answers.unity.com/questions/1355795/can-unity-56s-test-runner-optionally-output-nunit.html

.yamato/csharp-tests.yml
```diff
@@ -12,12 +12,16 @@ test_mac_editmode_{{ editor.version }}:
   variables:
     UNITY_VERSION: {{ editor.version }}
   commands:
-    - ./run-tests-editmode-osx-editor.sh
+    - python -u ml-agents/tests/yamato/editmode_tests.py
```
chriselion: `-u` means "unbuffered output", otherwise we get all the python output after the unity output.

PR #3241, opened 2020-01-16
Refactor file logic in demo_loader and add unit tests.

kevtan:


kevtan:
I didn't know how to revert my changes and thought that creating a new pull request would be a safer option.

chriselion:
Note: migrated from https://github.com/Unity-Technologies/ml-agents/pull/3149

PR #3283, opened 2020-01-24
Fix base port in ipynb example

chriselion:


Jeffhere:
env_name = "../Project/Assets/ML-Agents/Examples/3DBall" 
env = UnityEnvironment(file_name=env_name, side_channels = [engine_configuration_channel])

UnityEnvironmentException: Couldn't launch the 3DBall environment. Provided filename does not match any environments.
I got this error，is anyone know how to solve it? Is env_name'path wrroy?

PR #3299, opened 2020-01-28
Convert pyramids to Raycast sensor

chriselion:
Removing the last usage of RayPerception (the non-sensor-based approach) from the examples and retrain the network.

After this and https://github.com/Unity-Technologies/ml-agents/pull/3297 we can remove the deprecated code.


com.unity.ml-agents/csharp_timers.json.meta
```diff
@@ -1,7 +0,0 @@
-fileFormatVersion: 2
```
chriselion: This probably shouldn't have gotten added, might need to add it to .gitignore

Project/Assets/ML-Agents/Examples/Pyramids/Scripts/PyramidAgent.cs
```diff
@@ -19,23 +18,13 @@ public override void InitializeAgent()
...
-        m_RayPer = GetComponent<RayPerception>();
         m_SwitchLogic = areaSwitch.GetComponent<PyramidSwitch>();
     }
 
     public override void CollectObservations()
     {
         if (useVectorObs)
         {
-            const float rayDistance = 35f;
-            float[] rayAngles = { 20f, 90f, 160f, 45f, 135f, 70f, 110f };
```
chriselion: I matched the maxRayDegrees in the new component to the min/max of the old ones, but the exact distribution won't be the same.

surfnerd: does it need to be retrained?

chriselion: Yes, new Pyramids.nn file is in PR. I ran it last night.

PR #3304, opened 2020-01-29
remove deprecated RayPerception, legacy codepath

chriselion:
The RayPerception MonoBehaviour (and its subclasses) have been deprecated for a few releases now. This PR removes them, and also the legacy hitfraction behavior that was kept for compatibility.


chriselion:
Will update migration guide later

Issue #2595, opened 2019-09-19
IL2CPP build fails


wightwhale:
**Describe the bug**
IL2CPP build fails in Unity 2018.4.9f1

**To Reproduce**
Download and import ML-Agents .9 import unity package into project
Edit->Project Settings->Player-> Scripting Backend_>IL2CPP

**Console logs / stack traces**
Please wrap in [triple backticks (```)](https://help.github.com/en/articles/creating-and-highlighting-code-blocks) to make it easier to read.
```Failed running C:\Program Files\Unity\Hub\Editor\2018.4.9f1\Editor\Data\il2cpp/build/il2cpp.exe --convert-to-cpp --emit-null-checks --enable-array-bounds-check --dotnetprofile="unityaot" --compile-cpp --libil2cpp-static --platform="WindowsDesktop" --architecture="x64" --configuration="Release" --outputpath="C:\Users\\Documents\ML Test 4\Temp/StagingArea/Data\Native\GameAssembly.dll" --cachedirectory="C:\Users\\Documents\ML Test 4\Assets\..\Library/il2cpp_cache" --map-file-parser="C:\Program Files\Unity\Hub\Editor\2018.4.9f1\Editor\Data\Tools\MapFileParser\MapFileParser.exe" --directory="C:\Users\\Documents\ML Test 4\Temp\StagingArea\Data\Managed" --generatedcppdir="C:\Users\\Documents\ML Test 4\Temp\StagingArea\Data\il2cppOutput" 

stdout:
Building GameAssembly.dll with MsvcDesktopToolChain
	Msvc Install Version: 15.0
	Msvc Install SDK Directory: C:\Program Files (x86)\Windows Kits\10
	Msvc Linker Path: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\bin\HostX64\x64\link.exe
	Msvc Compiler Path: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\MSVC\14.16.27023\bin\HostX64\x64\cl.exe

	Output directory: C:\Users\\Documents\ML Test 4\Temp\StagingArea\Data\Native
	Cache directory: C:\Users\\Documents\ML Test 4\Library\il2cpp_cache
ObjectFiles: 598 of which compiled: 149
	Time Compile: 43679 milliseconds Bulk_mscorlib_5.cpp
[...]

PR #3315, opened 2020-01-29
Fix Windows IL2CPP compilation with dlopen/dlsym stubs

AcelisWeaven:
This pull request fixes and issue with IL2CPP and Windows, where GRPC needs the `dlopen` and `dlsym` functions. (see #2595)

Since we don't want to recompile GRPC from scratch or include it fully, the solution here is to make stubs for the missing functions. The .meta makes it so it's only compiled on Windows by IL2CPP.

This sounds like an ugly hack and may not be wanted. Thoughts?

_Help wanted:_ I don't really know how to properly name this file. Any suggestion?

### DIFF
--- a/com.unity.ml-agents/Plugins/IL2CPP.DL.Stubs.c
+++ b/com.unity.ml-agents/Plugins/IL2CPP.DL.Stubs.c
@@ -1,4 +1,5 @@
 // These stubs fix an issue compiling GRPC on Windows with IL2CPP.
+// For the moment, only Inference works. (training doesn't)
 
 void * dlopen(const char *filename, int flags) {
     return 0;
--- a/com.unity.ml-agents/Plugins/IL2CPP.DL.Stubs.c
+++ b/com.unity.ml-agents/Plugins/IL2CPP.DL.Stubs.c
@@ -1,3 +1,5 @@
+// These stubs fix an issue compiling GRPC on Windows with IL2CPP.
+
 void * dlopen(const char *filename, int flags) {
     return 0;
 }
--- a/com.unity.ml-agents/Plugins/IL2CPP.DL.Stubs.c
+++ b/com.unity.ml-agents/Plugins/IL2CPP.DL.Stubs.c
@@ -1,0 +1,7 @@
+void * dlopen(const char *filename, int flags) {
+    return 0;
+}
+
+void * dlsym(void *handle, const char *symbol) {
+    return 0;
+}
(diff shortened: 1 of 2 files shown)


surfnerd:
Hey @AcelisWeaven,
Thanks for your contribution.  Does this break any other behavior on windows?  Does training still work in the editor with this?  Could you provide a little more information on how you tested this?  
Thanks!
Chris

AcelisWeaven:
> Does training still work in the editor with this?

Yes it does. There should be no issue here, the file is marked as Standalone-only.
image

> Does this break any other behavior on windows?

Sadly there is an issue I forgot to mention : The compilation works, but when running as a standalone, it runs in Inference mode.
`Couldn't connect to trainer on port 5005 using API version API-14-dev0. Will perform inference instead.`
Switching back to Mono and it works. But I'm not sure the issue is related to this PR. Also, back on v0.8, I had a similar issue because of code stripping being enable (I don't know if this has been solved). With IL2CPP is can only set it to Low or higher. I've been trying IL2CPP Release and Debug.

> Could you provide a little more information on how you tested this?

I was able to compile both a custom environment and a freshly-cloned Tennis (both envs are similar), whereas I couldn't before. Both only work in Editor mode only with IL2CPP.

Here's the command I use for in-editor training:
```
```
Standalone training:
```
mlagents-learn.exe trainer_config.yaml --train --env=MyEnv --save-freq=10000 --run-id=myrun
```

TLDR: Inference and training in editor works, training with a standalone build doesn't :)

surfnerd:
Hey @AcelisWeaven,
Thanks for the information!  Very helpful. I will bring this up with the team and consider merging it.

surfnerd:
Hey @AcelisWeaven,
Can you update the comment to make it clear that training still doesn't work with this change?  After that, I will push a branch to our repo with your commit and merge it.

AcelisWeaven:
@surfnerd Done! Thanks.

caioc2:
Hi,

As someone who have tried out stubs for this problem, here are my 2 cents:

Stubs does not work when building the standalone to run multiple environments for training, which is the main point of ever using IL2CPP, **speed up training**.
Using it in the editor makes not much sense, since the Unity editor uses  Mono backend.

While the IL2CPP is a much needed fix, I think this is not the fix. It is a difficult issue which may be only solved with having directives in the GRPC and or some other clever idea to be found.

surfnerd:
Hi @caioc2,
I agree that this doesn't help with the training aspect of this problem. Yet it would still be beneficial to anyone who is running inference and wants to build for IL2CPP.  You still get performance benefits by building with IL2CPP.  We fully understand that this doesn't fix the training issue, which is why we asked for the extra comment.  

That being said, we will be spending more time fixing the IL2CPP training issue.  Thanks for you comment.

surfnerd:
Moved the PR here #3330

PR #3332, opened 2020-01-31
testing: exit playback after specified number of episodes

chriselion:
Update ModelOverrider to quit after a certain number of episodes are run (steps > maxEpisodes * agent.maxSteps). This lets us output inference rewards during testing.

Example commandline usage:
```shell
Project/Builds/ball3d.app/Contents/MacOS/ball3d --mlagents-override-model 3DBall ~/code/ml-agents/Project/Assets/ML-Agents/Examples/3DBall/TFModels/3DBall.nn --mlagents-quit-after-episodes 1 --no-graphics -logFile -
```

### DIFF
--- a/Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ModelOverrider.cs
+++ b/Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ModelOverrider.cs
@@ -9,20 +9,33 @@ namespace MLAgentsExamples
 {
     /// <summary>
     /// Utility class to allow the NNModel file for an agent to be overriden during inference.
-    /// This is useful to validate the file after training is done.
+    /// This is used internally to validate the file after training is done.
     /// The behavior name to override and file path are specified on the commandline, e.g.
     /// player.exe --mlagents-override-model behavior1 /path/to/model1.nn --mlagents-override-model behavior2 /path/to/model2.nn
+    ///
+    /// Additionally, a number of episodes to run can be specified; after this, the application will quit.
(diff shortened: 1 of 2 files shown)


Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ModelOverrider.cs
```diff
@@ -32,27 +45,56 @@ void GetAssetPathFromCommandLine()
...
             {
                 OverrideModel();
             }
         }
 
+        void FixedUpdate()
+        {
+            if (m_MaxEpisodes > 0)
+            {
+                if (m_NumSteps > m_MaxEpisodes * m_Agent.maxStep)
```
chriselion: I want to make sure we run at least `Agent.maxSteps` steps, so even if an Agent finishes early we keep running at least a full episode for the rest. This approach made the most sense to me, so that if maxSteps changes in an example, we don't need to change any other config. Open to other ideas though.

com.unity.ml-agents/Runtime/Timer.cs
```diff
@@ -151,12 +151,16 @@ public void Begin()
...
         /// Stop timing a block of code, and increment internal counts.
         /// </summary>
-        public void End()
+        public void End(bool isRecording)
         {
-            var elapsed = DateTime.Now.Ticks - m_TickStart;
-            m_TotalTicks += elapsed;
-            m_TickStart = 0;
-            m_NumCalls++;
+            if (isRecording)
```
chriselion: This was a bit hacky - we don't want Agents to send their rewards to the timers when we're shutting down (since a bunch will have approx 0 reward).

Issue #3321, opened 2020-01-30
RayPerceptionSensorComponent3D does not function as expected when the agent's scale is not 1.


WanNJ:
All the agents in the example scenes have a scale of 1. However, if changed the scale to other numbers, the Debug gizmo would also scale, resulting in the wrong rays drawn in the editor.

I haven't checked whether the observed values are correct yet. Presumably, it may have the same issue.

- ML-Agents v0.13.0
- TensorFlow v2.0
- Environment_: Change any of the agent's scale to not equal to 1.

To reproduce this bug, just change the scale of the agent of WallJump example and run using the trained model. You will see that the rays drawn in the editor is wrong.

chriselion:
I'll look into this tomorrow.

Can you explain more what are you seeing and what do you expect? I would expect that when the scale changes, the length of the rays also increase.

WanNJ:
> I'll look into this tomorrow.
> 
> Can you explain more what are you seeing and what do you expect? I would expect that when the scale changes, the length of the rays also increase.

The expected: the red sphere drawn representing the hit point should be the actual hitpoint. However, when the scale changes, the red sphere is drawn to somewhere else, scaling with the scale of the agent. However, I don't know if the observation feed into the RL component is corrent or not.
<img width="1440" alt="Screen Shot 2020-01-30 at 10 59 42" src="https://user-images.githubusercontent.com/17922409/73466369-c718f380-434f-11ea-85c8-debe2693be59.png">

chriselion:
Thanks, the screenshot helps a lot. Looking into it now, I get the feeling the scale is being applied twice somehow.

chriselion:
OK, I think I know what's going on. I believe this is just a visual bug, but it's definitely something we should fix.
When we do the raycasting here:
https://github.com/Unity-Technologies/ml-agents/blob/4ba8b973369435fc76fc67b32a753536d2f126c9/com.unity.ml-agents/Runtime/Sensor/RayPerceptionSensor.cs#L219-L228
we cast up to `rayLength` units, and save the hit fraction. This fraction is based on `rayLength` units.

When doing the gizmo drawing here
https://github.com/Unity-Technologies/ml-agents/blob/4ba8b973369435fc76fc67b32a753536d2f126c9/com.unity.ml-agents/Runtime/Sensor/RayPerceptionSensorComponentBase.cs#L127-L128
[...]

PR #3334, opened 2020-02-01
Make RayPerception sensor work better with transforms that have scale

chriselion:
Fixes https://github.com/Unity-Technologies/ml-agents/issues/3321

Thanks @WanNJ for reporting.

This makes the Agent's transform scale behave more sensibly for raycasts; the ray length and sphere sizes are also scaled. It handles non-uniform scale reasonably, by using the length of the transforms ray to adjust these.

Also adds a few checks for division by zero.

Note that existing trained models for agents that have scale will need to be retrained, since the hit fractions will differ at runtime.

image


com.unity.ml-agents/Runtime/Sensor/RayPerceptionSensorComponentBase.cs
```diff
@@ -130,14 +130,14 @@ public void OnDrawGizmos()
                 // hit fraction ^2 will shift "far" hits closer to the hit color
                 var lerpT = rayInfo.hitFraction * rayInfo.hitFraction;
                 var color = Color.Lerp(rayHitColor, rayMissColor, lerpT);
-                color.a = alpha;
+                color.a *= alpha;
```
chriselion: Ran into this when trying to get a screenshot; I wanted to hide the "higher" raycast component by setting its alpha to 0

PR #3347, opened 2020-02-03
Removing general questions and installation/setup template

anupambhatnagar:
ML Agents users are creating github issues using the templates that we provide. In order for the github issues to be limited to bugs and feature requests I suggest removing the templates for general questions and installation/setup.


chriselion:
I updated the config.yml so that it _should_ keep entries for `Discussion / General Questions` and `Installation / Setup` but link to the forum instead.

PR #3353, opened 2020-02-04
make some more classes and methods internal

chriselion:


com.unity.ml-agents/Runtime/Policy/RemotePolicy.cs
```diff
@@ -15,7 +15,7 @@ internal class RemotePolicy : IPolicy
         int m_AgentId;
         string m_FullyQualifiedBehaviorName;
 
-        protected ICommunicator m_Communicator;
+        internal ICommunicator m_Communicator;
```
chriselion: Hmm, maybe this doesn't matter if the whole class is internal now.

surfnerd: seems ok

PR #3392, opened 2020-02-07
csharp cleanup

chriselion:
Apply a lot (but not 100%) of Rider's suggestions.

### DIFF
--- a/Project/Assets/ML-Agents/Examples/Soccer/Scripts/AgentSoccer.cs
+++ b/Project/Assets/ML-Agents/Examples/Soccer/Scripts/AgentSoccer.cs
@@ -26,15 +26,14 @@ public class AgentSoccer : Agent
     [HideInInspector]
     public Rigidbody agentRb;
     SoccerSettings m_SoccerSettings;
-    Renderer m_AgentRenderer;
-    BehaviorParameters m_BP;
+    BehaviorParameters m_BehaviorParameters;
     Vector3 m_Transform;
 
     public override void InitializeAgent()
     {
         base.InitializeAgent();
-        m_BP = gameObject.GetComponent<BehaviorParameters>();
-        if (m_BP.m_TeamID == (int)Team.Blue)
+        m_BehaviorParameters = gameObject.GetComponent<BehaviorParameters>();
+        if (m_BehaviorParameters.m_TeamID == (int)Team.Blue)
         {
             team = Team.Blue;
             m_Transform = new Vector3(transform.position.x - 4f, .5f, transform.position.z);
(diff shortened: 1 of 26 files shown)


Project/Assets/ML-Agents/Examples/Soccer/Scripts/AgentSoccer.cs
```diff
@@ -26,15 +26,14 @@ public enum Team
     [HideInInspector]
     public Rigidbody agentRb;
     SoccerSettings m_SoccerSettings;
-    Renderer m_AgentRenderer;
```
chriselion: @andrewcoh A few unused variables from soccer. Just want to make sure there's no plans to bring these back.

com.unity.ml-agents/Runtime/Policy/BehaviorParameters.cs
```diff
@@ -37,7 +36,7 @@ private enum BehaviorType
         string m_BehaviorName = "My Behavior";
         [HideInInspector]
         [SerializeField]
-        public int m_TeamID = 0;
+        public int m_TeamID;
```
chriselion: This and m_useChildSensors are still public with `m_` prefixes. Not sure if we should rename or make internal. `m_TeamID` seems like it needs to be mutable at runtime.

com.unity.ml-agents/Runtime/Policy/BehaviorParameters.cs
```diff
@@ -93,13 +92,13 @@ internal IPolicy GeneratePolicy(Func<float[]> heuristic)
         }
 
         public void GiveModel(
-            string behaviorName,
+            string newBehaviorName,
```
chriselion: Argument name shadowed the `behaviorName` property

com.unity.ml-agents/Tests/Editor/EditModeTestInternalBrainTensorApplier.cs
```diff
@@ -12,12 +11,7 @@ public class EditModeTestInternalBrainTensorApplier
     {
         class TestAgent : Agent
         {
-            public AgentAction GetAction()
```
chriselion: Agent has a GetAction now, and this wasn't actually being used.

Project/Assets/ML-Agents/Examples/Soccer/Scripts/AgentSoccer.cs
andrewcoh: Ah I missed this. Good to remove.

PR #3404, opened 2020-02-10
fix BehaviorParametersEditor for m_UseChildSensors

chriselion:
Missed this rename in https://github.com/Unity-Technologies/ml-agents/pull/3395


surfnerd:
does it need the `PreviouslySerializedAs` attribute?

chriselion:
That was added by Rider: https://github.com/Unity-Technologies/ml-agents/blob/566770f877092e23a408ab55e8cc6c62a390d1ce/com.unity.ml-agents/Runtime/Policy/BehaviorParameters.cs#L41

PR #3413, opened 2020-02-11
Adding new models

anupambhatnagar:
Adding new .nn files


chriselion:
We need to merge https://github.com/Unity-Technologies/ml-agents/pull/3414 and retrain the FoodCollecter model

PR #3418, opened 2020-02-12
C# and Python checks for infinity and NaN.

surfnerd:
DebugAssert in C# if we are getting NaNs or Infs.  Raise in Python if we are receiving NaNs or Infs.

### DIFF
--- a/ml-agents-envs/mlagents_envs/rpc_utils.py
+++ b/ml-agents-envs/mlagents_envs/rpc_utils.py
@@ -117,6 +117,9 @@ def _raise_on_nan_and_inf(data: np.array, source: str) -> np.array:
     # See https://stackoverflow.com/questions/6736590/fast-check-for-nan-in-numpy for background
     # Note that a very large values (larger than sqrt(float_max)) will result in an Inf value here
     # This is OK though, worst case it results in an unnecessary (but harmless) nan_to_num call.
+    if data.size == 0:
+        return data
+
     d = np.mean(data)
     has_nan = np.isnan(d)
     has_inf = not np.isfinite(d)
--- a/ml-agents-envs/mlagents_envs/rpc_utils.py
+++ b/ml-agents-envs/mlagents_envs/rpc_utils.py
@@ -110,6 +110,23 @@ def _process_visual_observation(
     return np.array(batched_visual, dtype=np.float32)
 
 
+def _raise_on_nan_and_inf(data: np.array, source: str) -> np.array:
+    # Check for NaNs or infs in the observations
+    # If there's a NaN in the observations, the np.mean() result will be NaN
+    # If there's an Inf (either sign) then the result will be Inf
+    # See https://stackoverflow.com/questions/6736590/fast-check-for-nan-in-numpy for background
+    # Note that a very large values (larger than sqrt(float_max)) will result in an Inf value here
+    # This is OK though, worst case it results in an unnecessary (but harmless) nan_to_num call.
+    d = np.mean(data)
+    has_nan = np.isnan(d)
+    has_inf = not np.isfinite(d)
+
+    if has_nan:
+        raise RuntimeError(f"The {source} provided had NaN values.")
+    if has_inf:
+        raise RuntimeError(f"The {source} provided had Infinite values.")
+
+
 @timed
 def _process_vector_observation(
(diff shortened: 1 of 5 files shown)


ml-agents-envs/mlagents_envs/rpc_utils.py
```diff
@@ -110,6 +110,23 @@ def _process_visual_observation(
...
+    # If there's a NaN in the observations, the np.mean() result will be NaN
+    # If there's an Inf (either sign) then the result will be Inf
+    # See https://stackoverflow.com/questions/6736590/fast-check-for-nan-in-numpy for background
+    # Note that a very large values (larger than sqrt(float_max)) will result in an Inf value here
+    # This is OK though, worst case it results in an unnecessary (but harmless) nan_to_num call.
+    d = np.mean(data)
+    has_nan = np.isnan(d)
+    has_inf = not np.isfinite(d)
+
+    if has_nan:
```
ervteng: Should we raise this error or just zero it out and print a warning? Some envs (e.g. Crawler) will produce occasional NaNs but the training will recover eventually.

ml-agents-envs/mlagents_envs/rpc_utils.py
```diff
@@ -110,6 +110,23 @@ def _process_visual_observation(
...
 
 
+def _raise_on_nan_and_inf(data: np.array, source: str) -> np.array:
+    # Check for NaNs or infs in the observations
+    # If there's a NaN in the observations, the np.mean() result will be NaN
+    # If there's an Inf (either sign) then the result will be Inf
+    # See https://stackoverflow.com/questions/6736590/fast-check-for-nan-in-numpy for background
+    # Note that a very large values (larger than sqrt(float_max)) will result in an Inf value here
+    # This is OK though, worst case it results in an unnecessary (but harmless) nan_to_num call.
+    d = np.mean(data)
```
ervteng: If data is empty (e.g. no vector obs but visual obs) then the mean is a NaN, and it will trigger the exception

ml-agents-envs/mlagents_envs/rpc_utils.py
chriselion: I don't think we should cover it up, since the NaNs will cause problems at inference too.

ml-agents-envs/mlagents_envs/rpc_utils.py
chriselion: I don't _think_ that can happen (if you only have visual obs, we shouldn't send an empty array for the "vector observations") but maybe worth checking for.

ervteng: This is currently happening in the tests - not sure if it's limited to the demonstration loader or if it can happen elsewhere too. Might also be an incompatibility between the old demos and the new BatchedStepResult based API.

surfnerd: so is the consensus here to check for empty, and early out if it is?

ervteng: That's my vote. @chriselion?

chriselion: Sounds reasonable.

surfnerd: alright, thanks.

ml-agents-envs/mlagents_envs/rpc_utils.py
```diff
@@ -110,6 +110,26 @@ def _process_visual_observation(
     return np.array(batched_visual, dtype=np.float32)
 
 
+def _raise_on_nan_and_inf(data: np.array, source: str) -> np.array:
+    # Check for NaNs or infs in the observations
+    # If there's a NaN in the observations, the np.mean() result will be NaN
+    # If there's an Inf (either sign) then the result will be Inf
+    # See https://stackoverflow.com/questions/6736590/fast-check-for-nan-in-numpy for background
+    # Note that a very large values (larger than sqrt(float_max)) will result in an Inf value here
```
ervteng: This comment isn't correct anymore right? It will throw a RuntimeError

surfnerd: yes, updated it.

PR #3438, opened 2020-02-13
Raise Exception if path does not exist [Bug Fix]

alphonsocrawford:
If you run mlagents-learn with a non-existent environment path, it returns 0; after this fix, an error will be thrown with a non-zero error code.


ml-agents/mlagents/trainers/learn.py
```diff
@@ -388,6 +390,14 @@ def create_environment_factory(
     env_args: Optional[List[str]],
 ) -> Callable[[int, List[SideChannel]], BaseEnv]:
     if env_path is not None:
+        if not path.exists(env_path):
```
harperj: We actually don't launch the environment using the exact path -- there is some slightly more involved logic.  Maybe you could refactor `UnityEnvironment` slightly to make sure we use the same path here when checking for existence as is used in `executable_launcher(...)`: https://github.com/Unity-Technologies/ml-agents/blob/master/ml-agents-envs/mlagents_envs/environment.py#L152

alphonsocrawford: Maybe I'm a bit confused. Isn't this the same `env_path` that is within the `env_factory` and passed as input to the `SubprocessEnvManager`?

chriselion: The argument to `--env` might not be an actual file. If I build the scene as "Project/Build/grid" and run 
> `mlagents-learn config/trainer_config.yaml --train --env=Project/Build/grid`

it works. But `Project/Build/grid` doesn't exist.
```
$ ls Project/Build/grid
ls: Project/Build/grid: No such file or directory
$ ls Project/Build/grid.app/
Contents
```

ml-agents-envs/mlagents_envs/environment.py
```diff
@@ -147,55 +147,64 @@ def __init__(
     def get_communicator(worker_id, base_port, timeout_wait):
         return RpcCommunicator(worker_id, base_port, timeout_wait)
 
-    def executable_launcher(self, file_name, docker_training, no_graphics, args):
+    @staticmethod
+    def environment_launch_check(env_path):
```
chriselion: Typehints:
```suggestion
    def environment_launch_check(env_path: str) -> Optional[str]:
```

ml-agents-envs/mlagents_envs/environment.py
```diff
@@ -147,55 +147,64 @@ def __init__(
...
+        if not (glob.glob(env_path) or glob.glob(env_path + ".*")):
```
chriselion: Is this necessary, or just an early-out?

ml-agents-envs/mlagents_envs/environment.py
```diff
@@ -147,55 +147,64 @@ def __init__(
...
+            raise UnityEnvironmentException(
```
chriselion: Seems like just returning `None` instead of is more consistent, since you check for None in the calling code.

ml-agents-envs/mlagents_envs/environment.py
harperj: minor / subjective: maybe this method could be called `validate_environment_path`

ml-agents/mlagents/trainers/learn.py
```diff
@@ -396,6 +397,12 @@ def create_environment_factory(
             .replace(".x86_64", "")
             .replace(".x86", "")
         )
+        launch_string = UnityEnvironment.environment_launch_check(env_path)
```
chriselion: This isn't quite the same check as in `executable_launcher()`, since we don't do the `strip()` and replace the extensions - should that be moved into `environment_launch_check`?

ml-agents-envs/mlagents_envs/environment.py
```diff
@@ -147,55 +147,64 @@ def __init__(
...
+
+    def executable_launcher(self, file_name, docker_training, no_graphics, args):
+        file_name = (
+            file_name.strip()
+            .replace(".app", "")
+            .replace(".exe", "")
+            .replace(".x86_64", "")
+            .replace(".x86", "")
+        )
+        true_filename = os.path.basename(os.path.normpath(file_name))
```
harperj: Could you move this line and the one above into `environment_launch_check`?  I'm not 100% sure it's necessary, but seems like it is more of the logic around validating the path.

ml-agents-envs/mlagents_envs/environment.py
chriselion: Maybe rename to something like `env_path_to_launch_string`?

ml-agents/mlagents/trainers/learn.py
```diff
@@ -396,6 +397,12 @@ def create_environment_factory(
...
+        if launch_string is None:
+            raise UnityEnvironmentException(
```
harperj: Perhaps add a test that this exception is raised.

ml-agents-envs/mlagents_envs/environment.py
alphonsocrawford: Changed

alphonsocrawford: I renamed it to `validate_enviornment_path` per @harperj's request, I think validate_environment_path encapsulates the intention of this function

ml-agents-envs/mlagents_envs/environment.py
alphonsocrawford: I thought of it as an early out, ensuring there is a file in the first place before checking if it is launch-able

ml-agents-envs/mlagents_envs/environment.py
alphonsocrawford: That's a solid point, changing

ml-agents/mlagents/trainers/learn.py
```diff
@@ -395,14 +396,12 @@ def create_environment_factory(
...
-            env_path.strip()
-            .replace(".app", "")
-            .replace(".exe", "")
-            .replace(".x86_64", "")
-            .replace(".x86", "")
-        )
+        launch_string = UnityEnvironment.validate_environment_path(env_path)
+        if launch_string is None:
+            raise UnityEnvironmentException(
+                "Couldn't launch the {0} environment. "
```
chriselion: Small nit: I'd like to move away from `str.format()` like this for new code, and use [f-strings](https://realpython.com/python-f-strings/) when possible. This could be written instead as
```python
raise UnityEnvironmentException(
    f"Couldn't launch the {env_path} environment. Provided filename does not match any environments."
)
```
We're not enforcing this, so feel free to ignore.

harperj:
Looks like this just failed on formatting.  You can have the pre-commit hooks automatically run locally by running:
```
pip install pre-commit
pre-commit install
```
 from your ML-Agents repo root.

PR #3439, opened 2020-02-14
[yamato] Use triggers to execute tests for specific changes.

surfnerd:
Only trigger package tests if the package, or its test template was changed.

### DIFF
--- a/.yamato/standalone-build-test.yml
+++ b/.yamato/standalone-build-test.yml
@@ -15,19 +15,13 @@ test_mac_standalone_{{ editor.version }}:
     - pip install pyyaml
     - python -u -m ml-agents.tests.yamato.standalone_build_tests
   triggers:
-    pull_requests:
-      - targets:
-          only:
-            - "master"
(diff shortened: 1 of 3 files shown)


.yamato/standalone-build-test.yml
```diff
@@ -15,10 +15,13 @@ test_mac_standalone_{{ editor.version }}:
...
-          only:
-            - "master"
-            - "/release-.*/"
-            - "/hotfix-.*/"
+    changes:
+      only:
+        - "com.unity.ml-agents/**"
+        - "Project/**"
+        - ".yamato/standalone-build-test.yml"
+      exept:
```
chriselion: exept or except?

chriselion: Looks like `except` according to https://internaldocs.hq.unity3d.com/yamato/yaml/triggers/#file-changes

surfnerd: Whoops, typo

PR #3451, opened 2020-02-14
[bug-fix] Empty ignored trajectory queues, make sure queues don't overflow

ervteng:
# Proposed change(s)

The current GhostTrainer subscribes to all trajectory queues (both learning and ghost) but doesn't empty the ghost queues as it doesn't use those trajectories for learning. This leads to a memory leak, at least until the queue is full (1000 trajectories, for Soccer it was about 5 GB). 

This PR makes sure the GhostTrainer empties all queues (even if they are ignored) and also ensures queues are emptied even if they are filling faster than advance() is called (rare). 

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist
- [x] I have added updated the changelog (if applicable)
- [ ] I have added necessary documentation (if applicable) - Not Applicable
- [ ] I have updated the migration guide (if applicable) - Not Applicable

### DIFF
--- a/com.unity.ml-agents/CHANGELOG.md
+++ b/com.unity.ml-agents/CHANGELOG.md
@@ -18,7 +18,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
  - Update Barracuda to 0.6.0-preview
 
 ### Bugfixes
-
+- Fixed an issue which caused self-play training sessions to consume a lot of memory. (#3451)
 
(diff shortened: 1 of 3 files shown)


chriselion:
Are we considering this for a hotfix release?

ervteng:
> Are we considering this for a hotfix release?

Yes; probably would want to bundle it with other fixes though.

PR #3472, opened 2020-02-19
Make the timer output format consistent

chriselion:
### Proposed change(s)

This changes the output format of the python timers. Instead of children nodes as a list with the child name in the node, the child names are keys in the dictionary. This makes the timers consistent with the C# timers (which are less flexible in the output format because we're using C#'s serialization).

It also adds new gauges when we report stats (e.g. policy value estimates). This should let us estimate the ELO rating after training.

https://jira.unity3d.com/browse/MLA-661

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents-envs/mlagents_envs/timers.py
+++ b/ml-agents-envs/mlagents_envs/timers.py
@@ -183,16 +183,16 @@ class TimerStack:
             res["is_parallel"] = True
 
         child_total = 0.0
-        child_list = {}
+        child_dict = {}
         for child_name, child_node in node.children.items():
             child_res: Dict[str, Any] = self.get_timing_tree(child_node)
-            child_list[child_name] = child_res
+            child_dict[child_name] = child_res
             child_total += child_res["total"]
 
         # "self" time is total time minus all time spent on children
         res["self"] = max(0.0, node.total - child_total)
-        if child_list:
-            res["children"] = child_list
+        if child_dict:
+            res["children"] = child_dict
 
         return res
 
--- a/ml-agents-envs/mlagents_envs/timers.py
+++ b/ml-agents-envs/mlagents_envs/timers.py
@@ -32,7 +32,7 @@ import math
 from time import perf_counter
 
 from contextlib import contextmanager
-from typing import Any, Callable, Dict, Generator, List, TypeVar
+from typing import Any, Callable, Dict, Generator, TypeVar
 
 
 class TimerNode:
@@ -183,13 +183,10 @@ class TimerStack:
             res["is_parallel"] = True
 
         child_total = 0.0
-        child_list = []
+        child_list = {}
         for child_name, child_node in node.children.items():
-            child_res: Dict[str, Any] = {
-                "name": child_name,
-                **self.get_timing_tree(child_node),
-            }
-            child_list.append(child_res)
+            child_res: Dict[str, Any] = self.get_timing_tree(child_node)
+            child_list[child_name] = child_res
             child_total += child_res["total"]
 
         # "self" time is total time minus all time spent on children
@@ -208,11 +205,10 @@ class TimerStack:
         else:
             self.gauges[name] = GaugeNode(value)
 
-    def _get_gauges(self) -> List[Dict[str, Any]]:
-        gauges = []
+    def _get_gauges(self) -> Dict[str, Dict[str, float]]:
+        gauges = {}
         for gauge_name, gauge_node in self.gauges.items():
-            gauge_dict: Dict[str, Any] = {"name": gauge_name, **gauge_node.as_dict()}
-            gauges.append(gauge_dict)
+            gauges[gauge_name] = gauge_node.as_dict()
         return gauges
 
 
--- a/ml-agents/mlagents/trainers/stats.py
+++ b/ml-agents/mlagents/trainers/stats.py
@@ -6,6 +6,7 @@ import csv
 import os
 
 from mlagents.tf_utils import tf
+from mlagents_envs.timers import set_gauge
 
 
 class StatsSummary(NamedTuple):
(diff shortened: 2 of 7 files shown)


ml-agents/mlagents/trainers/stats.py
```diff
@@ -35,6 +36,21 @@ def write_text(self, category: str, text: str, step: int) -> None:
...
+class GaugeWriter(StatsWriter):
+    """
+    Write all stats that we recieve to the timer gauges, so we can track them offline easily
+    """
+
+    def write_stats(
+        self, category: str, values: Dict[str, StatsSummary], step: int
+    ) -> None:
+        for val, stats_summary in values.items():
+            set_gauge(f"{category}.{val}.mean", float(stats_summary.mean))
```
chriselion: This currently produces gauges like
```
"ppo_3DBall.Policy/Entropy.mean": {
    "value": 1.4189382791519165,
    "min": 1.4189382791519165,
    "max": 1.4189382791519165,
    "count": 1
},
"ppo_3DBall.Policy/Learning Rate.mean": {
    "value": 0.00029642722802236676,
    "min": 0.00029642722802236676,
    "max": 0.00029642722802236676,
    "count": 1
},
"ppo_3DBall.Environment/Cumulative Reward.mean": {
    "value": 1.0899818162069077,
    "min": 1.0899818162069077,
    "max": 1.0899818162069077,
    "count": 1
},
"ppo_3DBall.Environment/Episode Length.mean": {
    "value": 20.899817850637522,
    "min": 20.899817850637522,
    "max": 20.899817850637522,
    "count": 1
},
"ppo_3DBall.Policy/Extrinsic Value Estimate.mean": {
    "value": -0.3887239396572113,
    "min": -0.3887239396572113,
    "max": -0.3887239396572113,
    "count": 1
},
"ppo_3DBall.Policy/Extrinsic Reward.mean": {
    "value": 1.088868589949434,
    "min": 1.088868589949434,
    "max": 1.088868589949434,
    "count": 1
}
```
which is good that they all show up, but a bit ugly in the naming. Any suggestions? I'm thinking about stripping spaces and replacing `/` with `.`

chriselion: @xiaomaogy I think this will record ELO like you asked about.

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -109,7 +109,7 @@ def _write_timing_tree(self) -> None:
         timing_path = f"{self.summaries_dir}/{self.run_id}_timers.json"
         try:
             with open(timing_path, "w") as f:
-                json.dump(get_timer_tree(), f, indent=2)
+                json.dump(get_timer_tree(), f, indent=4)
```
chriselion: Changed to match the output of `json.tool` (which I was using to format C# output)

ml-agents-envs/mlagents_envs/timers.py
```diff
@@ -183,13 +183,10 @@ def get_timing_tree(self, node: TimerNode = None) -> Dict[str, Any]:
             res["is_parallel"] = True
 
         child_total = 0.0
-        child_list = []
+        child_list = {}
```
harperj: `child_dict`?

ml-agents/mlagents/trainers/stats.py
harperj: On one hand I agree that these are ugly, but on the other this way keeps them identical to how they are written out by the trainer in case we needed to match them exactly in the future.  So, not sure what the best path would be here.

ervteng: This records the last min-max-mean right? So grabbing the "value" of the reward would be equivalent of grabbing the last reported reward. Might be useful @anupambhatnagar

ervteng: @xiaomaogy yep it will record ELO - actually it will record everything in Tensorboard. 

The category names (e.g. `Policy/Extrinsic Reward`) are currently structured like that so that they show up as proper categories in Tensorboard. I'm OK with stripping out / and space if needed. 

Would it be possible to make them children? e.g. rather than appending ppo_3DBall to every entry have a 
```
ppo_3DBall.stats: {
    Policy/Learning Rate.mean": {
    "value": 0.00029642722802236676,
    "min": 0.00029642722802236676,
    "max": 0.00029642722802236676,
    "count": 1
    },
etc.
}
```

ml-agents-envs/mlagents_envs/timers.py
chriselion: Doh, meant to change that.

ml-agents/mlagents/trainers/stats.py
chriselion: > This records the last min-max-mean right

The gauges record the min, max, and most recent value that has been passed to them. They're analogous to gauges in statsd and prometheus:
https://github.com/statsd/statsd/blob/master/docs/metric_types.md#gauges
https://prometheus.io/docs/concepts/metric_types/#gauge
So the value at the end of training is the average reward from the last summary interval. I only ran training for a few steps, so in this example, the min, max, and current value all look the same.

I didn't plan on stripping anything out from the tensorboard names, just the ones for the keys here.

I'd prefer to keep the gauges flat here, and go for a naming scheme like statsd (again), e.g. with `.` as the delimited. Some examples here https://matt.aimonetti.net/posts/2013-06-practical-guide-to-graphite-monitoring/#naming-metrics

chriselion: And just to be clear, the gauges aren't doing any averaging - that's just the mean from the StatsSummary.

xiaomaogy: @chriselion This looks perfect! exactly what I need.

PR #3473, opened 2020-02-19
Add command-line flag for example env scene selection

harperj:
### Proposed change(s)

The Startup scene used for us to retrain and bundle example scenes
together already accepts the SCENE_NAME environment variable to choose
a scene.  This change adds support for command-line selection of the
scene name so that `--env-args` can be used to choose the scene.

If both the environment variable and command-line flag are provided, the
scene will use the environment variable.

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist
- [ ] I have added tests that prove my fix is effective or that my feature works **Manual testing only at this time.**
- [x] I have added updated the changelog (N/A)
- [x] I have added necessary documentation (N/A)
- [x] I have updated the migration guide (N/A)


Project/Assets/ML-Agents/Examples/Startup/Scripts/Startup.cs
```diff
@@ -7,24 +7,48 @@ namespace MLAgentsExamples
     internal class Startup : MonoBehaviour
     {
         const string k_SceneVariableName = "SCENE_NAME";
+        private const string k_SceneCommandLineFlag = "--scene-name";
```
chriselion: I know this is in Examples, but can you prefix this with something like `--mlagents-scene-name`?

Project/Assets/ML-Agents/Examples/Startup/Scripts/Startup.cs
```diff
@@ -7,24 +7,48 @@ namespace MLAgentsExamples
...
 
         void Awake()
         {
-            var sceneName = Environment.GetEnvironmentVariable(k_SceneVariableName);
+            var sceneName = "";
+            
+            // Check for the CLI '--scene-name' flag.  This will be used if
+            // no scene environment variable is found.
+            var args = Environment.GetCommandLineArgs();
+            Console.WriteLine(String.Join(" ", args));
```
chriselion: Remove logging?

Project/Assets/ML-Agents/Examples/Startup/Scripts/Startup.cs
```diff
@@ -7,24 +7,48 @@ namespace MLAgentsExamples
...
         }
 
         static void SwitchScene(string sceneName)
         {
             if (sceneName == null)
             {
-                throw new ArgumentException(
-                    $"You didn't specified the {k_SceneVariableName} environment variable");
+                Console.WriteLine(
+                    $"You didn't specified the {k_SceneVariableName} environment variable."
```
chriselion:
```suggestion
                    $"You didn't specify the {k_SceneVariableName} environment variable or the {k_SceneCommandLineFlag} command line argument."
```

Project/Assets/ML-Agents/Examples/Startup/Scripts/Startup.cs
```diff
@@ -7,24 +7,48 @@ namespace MLAgentsExamples
...
+                );
+                Application.Quit();
```
chriselion: Non-zero return code?

Project/Assets/ML-Agents/Examples/Startup/Scripts/Startup.cs
```diff
@@ -7,24 +7,48 @@ namespace MLAgentsExamples
...
+                return;
             }
             if (SceneUtility.GetBuildIndexByScenePath(sceneName) < 0)
             {
-                throw new ArgumentException(
-                    $"The scene {sceneName} doesn't exist within your build. ");
+                Console.WriteLine(
+                    $"The scene {sceneName} doesn't exist within your build."
+                );
+                Application.Quit();
```
chriselion: Non-zero return code?

Project/ProjectSettings/EditorBuildSettings.asset
```diff
@@ -6,6 +6,12 @@ EditorBuildSettings:
   serializedVersion: 2
   m_Scenes:
   - enabled: 1
-    path: Packages/com.unity.ml-agents/Tests/Runtime/SerializeTestScene.unity
-    guid: 60783bd849bd242eeb66243542762b23
+    path: Assets/ML-Agents/Examples/Startup/Startup.unity
```
chriselion: Are these changes intentional?

Project/Assets/ML-Agents/Examples/Startup/Scripts/Startup.cs
harperj: Sure thing, will change.  I was on the fence about that one in any case but fell on the side of matching the env var.

Project/ProjectSettings/EditorBuildSettings.asset
harperj: Nope, reverting that change.

Project/Assets/ML-Agents/Examples/Startup/Scripts/Startup.cs
harperj: I was thinking this could be useful for making sure arguments were passed correctly in the logs, since the arguments passed come via some misdirection from `mlagents-learn`.  What do you think about leaving it?

chriselion: In that case, maybe preface it with "Command line args: " or something like that?

harperj: Done.

xiaomaogy:
This seems to deprecate the old way of using the environment variable to control the loading of the scene. I think it would be helpful to enable both for a while before directly deprecate the existing one.

harperj:
@xiaomaogy It explicitly was not supposed to deprecate using the environment variable, as I described in the PR description.  Did you do somehow test it and find that it didn't work?  because I'd like to fix it if so.

xiaomaogy:
> @xiaomaogy It explicitly was not supposed to deprecate using the environment variable, as I described in the PR description. Did you do somehow test it and find that it didn't work? because I'd like to fix it if so.

Oh I missed the code you added below to load the environment variable if no command line argument is passed.

harperj:
Sounds good, I'll merge this and let's just keep an eye out to make sure no issues come up in the retraining.

PR #3517, opened 2020-02-27
add DoneReason enum to Agent

chriselion:
### Proposed change(s)

In order to accurately track inference rewards, we need to make sure that we don't record incomplete results when an Agent is disabled (namely at shutdown). Previously, we were stopping updates of the timers. This changes Agent.NotifyAgentDone to take an enum for the shutdown reason, so that we can exclude timers when coming from OnDisable.

This allows us to mark TimerStack internal.

No new public API.

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/com.unity.ml-agents/Runtime/Timer.cs
+++ b/com.unity.ml-agents/Runtime/Timer.cs
@@ -285,7 +285,7 @@ namespace MLAgents
     /// This implements the Singleton pattern (solution 4) as described in
     /// https://csharpindepth.com/articles/singleton
     /// </remarks>
-    public class TimerStack : IDisposable
+    internal class TimerStack : IDisposable
     {
         static readonly TimerStack k_Instance = new TimerStack();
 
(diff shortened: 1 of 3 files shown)


com.unity.ml-agents/Runtime/Timer.cs
```diff
@@ -285,7 +285,7 @@ public void Update(float newValue)
     /// This implements the Singleton pattern (solution 4) as described in
     /// https://csharpindepth.com/articles/singleton
     /// </remarks>
-    public class TimerStack : IDisposable
+    internal class TimerStack : IDisposable
```
surfnerd: is this supposed to be here?

chriselion: Yes. That was the main goal.

surfnerd: ok

PR #3537, opened 2020-03-02
Simple1DEnv refactor and additional ghost trainer tests

andrewcoh:
### Proposed change(s)

The Simple1DEnv used in testing has been modified

- In it's own file so that we can more cleanly create variants for testing other algorithms.
- Accepts a list of brain names for agents so that we can test multi agent.
- Attributes have been updated to dicts so that we can track position and other statistics of multiple agents.
- The state is no longer the agent's goal but its absolute position in 1d space.

Tests for the ghost trainer to make sure

- Agent's can learn to solve the environment when the trainer is wrapped by the ghost trainer
- The ghost trainer swaps the correct policies by checking the performance of the ghosted agent when it receives the current trained policy (it should solve the env) and when it receives the initial untrained policy (it should fail).

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Other (please describe)
Adding more tests

### Checklist


ml-agents/mlagents/trainers/tests/test_simple_rl.py
```diff
@@ -262,28 +182,45 @@ def _check_environment_trains(
...
         tc.start_learning(env_manager)
-        print(tc._get_measure_vals())
         if (
             success_threshold is not None
         ):  # For tests where we are just checking setup and not reward
-            for mean_reward in tc._get_measure_vals().values():
-                assert not math.isnan(mean_reward)
-                assert mean_reward > success_threshold
+
+            processed_rewards = [
```
chriselion: Nice, this seems a lot cleaner than the old `_get_measure_vals()` approach.

We might be able to remove or refactor `_get_measure_vals()` now too.

ml-agents/mlagents/trainers/tests/test_simple_rl.py
```diff
@@ -262,28 +182,45 @@ def _check_environment_trains(
...
+                reward_processor(rewards) for rewards in env.final_rewards.values()
+            ]
+            assert all(not math.isnan(reward) for reward in processed_rewards)
+            if not check_fail:
+                assert all(reward > success_threshold for reward in processed_rewards)
+            else:
+                assert any(
+                    reward > success_threshold for reward in processed_rewards
```
chriselion: If this is only going to be used for the ghost training test, maybe it would be better to leave out of the general function and just grab the rewards from `env` in `test_simple_ghost_fails()`

ml-agents/mlagents/trainers/tests/test_simple_rl.py
```diff
@@ -262,28 +182,45 @@ def _check_environment_trains(
...
-    env = Simple1DEnvironment(use_discrete=use_discrete)
+    env = Simple1DEnvironment([BRAIN_NAME], use_discrete=use_discrete)
     _check_environment_trains(env, SAC_CONFIG)
 
 
 @pytest.mark.parametrize("use_discrete", [True, False])
 def test_simple_ghost(use_discrete):
-    env = Simple1DEnvironment(use_discrete=use_discrete)
-    _check_environment_trains(env, GHOST_CONFIG)
+    env = Simple1DEnvironment(
```
chriselion: Maybe move these (and GHOST_CONFIG_*) to their own file?

ml-agents/mlagents/trainers/tests/test_simple_rl.py
ervteng: I think the `_get_measure_vals()` is still used by the MetaCurriculum to increment the lessons. It's a bit weird since it's the only reason the Trainer keeps track of rewards at all; for most reporting it happens in the EnvManager's AgentProcessors. 
Instead we can have the `_get_measure_vals()` or equivalent grab the rewards from the EnvManager, this should be pretty easy

ml-agents/mlagents/trainers/tests/test_simple_rl.py
ervteng: @andrewcoh should we just move these tests to their respective places? e.g. `test_ghost.py`, `test_ppo.py`, etc.

ml-agents/mlagents/trainers/tests/test_simple_rl.py
andrewcoh: True. This is actually only special casing for the ghost failure test which I agree should be done elsewhere.  I can remove the check_fail flag too.

ml-agents/mlagents/trainers/tests/test_simple_rl.py
andrewcoh: I could move this to test_ghost.py. Then, it also might make sense to move the ppo/sac stuff to test_ppo.py and test_sac.py and leave this file as just _check_env_trains.  Thoughts @ervteng ?

andrewcoh: I should've refreshed :)

PR #3558, opened 2020-03-04
[bug-fix] Change Simple1DEnvironment to spawn new agent IDs on reset

ervteng:
### Proposed change(s)

In recent versions of ML-Agents, an agent reset happens immediately and changes that agent's ID. Then, the step that is seen after the reset is seen at the same time with the new agent ID.

This PR changes the Simple1D environment (and the memory version of it) to do that. We also change the batched step result to not send max step reached (tells the trainer to treat the episode as ongoing). 

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/mlagents/trainers/tests/simple_test_envs.py
+++ b/ml-agents/mlagents/trainers/tests/simple_test_envs.py
@@ -57,7 +57,7 @@ class Simple1DEnvironment(BaseEnv):
         self.rewards: Dict[str, float] = {}
         self.final_rewards: Dict[str, List[float]] = {}
         self.step_result: Dict[str, BatchedStepResult] = {}
-        self.agent_id: Dict[str.int] = {}
+        self.agent_id: Dict[str, int] = {}
         self.step_size = step_size  # defines the difficulty of the test
 
         for name in self.names:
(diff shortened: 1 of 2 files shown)


ml-agents/mlagents/trainers/tests/simple_test_envs.py
```diff
@@ -57,9 +57,11 @@ def __init__(
         self.rewards: Dict[str, float] = {}
         self.final_rewards: Dict[str, List[float]] = {}
         self.step_result: Dict[str, BatchedStepResult] = {}
+        self.agent_id: Dict[str.int] = {}
```
chriselion:
```suggestion
        self.agent_id: Dict[str, int] = {}
```
Very surprised that mypy didn't complain about that (but it might skip tests)

ml-agents/mlagents/trainers/tests/simple_test_envs.py
```diff
@@ -158,6 +211,7 @@ def _reset_agent(self, name):
         self.step_count[name] = 0
         self.final_rewards[name].append(self.rewards[name])
         self.rewards[name] = 0
+        self.agent_id[name] = (self.agent_id[name] + 1) % 2
```
chriselion: Why not constantly increasing? That's closer to the C# behavior.

ervteng: I can do that. Side note: does C# ever loop around?

PR #3560, opened 2020-03-04
Add timestamps to logs

anupambhatnagar:
### Proposed change(s)

1. Modified the current logger to include timestamps.
1. `basicConfig` should be used exactly once. It was used in `environment.py` and `learn.py`. code has been refactored. it is called only in `logging_util.py` now.

Some code cleanup
1. Removed "import logging" statements and loggers from files which do not use them.
1. Minor formatting changes:
   - removed blank lines
   - some uppercase to lowercase conversion to maintain consistency

https://jira.unity3d.com/browse/MLA-614

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### Other comments
Here's a sample of the new log format

```2020-03-03 23:56:51.882456: I tensorflow/core/platform/cpu_feature_guard.cc:142] Your CPU supports instructions that this TensorFlow binary was not compiled to use: AVX2 FMA
2020-03-03 23:56:51.897480: I tensorflow/compiler/xla/service/service.cc:168] XLA service 0x7faa22c166b0 executing computations on platform Host. Devices:
2020-03-03 23:56:51.897496: I tensorflow/compiler/xla/service/service.cc:175]   StreamExecutor device (0): Host, Default Version
2020-03-03 23:57:04 INFO [trainer.py:213] 3dball-no-logging3: 3DBall: Step: 12000. Time Elapsed: 12.269 s Mean Reward: 1.198. Std of Reward: 0.741. Training.
2020-03-03 23:57:17 INFO [trainer.py:213] 3dball-no-logging3: 3DBall: Step: 24000. Time Elapsed: 25.191 s Mean Reward: 1.405. Std of Reward: 0.935. Training.
2020-03-03 23:57:29 INFO [trainer.py:213] 3dball-no-logging3: 3DBall: Step: 36000. Time Elapsed: 37.932 s Mean Reward: 1.851. Std of Reward: 1.286. Training.

### DIFF
--- a/ml-agents/mlagents/trainers/ghost/trainer.py
+++ b/ml-agents/mlagents/trainers/ghost/trainer.py
@@ -1,7 +1,6 @@
 # # Unity ML-Agents Toolkit
 # ## ML-Agent Learning (Ghost Trainer)
 
-# import logging
 from typing import Deque, Dict, List, Any, cast
 
 import numpy as np
@@ -15,7 +14,7 @@ from mlagents.trainers.trainer import Trainer
 from mlagents.trainers.trajectory import Trajectory
 from mlagents.trainers.agent_processor import AgentManagerQueue
 
-LOGGER = logging.getLogger("mlagents.trainers")
+logger = logging.getLogger("mlagents.trainers")
 
 
 class GhostTrainer(Trainer):
@@ -93,7 +92,7 @@ class GhostTrainer(Trainer):
         Saves training statistics to Tensorboard.
         """
         opponents = np.array(self.policy_elos, dtype=np.float32)
-        LOGGER.info(
+        logger.info(
             " Learning brain {} ELO: {:0.3f}\n"
             "Mean Opponent ELO: {:0.3f}"
             " Std Opponent ELO: {:0.3f}".format(
@@ -234,7 +233,7 @@ class GhostTrainer(Trainer):
                 x = "current"
                 self.policy_elos[-1] = self.current_elo
             self.current_opponent = -1 if x == "current" else x
-            LOGGER.debug(
+            logger.debug(
                 "Step {}: Swapping snapshot {} to id {} with {} learning".format(
                     self.get_step, x, name_behavior_id, self.learning_behavior_name
                 )
--- a/ml-agents/mlagents/trainers/learn.py
+++ b/ml-agents/mlagents/trainers/learn.py
@@ -451,16 +451,14 @@ def run_cli(options: RunOptions) -> None:
     except Exception:
         print("\n\n\tUnity Technologies\n")
     print(get_version_string())
-    trainer_logger = create_logger("mlagents.trainers")
-    env_logger = create_logger("mlagents_envs")
+    trainer_logger = create_logger("mlagents.trainers", log_level=logging.INFO)
+    env_logger = create_logger("mlagents_envs", log_level=logging.INFO)
 
     if options.debug:
         trainer_logger.setLevel("DEBUG")
         env_logger.setLevel("DEBUG")
     else:
-        trainer_logger.setLevel("INFO")
-        env_logger.setLevel("INFO")
-        # disable noisy warnings from tensorflow.
+        # disable noisy warnings from tensorflow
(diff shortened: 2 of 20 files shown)

### CODE REVIEW

File: ml-agents/mlagents/trainers/ghost/trainer.py [hunk 1]
```diff
@@ -15,7 +14,7 @@
 from mlagents.trainers.trajectory import Trajectory
 from mlagents.trainers.agent_processor import AgentManagerQueue
 
-LOGGER = logging.getLogger("mlagents.trainers")
+logger = logging.getLogger("mlagents.trainers")
```
chriselion: Why not create_logger here?

ml-agents/mlagents/trainers/learn.py
```diff
@@ -450,8 +451,8 @@ def run_cli(options: RunOptions) -> None:
     except Exception:
         print("\n\n\tUnity Technologies\n")
     print(get_version_string())
-    trainer_logger = logging.getLogger("mlagents.trainers")
-    env_logger = logging.getLogger("mlagents_envs")
+    trainer_logger = create_logger("mlagents.trainers")
```
chriselion: How about letting `create_logger()` take the log level as an optional argument (and default to INFO)? That would simplify some of the code below.

chriselion: Actually, I think what you want to do is
* call `logging.basicConfig(level=logging.DEBUG)` or `logging.INFO` depending on `options.debug`
* don't set the logging level in `create_logger()`

I think if you do that, the new loggers will inherit the top-level log level.

ml-agents/mlagents/trainers/ghost/trainer.py
anupambhatnagar: its not clear to me where the logger is initialized. there are several import statements in each file and since learn.py is the entry point of mlagents-learn I think it makes sense to create the logger there.

ml-agents/mlagents/trainers/learn.py
anupambhatnagar: that makes sense. I'll add the log_level as an argument to the create_logger signature.

ml-agents/mlagents/trainers/learn.py
```diff
@@ -450,8 +451,8 @@ def run_cli(options: RunOptions) -> None:
...
+    env_logger = create_logger("mlagents_envs")
 
     if options.debug:
         trainer_logger.setLevel("DEBUG")
```
ervteng: We might be able to delete the `else` in this block since we're setting the level in `create_logger`

anupambhatnagar: @ervteng are you suggesting that we make `tf_utils.set_warnings_enabled(False)` the default?

ml-agents/mlagents/trainers/learn.py
anupambhatnagar: the reason it has been hard to debug the logger initialization is that basicConfig was called twice in different files. given the order of import statements the config can be overridden. Hence I prefer to centralize the call to basicConfig in exactly one location and make sure its called exactly once per logger. I shall leave the basicConfig in logging_util.py

ml-agents/mlagents/trainers/ghost/trainer.py
ervteng: I think the only way to be absolutely sure that the basicConfig is applied is to do `import logging` and `logger = create_logger("mlagents.trainers")` at the very top of `learn.py`, since accidentally importing any file with the `getLogger` will cause the config to be ignored. 

@chriselion could we do something similar to the `tf_util.py` where we create and import the logger from logging_utils rather than the default module? It would be weird since we'd have to create the envs logger as well here.

chriselion: I think ideally we should get all the loggers going through create_logger, and they all should use the logging level decided by `options.debug`. This is probably no worse than before though, so OK to merge this today.

PR #3583, opened 2020-03-06
[change] Throw a proper error when sequence length is greater than batch size. 

ervteng:
### Proposed change(s)

When using recurrent networks, it was previously possible to make sequence length greater than batch size. In PPO, the code sneakily set the sequence length equal to the batch size. In SAC, it crashed.

Now both throw a logical error message. 

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


ml-agents/mlagents/trainers/sac/trainer.py
```diff
@@ -99,6 +100,19 @@ def __init__(
             else False
         )
 
+    def _check_param_keys(self):
```
ervteng: This code is duplicated (and not put in a higher-level class) in case we decided to add algorithm-specific behavior, or remove one of these parameters from one but not both of the algorithms.

ml-agents/mlagents/trainers/ppo/trainer.py
```diff
@@ -69,6 +71,19 @@ def __init__(
...
+        super()._check_param_keys()
+        # Check that batch size is greater than sequence length. Else, throw
+        # an exception.
+        if (
+            self.trainer_parameters["sequence_length"]
+            > self.trainer_parameters["batch_size"]
+            and self.trainer_parameters["use_recurrent"]
+        ):
+            raise UnityTrainerException(
+                "batch_size must be greater than or equal to sequence_length."
```
chriselion:
```suggestion
                "batch_size must be greater than or equal to sequence_length when use_recurrent is True."
```
(and same below)

chriselion:
> In PPO, the code sneakily set the sequence length equal to the batch size

How about being less sneaky (log a warning) but still doing this?

ervteng:
> > In PPO, the code sneakily set the sequence length equal to the batch size
> 
> How about being less sneaky (log a warning) but still doing this?

Open to this. In general my feeling is that warnings tend to get lost (and thus ignored) in the stream of prints that occur when the trainer starts, and the user will never know what their actual batch size is - perhaps this isn't a real issue and most users won't care if their batch size is bigger than it was set

chriselion:
Yeah, I don't think they care that much, but the downside is they might waste time "optimizing" the sequence length when it's not because it's getting overridden. I'm fine with the exception approach.

ml-agents/mlagents/trainers/sac/trainer.py
andrewcoh: That makes sense. In this case, do you have an algorithm in mind where you wouldn't want to check these hyperparameters? Else, it would make sense to me to put this in a higher level class.

ervteng: Anything that doesn't support LSTM :D

andrewcoh: Alright.

PR #3586, opened 2020-03-07
fix RenderTexture unit tests in DevProject

chriselion:
### Proposed change(s)

* Add a placeholder file in DevProject/Assets. Otherwise a clean git checkout won't be a valid project according to unity hub 😠 
* Fix typos in test class names or file names so that they actually run in the 2019.3 test runner.

image


surfnerd:
thanks, i meant to take care of this a while ago 😬

PR #3594, opened 2020-03-08
Custom tensorboard observations

MikeWise2718:
### Proposed change(s)

Added custom TensorBoard observations - This is code I had running last December, but I worked this weekend to bring it up to date with the current master branch.

image

### Overview
Here is a summary of the important changes I had to do to get this to work

#### Protobuf changes:
- Added a new type to the protobuf defintions (**EnvironmentStatisticsProto**) to communicate data to the Python UnityEnvironment and merge it into **BatchedStepResult**. 
- Added **EnvironmentStatisticsProto** to **UnityRLOutputProto**

#### Python changes:
- Modified **base_env.py**, **rpc_utils.py**, and **environment.py** to read those.
- Modified **agent_processor.py** to pass those values to TensorBoard

#### C# Changes:
- Added a class to manage the collections **EnvStatMan.cs** in **com.unity.agents/Runtime**
- Made some minor modifications to **Academy.cs** and **RpcCommunication.cs** to integrate the **EnvStatMan** class into the RPC communicatino.
- Added a couple of counters (**foodEaten**, **poisonEaten**) to the **FoodCollectionSettings.cs**
- Added a FixedUpdate event handler to **FoodCollectionSettings.cs** to update those settings.

#### Other Changes
- There are a lot of little unimportant changes in this pull request that should not be there, but I am not sure how to quickly get rid of these so I am leaving them in for now. I think it is obvious what is signal and what is noise.

Let me know if you have questions or requests.

Mike Wise

mwise@microsoft.com

### Types of change(s)

- [ ] New feature


chriselion:
Hi @MikeWise2718,
Thanks for the submission. The team needs to talk internally about this a bit more - it's definitely a useful feature (there have been several requests for it), but I don't think think this is the way we want it implemented. 

I'm trying to wrap up a few other loose ends, but I'll return to this later this week...

MikeWise2718:
Thanks for taking a look.

Is it about using the `bytes side_channel` instead of a custom `map` in the protobuf definitions? I could give it a shot of using that. I thought about using that, but I didn't see how I could  `side_channel` in a way that would accommodate usage by other features - there needs to be some kind of id I would think.

I could also clean up the submission if that would help. I didn't do a great job on getting rid of extraneous changes before submitting, which is kind of stupid I guess.

chriselion:
Hi @MikeWise2718,
Sorry for the delay, as you can imagine things are a bit crazy right now, and it took a while to get the group together to discuss how to handle this.

I'm going to start working on a sidechannel approach for this; hopefully I'll have something to review today or tomorrow.

chriselion:
Hi @MikeWise2718,
This was implemented in https://github.com/Unity-Technologies/ml-agents/pull/3660 and I think it should be equivalent to your PR in terms of interface and end results.

chriselion:
Closing this off since the equivalent functionality was added.

PR #3600, opened 2020-03-10
enforce onnx conversion (expect tf2 CI to fail)

chriselion:
### Proposed change(s)

We currently log exceptions for onnx conversion, but don't raise the exceptions, so we might be getting silent errors during tests. This PR will optionally re-raise the exception if an environment variable is set, and set this in CI for one of our configurations (the latest 1.x version of tensorflow).

One thing that came out of this is that tf2onnx doesn't play nicely with very old versions of tensorflow. They say that they only test with 1.12.0 and later, so don't try to convert to onnx if that's the case.

https://jira.unity3d.com/browse/MLA-740

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/mlagents/model_serialization.py
+++ b/ml-agents/mlagents/model_serialization.py
@@ -2,6 +2,7 @@ from distutils.util import strtobool
 import os
 import logging
 from typing import Any, List, Set, NamedTuple
+from distutils.version import LooseVersion
 
 try:
     import onnx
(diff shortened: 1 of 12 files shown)


ml-agents/mlagents/model_serialization.py
```diff
@@ -18,6 +21,11 @@
 from tensorflow.python.framework import graph_util
 from mlagents.trainers import tensorflow_to_barracuda as tf2bc
 
+if LooseVersion(tf.__version__) < LooseVersion("1.12.0"):
+    # ONNX is only tested on 1.12.0 and later
```
chriselion: Accoridng to https://github.com/onnx/tensorflow-onnx#tensorflow

PR #3626, opened 2020-03-12
Basic GAIL and BC integration tests

andrewcoh:
### Proposed change(s)

This PR adds GAIL and BC integration tests for both PPO and SAC or discrete and continuous actions

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Other (please describe)

### Checklist


ml-agents-envs/mlagents_envs/rpc_utils.py
```diff
@@ -199,6 +203,68 @@ def batched_step_result_from_proto(
...
+) -> List[AgentInfoProto]:
+    agent_info_protos: List[AgentInfoProto] = []
+    for agent_id in batched_step_result.agent_id:
+        agent_id_index = batched_step_result.get_index(agent_id)
+        reward = batched_step_result.reward[agent_id_index]
+        done = batched_step_result.done[agent_id_index]
+        max_step_reached = batched_step_result.max_step[agent_id_index]
+        agent_mask = None
+        if batched_step_result.action_mask is not None:
+            mask = batched_step_result.action_mask[0]
```
andrewcoh: I could not get this to work otherwise.  The protobuf was not expecting a list of action masks as in the StepResult.  Thoughts @ervteng ?

ml-agents/mlagents/trainers/tests/test_simple_rl.py
```diff
@@ -323,3 +330,43 @@ def test_simple_ghost_fails(use_discrete):
     assert any(reward > success_threshold for reward in processed_rewards) and any(
         reward < success_threshold for reward in processed_rewards
     )
+
+
+@pytest.mark.parametrize("use_discrete", [True, False])
+def test_simple_record(use_discrete):
```
andrewcoh: We don't actually need to run this test every time since the .demos are written to a file but it does test the new write demo functions etc.

ml-agents-envs/mlagents_envs/base_env.py
```diff
@@ -109,6 +109,15 @@ def agent_id_to_index(self) -> Dict[AgentId, int]:
...
         return agent_id in self.agent_id_to_index
 
+    def get_index(self, agent_id: AgentId) -> int:
+        if not self.contains_agent(agent_id):
+            raise IndexError(
+                "get_index failed. agent_id {} is not present in the BatchedStepResult".format(
+                    agent_id
+                )
+            )
+        return self._agent_id_to_index[agent_id]  # type: ignore
```
chriselion: How about not calling `contains_agent` and just do `self.agent_id_to_index.get(agent_id, -1)`?

andrewcoh: No problem. I was following suit with get_agent_step_result below.

ervteng:
We shouldn't have the `1DTestContinuous.demo` and `1DTestDiscrete.demo` files right?

ml-agents-envs/mlagents_envs/base_env.py
```diff
@@ -109,6 +109,9 @@ def agent_id_to_index(self) -> Dict[AgentId, int]:
     def contains_agent(self, agent_id: AgentId) -> bool:
         return agent_id in self.agent_id_to_index
 
+    def get_index(self, agent_id: AgentId) -> int:
```
ervteng: Why do we need this function rather than just using `agent_id_to_index[agent_id]?` Since you're iterating over the list of agents it's guaranteed to exist

andrewcoh:
> We shouldn't have the `1DTestContinuous.demo` and `1DTestDiscrete.demo` files right?

I was including them so that one test did not depend on another.

ml-agents/mlagents/trainers/tests/test_simple_rl.py
```diff
@@ -323,3 +330,43 @@ def test_simple_ghost_fails(use_discrete):
...
+    brain_param_proto = BrainParametersProto(
+        vector_action_size=[1],
+        vector_action_descriptions=[""],
+        vector_action_space_type=discrete if use_discrete else continuous,
+        brain_name=BRAIN_NAME,
+        is_training=True,
+    )
+    action_type = "Discrete" if use_discrete else "Continuous"
+    demo_path = "demos/1DTest" + action_type + ".demo"
+    write_demo(demo_path, meta_data_proto, brain_param_proto, agent_info_protos)
```
ervteng: We should save this to a temporary directory once and use multiple times by the other tests. Pytest provides a way to do this: https://docs.pytest.org/en/latest/tmpdir.html#the-tmpdir-factory-fixture

ml-agents/mlagents/trainers/tests/test_simple_rl.py
```diff
@@ -323,3 +330,47 @@ def test_simple_ghost_fails(use_discrete):
...
+        write_demo(demo_path, meta_data_proto, brain_param_proto, agent_info_protos)
+        return demo_path
+
+    return record_demo
+
+
+@pytest.mark.parametrize("use_discrete", [True, False])
+@pytest.mark.parametrize("trainer_config", [PPO_CONFIG, SAC_CONFIG])
+def test_gail(simple_record, use_discrete, trainer_config):
+    demo_path = simple_record(use_discrete)
```
ervteng: Can we add an assert to check if the demo's rewards exceed the threshold? So if the demo recording training fails, we know why this fails

ml-agents-envs/mlagents_envs/rpc_utils.py
```diff
@@ -199,6 +203,71 @@ def batched_step_result_from_proto(
...
+                agent_mask = np.concatenate(
+                    (agent_mask, _branch[agent_id_index, :]), axis=0
+                )
+        observations: List[ObservationProto] = []
+        for all_observations_of_type in batched_step_result.obs:
+            observation = all_observations_of_type[agent_id_index]
+            if len(observation.shape) == 3:
+                observations.append(
+                    ObservationProto(
+                        compressed_data=observation,
```
ervteng: this probably should be float_data - not sure though (@chriselion)

chriselion: It's not safe to assume that (len(shape)==3) => compressed. I need to double-check the BatchedStep result, but I think a PNG observation will have been decompressed to floats already. If that's the case, you should just always do the float_data approach

andrewcoh: Ok. Then in our 1d test env, should I flatten the faked visual?

chriselion: I think keep it multi-D there, and flatten (I assume that's the reverse of `np.reshape`?) here.

ml-agents/mlagents/trainers/tests/simple_test_envs.py
```diff
@@ -228,7 +232,7 @@ def close(self):
 
 class Memory1DEnvironment(Simple1DEnvironment):
     def __init__(self, brain_names, use_discrete, step_size=0.2):
-        super().__init__(brain_names, use_discrete, step_size=0.2)
+        super().__init__(brain_names, use_discrete, step_size=step_size)
```
ervteng: Good call 👍

ml-agents/mlagents/trainers/demo_loader.py
```diff
@@ -164,3 +167,21 @@ def load_demonstration(
             f"No BrainParameters found in demonstration file at {file_path}."
         )
     return group_spec, info_action_pairs, total_expected
+
+
+def write_delimited(f, message):
```
ervteng: Should these two be in RPC tests? I'm OK with leaving these here as well

PR #3634, opened 2020-03-13
[change] Clean up trainer interface, clean up GhostTrainer stats

ervteng:
### Proposed change(s)

This PR moves most private methods out of the Trainer class and into RLTrainer, making it more of an interface. This also removed some methods in the Ghost Trainer. As part of this change, it was necessary to move the console printing out of the Ghost Trainer and into the ConsoleWriter. @andrewcoh I need to make sure that I'm writing the stats in the right place, and that the ELO reported where it is won't cause issues.

Also moved reward reporting from the AgentProcessor to the Trainer, so that the GhostTrainer's cumulative reward is (correctly) only one of the learning teams. This doesn't affect single-team training.

There is some special casing and property passing in the ConsoleWriter that is a bit hacky - open to ideas on how to solve that. 

Also fixes a bug where hyperparameters were printed twice for ghost trainer. 

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/mlagents/trainers/ghost/trainer.py
+++ b/ml-agents/mlagents/trainers/ghost/trainer.py
@@ -13,6 +13,8 @@ from mlagents.trainers.policy.tf_policy import TFPolicy
 from mlagents.trainers.trainer import Trainer
 from mlagents.trainers.trajectory import Trajectory
 from mlagents.trainers.agent_processor import AgentManagerQueue
+from mlagents.trainers.stats import StatsPropertyType
+from mlagents.trainers.behavior_id_utils import BehaviorIdentifiers
 
 logger = logging.getLogger("mlagents.trainers")
 
@@ -43,7 +45,9 @@ class GhostTrainer(Trainer):
         self.learning_policy_queues: Dict[str, AgentManagerQueue[Policy]] = {}
 
         # assign ghost's stats collection to wrapped trainer's
-        self.stats_reporter = self.trainer.stats_reporter
+        self._stats_reporter = self.trainer.stats_reporter
+        # Set the logging to print ELO in the console
+        self._stats_reporter.add_property(StatsPropertyType.SELF_PLAY, True)
 
         self_play_parameters = trainer_parameters["self_play"]
         self.window = self_play_parameters.get("window", 10)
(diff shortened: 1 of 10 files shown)


ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -203,6 +190,11 @@ def add_policy(self, name_behavior_id: str, policy: TFPolicy) -> None:
             self.trainer.add_policy(name_behavior_id, policy)
             self._save_snapshot(policy)  # Need to save after trainer initializes policy
             self.learning_behavior_name = name_behavior_id
+            behavior_id_parsed = BehaviorIdentifiers.from_name_behavior_id(
```
andrewcoh: Should we send the behavior ids as a named tuple to all trainers/ghost trainer instead of parsing this in every advance step?

ml-agents/mlagents/trainers/stats.py
```diff
@@ -77,6 +79,9 @@ def write_stats(
 class ConsoleWriter(StatsWriter):
     def __init__(self):
         self.training_start_time = time.time()
+        # If self-play, we want to print ELO instead of reward
```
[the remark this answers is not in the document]
ervteng: It's in addition to - I didn't change how the ghosted agent adds the reward (it's still done by the AgentProcessor/Manager). 

Currently the trainer's "Extrinsic Reward" log (under Policy in TB) is what we want. We can do a hack in the ConsoleWriter to output Extrinsic Reward instead of Cumulative Reward (meaningless) as the reward if using self-play

ml-agents/mlagents/trainers/ghost/trainer.py
ervteng: This is being parsed once right? Just at add policy. 

But you're right - maybe we should parse the behavior ID once in the trainer controller rather than inside the trainer, and introduce a new class (BehaviorIdentifier, with properties behavior_id and team_id).

ml-agents/mlagents/trainers/stats.py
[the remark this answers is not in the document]
andrewcoh: How would you feel about that? It could resolve this https://jira.unity3d.com/browse/MLA-760 in part.

ml-agents/mlagents/trainers/ghost/trainer.py
andrewcoh: The object the parsing returns is a named tuple.  It would probably be good to pass that around instead.  I can make this in a separate PR since I was considering it for multiagent training too.

PR #3653, opened 2020-03-19
Asymmetric self-play

andrewcoh:
### Proposed change(s)
Extends the toolkit to support ghost training of asymmetric games. Under our abstractions, this case would require multiple ghost trainer with higher level coordination. As a solution I've implemented something that's essentially a mutex.  All ghost trainers query the GhostController for the team that is learning and this value determines how the ghost trainer treats trajectories/policies. This also makes much more use of the behavior identifiers which is a step toward supporting multi-agent.
 

TODO:
* ~~add tests for the asymmetric case~~
* ~~add a real example environment to showcase this. i'll do that on a separate branch though to keep this small.~~ https://github.com/Unity-Technologies/ml-agents/pull/3699
* ~~update docs~~

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -38,11 +46,12 @@ def __init__(
         )
 
         self.trainer = trainer
+        self.controller = controller
```
vincentpierre: There is a cyclic reference between the controller and the ghost trainer. Not sure if we care about those in Python...

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -56,18 +65,24 @@ def __init__(
         )
         self.steps_between_save = self_play_parameters.get("save_steps", 20000)
         self.steps_between_swap = self_play_parameters.get("swap_steps", 20000)
+        self.ghost_step: int = 0
```
vincentpierre: More comments. Or else II will rename this variable to `spooky_stair_unit`.

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -21,12 +21,20 @@
...
+        trainer,
+        brain_name,
+        controller,
+        reward_buff_cap,
+        trainer_parameters,
+        training,
+        run_id,
     ):
         """
         Responsible for collecting experiences and training trainer model via self_play.
```
vincentpierre: Longer comment here. Especially around how many GhostTrainers per BehaviorName/TeamId.

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -91,6 +106,19 @@ def reward_buffer(self) -> Deque[float]:
          """
         return self.trainer.reward_buffer
 
+    @property
+    def current_elo(self) -> float:
```
vincentpierre: Method comments **everywhere**.

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -101,65 +129,77 @@ def _process_trajectory(self, trajectory: Trajectory) -> None:
             elif final_reward < 0:
                 result = 0.0
 
-            change = compute_elo_rating_changes(
-                self.current_elo, self.policy_elos[self.current_opponent], result
+            change = self.controller.compute_elo_rating_changes(
```
vincentpierre: Would it make more sense to have compute_elo_rating_changes on the Ghost trainer or is there some data only the controller has access to?

ml-agents/mlagents/trainers/learn.py
```diff
@@ -101,6 +101,14 @@ def _create_parser():
         type=int,
         help="Number of parallel environments to use for training",
     )
+
+    argparser.add_argument(
+        "--ghost-swap",
```
vincentpierre: This needs a less confusing name and a less confusing description. (If such a thing exists)

ml-agents/mlagents/trainers/trainer_util.py
```diff
@@ -38,6 +40,7 @@ def __init__(
         self.seed = seed
         self.meta_curriculum = meta_curriculum
         self.multi_gpu = multi_gpu
+        self.ghost_controller = GhostController(ghost_swap)
```
vincentpierre: It seems GhostController 1:1 with a trainer. Can you explain how this makes it work for asymmetric use cases (multiple trainers)
Do we always need a GhostController even when there is a single agent ? It seems we are creating an object we might not always need.

ml-agents/mlagents/trainers/behavior_id_utils.py
```diff
@@ -17,20 +17,14 @@ def from_name_behavior_id(name_behavior_id: str) -> "BehaviorIdentifiers":
...
-            if "&" in identifiers:
-                list_of_identifiers = identifiers.split("&")
-            else:
-                list_of_identifiers = [identifiers]
-
-            for identifier in list_of_identifiers:
-                key, value = identifier.split("=")
-                ids[key] = int(value)
+            name, team_and_id = name_behavior_id.rsplit("?", 1)
+            _, team_id_str = team_and_id.split("=")
```
chriselion: This is going to need to change again when we start passing the role (since split will return more that 2 values). I think we should just start using https://docs.python.org/3.7/library/urllib.parse.html for this, to avoid edge cases that we haven't considered. 

```
>>> from urllib.parse import urlparse, parse_qs
>>> parsed = urlparse("soccer?teamId=1&role=striker")
>>> parsed.path
'soccer'
>>> parse_qs(parsed.query)
{'teamId': ['1'], 'role': ['striker']}
```
Then we can extract team and role from the parse_qs result as needed.

docs/Training-Self-Play.md
```diff
@@ -15,7 +15,7 @@ Self-play is triggered by including the self-play hyperparameter hierarchy in th
 
 ![Team ID](images/team_id.png)
 
-See the trainer configuration and agent prefabs for our Tennis environment for an example.
+***Team ID must be 0 or an integer greater than 0. Negative numbers will cause unpredictable behavior.*** See the trainer configuration and agent prefabs for our Tennis environment for an example.
```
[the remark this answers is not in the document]
chriselion: Why does negative produce unpredictable results? If it's just from the parsing, we might need to fix the parsing instead.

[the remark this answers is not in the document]
andrewcoh: @vincentpierre This is set in the editor. What do you mean?

@chriselion Mypy wouldn't let me 'None' [this](https://github.com/Unity-Technologies/ml-agents/blob/43417e14340d50f2254b866713e19640949dc1d6/ml-agents/mlagents/trainers/ghost/controller.py#L18) so I set it to an int less than 0 to indicate that it had not been initialized.

ml-agents/mlagents/trainers/behavior_id_utils.py
andrewcoh: Yeah, I knew this would have to change when role was added.  Unfortunately, I'm not sure that's going to happen because Ervin and Arthur don't approve.

Good suggestion.

docs/Training-Self-Play.md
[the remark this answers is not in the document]
vincentpierre: `See the trainer configuration and agent prefabs for our Tennis environment for an example.`

ml-agents/mlagents/trainers/behavior_id_utils.py
andrewcoh: Pushed.

ml-agents/mlagents/trainers/trainer_util.py
andrewcoh: GhostController is not 1:1 with a trainer.  A reference to the ghost_controller is given to each ghost trainer when it's created.  On every advance, the ghost trainers query the GhostController for the learning team (only one team is learning at a time). The trainers pass its current training step as part of the query. When the training step exceeds the swap threshold, the next team in the queue becomes the learning team.  All the other trainers get this new team upon querying.

The trainer knows which team id is learning. The ghost trainer simply iterates through its queues. If the trajectory queue/policy queue match the learning team, it passes the trajectory to the wrapped trainer and updated policy to the environment.  Otherwise, it ignores trajectories and sends ghost policies to the environment.

andrewcoh: To your second point, yes it's always created and no it's not used unless there are ghost trainers.  I will figure out a better way to do this.

ml-agents/mlagents/trainers/ghost/trainer.py
andrewcoh: The GhostTrainer won't know the ELO ratings of policies associated with another ghost trainer, only its own.  The ghost controller is centralized.

ml-agents/mlagents/trainers/behavior_id_utils.py
```diff
@@ -1,10 +1,10 @@
-from typing import Dict, NamedTuple
+from typing import NamedTuple
 
 
 class BehaviorIdentifiers(NamedTuple):
```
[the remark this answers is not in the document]
andrewcoh: Done.

ml-agents/mlagents/trainers/ghost/controller.py
```diff
@@ -0,0 +1,48 @@
+from typing import Deque, Dict
+from collections import deque
+from mlagents.trainers.ghost.trainer import GhostTrainer
+
+
+class GhostController(object):
```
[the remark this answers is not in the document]
andrewcoh: Done.

ml-agents/mlagents/trainers/learn.py
andrewcoh: How is "team-change"? Check it out now @vincentpierre

ml-agents/mlagents/trainers/ghost/trainer.py
andrewcoh: done

ml-agents/mlagents/trainers/ghost/trainer.py
andrewcoh: done

docs/Training-Self-Play.md
```diff
@@ -1,21 +1,40 @@
...
-ML-Agents provides the functionality to train symmetric, adversarial games with [Self-Play](https://openai.com/blog/competitive-self-play/).
-A symmetric game is one in which opposing agents are *equal* in form and function. In reinforcement learning,
-this means both agents have the same observation and action spaces.
-With self-play, an agent learns in adversarial games by competing against fixed, past versions of itself
-to provide a more stable, stationary learning environment. This is compared
-to competing against its current self in every episode, which is a constantly changing opponent.
+ML-Agents provides the functionality to train both symmetric and asymmetric adversarial games with
+[Self-Play](https://openai.com/blog/competitive-self-play/).
+A symmetric game is one in which opposing agents are equal in form, function snd objective. Examples of symmetric games
+are Tennis and Soccer. In reinforcement learning, this means both agents have the same observation and
```
ervteng: nit: "are the Tennis and Soccer example environments".

ml-agents/mlagents/trainers/behavior_id_utils.py
```diff
@@ -1,36 +1,46 @@
...
+    """
+    BehaviorIdentifiers is a named tuple if the identifiers that uniquely distinguish
+    an agent encountered in the trainer_controller. The named tuple consists of the
+    fully qualified behavior name, the name of the brain name (corresponds to trainer
+    in the trainer controller) and the team id.  In the future, this can be extended
+    to support further identifiers.
+    """
+
+    behavior_id: str
     brain_name: str
```
ervteng: we should call this behavior_name not brain_name (to match C#). What's behavior_id?

ml-agents/mlagents/trainers/learn.py
```diff
@@ -101,6 +101,14 @@ def _create_parser():
         type=int,
         help="Number of parallel environments to use for training",
     )
+
+    argparser.add_argument(
```
ervteng: Would prefer this to not be a CLI command as it affects re-runability. But not obvious where it would go as it would not be inside either behavior's config. @awjuliani @andrewcoh is it time we re-invent the `trainer_config.yaml`?

andrewcoh: I'm actually considering making this a per trainer parameter so that they can be different for different teams.  Does it make sense that for some games we might want to train one agent team longer than another? An example I can think of is a team of different agents vs a boss. Training the boss might be super simple since it's just one agent but training the team might require learning tons of complex dynamics.  It might make sense to train the team for 1M timesteps for every 100K steps the boss is trained.

ml-agents/mlagents/trainers/behavior_id_utils.py
```diff
@@ -1,36 +1,46 @@
...
+        team_id: int = 0
+        if "team" in ids:
+            team_id = int(ids["team"][0])
         return BehaviorIdentifiers(
-            name_behavior_id=name_behavior_id, brain_name=name, behavior_ids=ids
+            behavior_id=name_behavior_id, brain_name=name, team_id=team_id
         )
+
+
+def create_name_behavior_id(name: str, team_id: int) -> str:
```
chriselion: Is this used anywhere? Would it be better as a method (or property) of BehaviorIdentifiers?

docs/Training-Self-Play.md
```diff
@@ -37,31 +57,68 @@ The ELO calculation (discussed below) depends on this final reward being either
...
+for precisely the same number of steps. Note, this is not specified in the trainer configuration yaml file, but as a command line argument.
+
+A larger value of `team-change` will allow the agent to train longer against it's opponents.  The longer an agent trains against the same set of opponents
+the more able it will be to defeat them. However, training against them for too long may result in overfitting to the particular opponent strategies
+and so the agent may fail against the next batch of opponents.
+
+The value of `team-change` will determine how many snapshots of the agent's policy are saved to be used as opponents for the other team.  So, we
+recommend setting this value as a function of the `save_steps` parameter which is discussed in the next section.
+
+Recommended Range : 4x-10x where x=`save_steps`
```
chriselion: Instead of specifying the value here, would it be easier to specify it as a multiple of `save_steps`?

ml-agents/mlagents/trainers/behavior_id_utils.py
andrewcoh: It's used [here](https://github.com/Unity-Technologies/ml-agents/blob/self-play-mutex/ml-agents/mlagents/trainers/ghost/trainer.py#L258) and [here](https://github.com/Unity-Technologies/ml-agents/blob/self-play-mutex/ml-agents/mlagents/trainers/ghost/trainer.py#L384).  In both instances, it's used so that the correct policies are pushed onto the correct queues if the learning team changes right before a swap.  I'm not sure it's appropriate to be a method/property of BehaviorIdentifiers because it's not really operating on data contained in a BehaviorIdentifier tuple.

docs/Training-Self-Play.md
andrewcoh: I don't have strong feelings either way.  If you think that's more intuitive then that works for me.

config/trainer_config.yaml
```diff
@@ -137,9 +137,10 @@ Tennis:
     time_horizon: 1000
     self_play:
         window: 10
-        play_against_current_self_ratio: 0.5
+        play_against_current_best_ratio: 0.5
```
vincentpierre: It is not the current best is it? It is the latest model right?

docs/Training-Self-Play.md
```diff
@@ -1,21 +1,40 @@
...
 
-ML-Agents provides the functionality to train symmetric, adversarial games with [Self-Play](https://openai.com/blog/competitive-self-play/).
-A symmetric game is one in which opposing agents are *equal* in form and function. In reinforcement learning,
-this means both agents have the same observation and action spaces.
-With self-play, an agent learns in adversarial games by competing against fixed, past versions of itself
-to provide a more stable, stationary learning environment. This is compared
-to competing against its current self in every episode, which is a constantly changing opponent.
+ML-Agents provides the functionality to train both symmetric and asymmetric adversarial games with
+[Self-Play](https://openai.com/blog/competitive-self-play/).
+A symmetric game is one in which opposing agents are equal in form, function snd objective. Examples of symmetric games
```
vincentpierre:
```suggestion
A symmetric game is one in which opposing agents are equal in form, function and objective. Examples of symmetric games
```

Can you run a spell check on the documentation once you are done modifying it to catch those?

docs/Training-Self-Play.md
```diff
@@ -1,21 +1,40 @@
...
+arguments and trainer configurations!
+
+With self-play, an agent learns in adversarial games by competing against fixed, past versions of its opponent
+(which could be itself as in symmetric games) to provide a more stable, stationary learning environment. This is compared
+to competing against the current, best opponent in every episode, which is constantly changing (because it's learning).
 
 Self-play can be used with our implementations of both [Proximal Policy Optimization (PPO)](Training-PPO.md) and [Soft Actor-Critc (SAC)](Training-SAC.md).
+However, from the perspective of an individual agent, these scenarios appear to have non-stationary dynamics because the opponent is often changing.
+This can cause significant issues in the experience replay mechanism used by SAC. Thus, we recommend that users use PPO. For further reading on
+this issue in particular, [see this paper](https://arxiv.org/pdf/1702.08887.pdf).
```
vincentpierre: I think you should name the paper

docs/Training-Self-Play.md
```diff
@@ -1,21 +1,40 @@
...
 For more general information on training with ML-Agents, see [Training ML-Agents](Training-ML-Agents.md).
 For more algorithm specific instruction, please see the documentation for [PPO](Training-PPO.md) or [SAC](Training-SAC.md).
 
 Self-play is triggered by including the self-play hyperparameter hierarchy in the trainer configuration file.  Detailed description of the self-play hyperparameters are contained below. Furthermore, to distinguish opposing agents, set the team ID to different integer values in the behavior parameters script on the agent prefab.
 
 ![Team ID](images/team_id.png)
 
-See the trainer configuration and agent prefabs for our Tennis environment for an example.
+***Team ID must be 0 or an integer greater than 0. Negative numbers will cause unpredictable behavior.***
```
vincentpierre: Can we enforce positive team id in the C# inspector?

docs/Training-Self-Play.md
```diff
@@ -39,29 +59,68 @@ The reward signal should still be used as described in the documentation for the
...
+The value of `team-change` will determine how many snapshots of the agent's policy are saved to be used as opponents for the other team.  So, we
+recommend setting this value as a function of the `save_steps` parameter discussed previously.
 
-The `swap_steps` parameter corresponds to the number of *trainer steps* between swapping the opponents policy with a different snapshot. As in the `save_steps` discussion, note that trainer steps are counted per agent. For more information, please see the [migration doc](Migrating.md) after v0.13.
+Recommended Range : 4x-10x where x=`save_steps`
 
 
+### Swap Steps
+
+The `swap_steps` parameter corresponds to the number of *ghost steps* (note, not trainer steps) between swapping the opponents policy with a different snapshot.
```
vincentpierre:
```suggestion
The `swap_steps` parameter corresponds to the number of *ghost steps* (not trainer steps) between swapping the opponents policy with a different snapshot.
```

docs/Training-Self-Play.md
```diff
@@ -39,29 +59,68 @@ The reward signal should still be used as described in the documentation for the
...
+A 'ghost step' refers to a step taken by an agent *that is following a fixed policy and not learning*. The reason for this distinction is that in asymmetric games,
+we may have teams with an unequal number of agents e.g. the 2v1 scenario in our Strikers Vs Goalie environment. The team with two agents collects
+twice as many agent steps per environment step as the team with one agent.  Thus, these two values will need to be distinct to ensure that the same number
+of trainer steps corresponds to the same number of opponent swaps for each team. The formula for `swap_steps` if
+a user desires `x` swaps of a team with `num_agents` agents against an opponent team with `num_opponent_agents`
+agents during `team-change` total steps is:
+
+```
+swap_steps = (num_agents / num_opponent_agents) * (team_change / x)
```
vincentpierre: Should we be doing the math in the code? I think math is hard...

docs/Training-Self-Play.md
```diff
@@ -39,29 +59,68 @@ The reward signal should still be used as described in the documentation for the
...
-### Play against current self ratio
+### Play against current best ratio
 
-The `play_against_current_self_ratio` parameter corresponds to the probability
-an agent will play against its ***current*** self. With probability
-1 - `play_against_current_self_ratio`, the agent will play against a snapshot of itself
-from a past iteration.
+The `play_against_current_best_ratio` parameter corresponds to the probability
+an agent will play against the current opponent. With probability
+1 - `play_against_current_best_ratio`, the agent will play against a snapshot of its
```
vincentpierre:
```suggestion
`1 - play_against_current_best_ratio`, the agent will play against a snapshot of its
```

docs/Training-Self-Play.md
```diff
@@ -39,29 +59,68 @@ The reward signal should still be used as described in the documentation for the
...
+opponent from a past iteration.
```
vincentpierre: from the past or the current best?

docs/Training-Self-Play.md
```diff
@@ -39,29 +59,68 @@ The reward signal should still be used as described in the documentation for the
...
 
-A larger value of `play_against_current_self_ratio` indicates that an agent will be playing against itself more often. Since the agent is updating it's policy, the opponent will be different from iteration to iteration.  This can lead to an unstable learning environment, but poses the agent with an [auto-curricula](https://openai.com/blog/emergent-tool-use/) of more increasingly challenging situations which may lead to a stronger final policy.
+A larger value of `play_against_current_best_ratio` indicates that an agent will be playing against the current opponent more often. Since the agent is updating it's policy, the opponent will be different from iteration to iteration.  This can lead to an unstable learning environment, but poses the agent with an [auto-curricula](https://openai.com/blog/emergent-tool-use/) of more increasingly challenging situations which may lead to a stronger final policy.
 
 Recommended Range : 0.0 - 1.0
```
vincentpierre: It is a ratio, of course it is between 0 and 1. Do you have a tighter window?

ml-agents/mlagents/trainers/ghost/controller.py
```diff
@@ -0,0 +1,90 @@
...
+                self._queue.append(team_id)
+
+    def get_learning_team(self) -> int:
+        """
+        Returns the current learning team.
+        :return: The learning team id
+        """
+        return self._learning_team
+
+    def finish_training(self, step: int) -> None:
```
vincentpierre: I do not think finish_training is an appropriate name given what the code is doing.

ml-agents/mlagents/trainers/ghost/controller.py
```diff
@@ -0,0 +1,90 @@
...
+        :param trainer: A GhostTrainer that manages this team_id.
+        """
+        if team_id not in self._ghost_trainers:
+            self._ghost_trainers[team_id] = trainer
+            if self._learning_team < 0:
+                self._learning_team = team_id
+            else:
+                self._queue.append(team_id)
+
+    def get_learning_team(self) -> int:
```
vincentpierre: make a property ?

ml-agents/mlagents/trainers/policy/tf_policy.py
```diff
@@ -144,6 +144,10 @@ def init_load_weights(self):
                 self.assign_ops.append(tf.assign(var, assign_ph))
 
     def load_weights(self, values):
+        if len(self.assign_ops) == 0:
+            logger.warning(
+                "Calling load_weights in tf_policy but assign_ops is empty. Did you forget to call init_load_weights?"
```
vincentpierre: Is this a warning or should it raise?

ml-agents/mlagents/trainers/ghost/controller.py
```diff
@@ -0,0 +1,90 @@
...
+from mlagents.trainers.ghost.trainer import GhostTrainer
+
+logger = logging.getLogger("mlagents.trainers")
+
+
+class GhostController(object):
+    """
+    GhostController contains a queue of team ids. GhostTrainers subscribe to the GhostController and query
+    it to get the current learning team.  The GhostController cycles through team ids every 'swap_interval'
+    which corresponds to the number of trainer steps between changing learning teams.
```
vincentpierre: Please mention this is a unique object and there can only be one in the training session.

docs/Training-Self-Play.md
andrewcoh: I think there are use cases for both extremes.  Maybe I should just remove the recommended range?

andrewcoh: or just "Range : 0.0 - 1.0"

ml-agents/mlagents/trainers/behavior_id_utils.py
andrewcoh: I'll do this in a separate PR and also change the use of brain_name/behavior_id/name_behavior_id  everywhere to behavior_name/fully_qualified_behavior_name

docs/Training-Self-Play.md
andrewcoh: We don't know how many agents are on each team

ml-agents/mlagents/trainers/behavior_id_utils.py
```diff
@@ -1,36 +1,46 @@
...
 
     @staticmethod
     def from_name_behavior_id(name_behavior_id: str) -> "BehaviorIdentifiers":
         """
-        Parses a name_behavior_id of the form name?team=0&param1=i&...
+        Parses a name_behavior_id of the form name?team=0
         into a BehaviorIdentifiers NamedTuple.
-        This allows you to access the brain name and distinguishing identifiers
-        without parsing more than once.
+        This allows you to access the brain name and team id og an agent
```
ervteng: team id of

ml-agents/mlagents/trainers/ghost/controller.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # Tracks last swap step for  each learning team because trainer
+        # steps of all GhostTrainers do not increment together
+        self._queue: Deque[int] = deque(maxlen=maxlen)
+        self._learning_team: int = -1
+        # Dict from team id to GhostTrainer for ELO calculation
+        self._ghost_trainers: Dict[int, GhostTrainer] = {}
+
+    @property
+    def get_learning_team(self) -> int:
```
ervteng: nit: I know we do this elsewhere in the code (and I want to remove it) but my feeling is that properties should be e.g. `learning_team` while `get_learning_team()` should be methods

ml-agents/mlagents/trainers/stats.py
```diff
@@ -113,19 +112,7 @@ def write_stats(
             )
             if self.self_play and "Self-play/ELO" in values:
                 elo_stats = values["Self-play/ELO"]
-                mean_opponent_elo = values["Self-play/Mean Opponent ELO"]
-                std_opponent_elo = values["Self-play/Std Opponent ELO"]
```
ervteng: Do we still want this info somewhere?

docs/Training-Self-Play.md
andrewcoh: replaced with latest_model. Does this clarify?

ml-agents/mlagents/trainers/ghost/controller.py
andrewcoh: Maybe save this for a separate house keeping PR? Along with the behavior name/brain name changes?

docs/Training-Self-Play.md
ervteng: Wouldn't say `unpredictable behavior` just say you can't do it. Or as Vince suggested enforce it.

ml-agents/mlagents/trainers/stats.py
andrewcoh: I'm not sure how valuable it is honestly. I don't think we need it.

ml-agents/mlagents/trainers/policy/tf_policy.py
andrewcoh: In the past, we've opted to raise a warning and not exceptions for things like this i.e. adding more than one policy to a ppo trainer.

ml-agents/mlagents/trainers/ghost/trainer.py
andrewcoh: I know that this isn't ideal but from investigating it doesn't seem to have caused any issues (yet).

docs/Training-Self-Play.md
```diff
@@ -1,21 +1,40 @@
...
 Self-play is triggered by including the self-play hyperparameter hierarchy in the trainer configuration file.  Detailed description of the self-play hyperparameters are contained below. Furthermore, to distinguish opposing agents, set the team ID to different integer values in the behavior parameters script on the agent prefab.
 
 ![Team ID](images/team_id.png)
 
-See the trainer configuration and agent prefabs for our Tennis environment for an example.
+***Team ID must be 0 or an integer greater than 0.***
+
+In symmetric games, since all agents (even on opposing teams) will share the same policy, they should have the same 'Behavior Name' in their
+Behavior Parameters Script.  In asymmetric games, they should have a different Behavior Name in their Behavior Parameters script.
+Note, in asymmetric games, the agents must have both different Behavior Names *and* different team IDs! Then, specify the trainer configuration
```
chriselion, line 31: So this means you can't have `zerg?teamId=0` and `protoss?teamId=0`? Is this a fundamental limitation? Sounds like something people are likely to get tripped up on.

If it's a removable restriction, don't let it block this PR, but can you log a jira for followup?

docs/Training-Self-Play.md
```diff
@@ -1,21 +1,40 @@
...
+this issue in particular, see the paper [Stabilising Experience Replay for Deep Multi-Agent Reinforcement Learning](https://arxiv.org/pdf/1702.08887.pdf).
 For more general information on training with ML-Agents, see [Training ML-Agents](Training-ML-Agents.md).
 For more algorithm specific instruction, please see the documentation for [PPO](Training-PPO.md) or [SAC](Training-SAC.md).
 
 Self-play is triggered by including the self-play hyperparameter hierarchy in the trainer configuration file.  Detailed description of the self-play hyperparameters are contained below. Furthermore, to distinguish opposing agents, set the team ID to different integer values in the behavior parameters script on the agent prefab.
 
 ![Team ID](images/team_id.png)
 
-See the trainer configuration and agent prefabs for our Tennis environment for an example.
+***Team ID must be 0 or an integer greater than 0.***
```
chriselion, line 27: Sorry, I know you said this was due to mypy and I never followed up with you on it. Similar to my other comment, if this was just done to make mypy happy, we can always get around that. Let's follow up on it afterwards.

docs/Training-Self-Play.md
chriselion: Not strong feelings, I just think it makes it easier to twist one knob at a time, instead of having to twist 2 in unison.

docs/Training-Self-Play.md
chriselion: Don't we know the number of steps coming in for each team though?

Also, would this need to be different if the agents and opponents were running at different decision intervals?

ml-agents/mlagents/trainers/behavior_id_utils.py
```diff
@@ -1,36 +1,46 @@
-from typing import Dict, NamedTuple
+from typing import NamedTuple
+from urllib.parse import urlparse, parse_qs
 
 
 class BehaviorIdentifiers(NamedTuple):
-    name_behavior_id: str
+    """
+    BehaviorIdentifiers is a named tuple if the identifiers that uniquely distinguish
```
chriselion, line 7:
```suggestion
    BehaviorIdentifiers is a named tuple of the identifiers that uniquely distinguish
```

ml-agents/mlagents/trainers/ghost/controller.py
```diff
@@ -0,0 +1,92 @@
+from mlagents_envs.logging_util import get_logger
+from typing import Deque, Dict
+from collections import deque
+from mlagents.trainers.ghost.trainer import GhostTrainer
+
+logger = get_logger(__name__)
+
+
+class GhostController(object):
```
chriselion, line 9:
```suggestion
class GhostController:
```
Not needed in python3

config/trainer_config.yaml
```diff
@@ -137,9 +137,10 @@ Tennis:
     time_horizon: 1000
     self_play:
         window: 10
-        play_against_current_self_ratio: 0.5
+        play_against_latest_model_ratio: 0.5
```
chriselion, line 140: Need to mention the parameter rename in the MigrationGuide (or check for the old name).

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -52,23 +81,48 @@ def __init__(
 
         self_play_parameters = trainer_parameters["self_play"]
         self.window = self_play_parameters.get("window", 10)
-        self.play_against_current_self_ratio = self_play_parameters.get(
```
vincentpierre, line 55: You should make sure this is between 0 and 1

docs/Training-Self-Play.md
andrewcoh: This will be something we support when we introduce a true multiagent trainer i.e. multiple behavior names that are on the same team.

config/trainer_config.yaml
andrewcoh: Added to the migration doc

docs/Training-Self-Play.md
andrewcoh: Basically, mypy wouldn't let me initialize the learning team ID int to None in the GhostController so I used -1.

ml-agents/mlagents/trainers/ghost/trainer.py
andrewcoh: Raising a warning

docs/Training-Self-Play.md
chriselion: You'd probably have to change the type to `Optional[int]` and handle the None case.

PR #3660, opened 2020-03-19
Stats SideChannel (for custom TensorBoard metrics)

chriselion:
### Proposed change(s)

Add a SideChannel that can be used to sent stats from C# to python, and later add those to TensorBoard.
Also adds methods to the Academy to find an instance (or instances) of specific SideChannel subclass (inspired by [MonoBehaviour.GetComponent(s)](https://docs.unity3d.com/ScriptReference/Component.GetComponent.html))

https://jira.unity3d.com/browse/MLA-675
https://forum.unity.com/threads/custom-graphs-in-tensorboard.843865/
https://github.com/Unity-Technologies/ml-agents/issues/2984
https://github.com/Unity-Technologies/ml-agents/pull/3594/files

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### Other comments
Sample graph of the number of time the 3DBallAgent drops the ball
image

### DIFF
--- a/Project/Assets/ML-Agents/Examples/FoodCollector/Scripts/FoodCollectorSettings.cs
+++ b/Project/Assets/ML-Agents/Examples/FoodCollector/Scripts/FoodCollectorSettings.cs
@@ -14,9 +14,12 @@ public class FoodCollectorSettings : MonoBehaviour
     public int totalScore;
     public Text scoreText;
 
+    StatsSideChannel m_statsSideChannel;
+
     public void Awake()
     {
         Academy.Instance.OnEnvironmentReset += EnvironmentReset;
+        m_statsSideChannel = Academy.Instance.GetSideChannel<StatsSideChannel>();
     }
 
     public void EnvironmentReset()
@@ -45,6 +48,13 @@ public class FoodCollectorSettings : MonoBehaviour
     public void Update()
     {
         scoreText.text = $"Score: {totalScore}";
-        Academy.Instance.GetSideChannel<StatsSideChannel>().AddStat("TotalScore", totalScore);
+
+        // Send stats via SideChannel so that they'll appear in TensorBoard.
+        // These values get averaged every summary_frequency steps, so we don't
+        // need to send every Update() call.
+        if ((Time.frameCount % 100)== 0)
+        {
+            m_statsSideChannel?.AddStat("TotalScore", totalScore);
+        }
     }
 }
--- a/Project/Assets/ML-Agents/Examples/FoodCollector/Scripts/FoodCollectorSettings.cs
+++ b/Project/Assets/ML-Agents/Examples/FoodCollector/Scripts/FoodCollectorSettings.cs
@@ -2,6 +2,7 @@ using System;
 using UnityEngine;
 using UnityEngine.UI;
 using MLAgents;
+using MLAgents.SideChannels;
 
 public class FoodCollectorSettings : MonoBehaviour
 {
@@ -44,5 +45,6 @@ public class FoodCollectorSettings : MonoBehaviour
     public void Update()
     {
         scoreText.text = $"Score: {totalScore}";
+        Academy.Instance.GetSideChannel<StatsSideChannel>().AddStat("TotalScore", totalScore);
     }
 }
--- a/com.unity.ml-agents/Runtime/Academy.cs
+++ b/com.unity.ml-agents/Runtime/Academy.cs
@@ -237,22 +237,28 @@ namespace MLAgents
 
         /// <summary>
         /// Returns the SideChannel of Type T if there is one registered, or null if it doesn't.
-        /// If there are multiple SideChannels of the same type registered, only the first found will be returned.
+        /// If there are multiple SideChannels of the same type registered, the returned instance is arbitrary.
         /// </summary>
         /// <typeparam name="T"></typeparam>
         /// <returns></returns>
         public T GetSideChannel<T>() where T: SideChannel
         {
-            return Communicator.GetSideChannel<T>();
+            return Communicator?.GetSideChannel<T>();
         }
 
         /// <summary>
-        /// Returns all SideChannels of Type T that are registered.
+        /// Returns all SideChannels of Type T that are registered. Use <see cref=GetSideChannel()> if possible, as
+        /// that does not make any memory allocations.
         /// </summary>
         /// <typeparam name="T"></typeparam>
         /// <returns></returns>
         public List<T> GetSideChannels<T>() where T: SideChannel
         {
+            if (Communicator == null)
+            {
+                // Make sure we return a non-null List.
+                return new List<T>();
+            }
             return Communicator.GetSideChannels<T>();
         }
 
--- a/com.unity.ml-agents/Runtime/Academy.cs
+++ b/com.unity.ml-agents/Runtime/Academy.cs
@@ -247,8 +247,8 @@ namespace MLAgents
         }
 
         /// <summary>
-        /// Returns all SideChannels of Type T that are registered. Use <see cref=GetSideChannel()> if possible, as
-        /// that does not make any memory allocations.
+        /// Returns all SideChannels of Type T that are registered. Use <see cref="GetSideChannel{T}()"/> if possible,
+        /// as that does not make any memory allocations.
         /// </summary>
         /// <typeparam name="T"></typeparam>
         /// <returns></returns>
--- a/com.unity.ml-agents/Runtime/Academy.cs
+++ b/com.unity.ml-agents/Runtime/Academy.cs
@@ -236,6 +236,27 @@ namespace MLAgents
         }
 
         /// <summary>
+        /// Returns the SideChannel of Type T if there is one registered, or null if it doesn't.
+        /// If there are multiple SideChannels of the same type registered, only the first found will be returned.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <returns></returns>
+        public T GetSideChannel<T>() where T: SideChannel
+        {
+            return Communicator.GetSideChannel<T>();
+        }
+
+        /// <summary>
+        /// Returns all SideChannels of Type T that are registered.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <returns></returns>
+        public List<T> GetSideChannels<T>() where T: SideChannel
+        {
+            return Communicator.GetSideChannels<T>();
+        }
+
+        /// <summary>
         /// Disable stepping of the Academy during the FixedUpdate phase. If this is called, the Academy must be
         /// stepped manually by the user by calling Academy.EnvironmentStep().
         /// </summary>
@@ -334,6 +355,7 @@ namespace MLAgents
             {
                 Communicator.RegisterSideChannel(new EngineConfigurationChannel());
                 Communicator.RegisterSideChannel(floatProperties);
+                Communicator.RegisterSideChannel(new StatsSideChannel());
                 // We try to exchange the first message with Python. If this fails, it means
                 // no Python Process is ready to train the environment. In this case, the
                 //environment must use Inference.
--- a/com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
+++ b/com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
@@ -538,6 +538,43 @@ namespace MLAgents
         }
 
         /// <summary>
+        /// Returns the SideChannel of Type T if there is one registered, or null if it doesn't.
+        /// If there are multiple SideChannels of the same type registered, only the first found will be returned.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <returns></returns>
+        public T GetSideChannel<T>() where T: SideChannel
+        {
+            foreach (var sc in m_SideChannels.Values)
+            {
+                if (sc.GetType() == typeof(T))
+                {
+                    return (T) sc;
+                }
+            }
+            return null;
+        }
+
+        /// <summary>
+        /// Returns all SideChannels of Type T that are registered.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <returns></returns>
+        public List<T> GetSideChannels<T>() where T: SideChannel
+        {
+            var output = new List<T>();
+
+            foreach (var sc in m_SideChannels.Values)
+            {
+                if (sc.GetType() == typeof(T))
+                {
+                    output.Add((T) sc);
+                }
+            }
+            return output;
+        }
+
+        /// <summary>
         /// Grabs the messages that the registered side channels will send to Python at the current step
         /// into a singe byte array.
         /// </summary>
--- a/com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
+++ b/com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
@@ -537,12 +537,7 @@ namespace MLAgents
             }
         }
 
-        /// <summary>
-        /// Returns the SideChannel of Type T if there is one registered, or null if it doesn't.
-        /// If there are multiple SideChannels of the same type registered, only the first found will be returned.
-        /// </summary>
-        /// <typeparam name="T"></typeparam>
-        /// <returns></returns>
+        /// <inheritdoc/>
         public T GetSideChannel<T>() where T: SideChannel
         {
             foreach (var sc in m_SideChannels.Values)
@@ -555,11 +550,7 @@ namespace MLAgents
             return null;
         }
 
-        /// <summary>
-        /// Returns all SideChannels of Type T that are registered.
-        /// </summary>
-        /// <typeparam name="T"></typeparam>
-        /// <returns></returns>
+        /// <inheritdoc/>
         public List<T> GetSideChannels<T>() where T: SideChannel
         {
             var output = new List<T>();
--- a/com.unity.ml-agents/Runtime/SideChannels/EngineConfigurationChannel.cs
+++ b/com.unity.ml-agents/Runtime/SideChannels/EngineConfigurationChannel.cs
@@ -13,7 +13,7 @@ namespace MLAgents.SideChannels
         /// <summary>
         /// Initializes the side channel.
         /// </summary>
-        public EngineConfigurationChannel()
+        internal EngineConfigurationChannel()
         {
             ChannelId = new Guid(k_EngineConfigId);
         }
--- a/com.unity.ml-agents/Runtime/SideChannels/EngineConfigurationChannel.cs
+++ b/com.unity.ml-agents/Runtime/SideChannels/EngineConfigurationChannel.cs
@@ -8,10 +8,11 @@ namespace MLAgents.SideChannels
     /// </summary>
     public class EngineConfigurationChannel : SideChannel
     {
-        private const string k_EngineConfigId = "e951342c-4f7e-11ea-b238-784f4387d1f7";
+        const string k_EngineConfigId = "e951342c-4f7e-11ea-b238-784f4387d1f7";
 
         /// <summary>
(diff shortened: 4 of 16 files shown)


com.unity.ml-agents/Runtime/Academy.cs
```diff
@@ -235,6 +235,27 @@ public void UnregisterSideChannel(SideChannel channel)
             Communicator?.UnregisterSideChannel(channel);
         }
 
+        /// <summary>
+        /// Returns the SideChannel of Type T if there is one registered, or null if it doesn't.
+        /// If there are multiple SideChannels of the same type registered, only the first found will be returned.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <returns></returns>
+        public T GetSideChannel<T>() where T: SideChannel
```
chriselion: @vincentpierre @surfnerd I think we had talked about this before. It seems like a clean way to get access to sidechannels without adding a custom method to Academy for each one.

com.unity.ml-agents/Runtime/SideChannels/EngineConfigurationChannel.cs
```diff
@@ -13,7 +13,7 @@ public class EngineConfigurationChannel : SideChannel
         /// <summary>
         /// Initializes the side channel.
         /// </summary>
-        public EngineConfigurationChannel()
+        internal EngineConfigurationChannel()
```
chriselion: Since this is basically a singleton, I don't think it should have a public constructor.

ml-agents/mlagents/trainers/agent_processor.py
```diff
@@ -258,3 +258,8 @@ def __init__(
             self.behavior_id
         )
         self.publish_trajectory_queue(self.trajectory_queue)
+
+    def set_environment_stats(self, env_stats: Dict[str, float]) -> None:
+        for stat_name, val in env_stats.items():
+            # full_stat = f"Environment/{stat_name}"
+            self.stats_reporter.set_stat(stat_name, val)
```
chriselion: Followup from the design review: we can pretty easily add averaging of multiple values by using `add_stat` here instead. I was thinking about:
* Adding on optional enum arg to AddStats() on C# for AggregationType
* Adding the AggregationType to the message
* using (name, AggregationType) as the dictionary key
* deciding between set_stat or add_stat here based on the AggregationType

com.unity.ml-agents/Runtime/Academy.cs
chriselion: I can replace Academy.FloatProperties with this later if we're happy with it.

com.unity.ml-agents/Runtime/Academy.cs
```diff
@@ -235,6 +235,27 @@ public void UnregisterSideChannel(SideChannel channel)
...
+        {
+            return Communicator.GetSideChannel<T>();
+        }
+
+        /// <summary>
+        /// Returns all SideChannels of Type T that are registered.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <returns></returns>
+        public List<T> GetSideChannels<T>() where T: SideChannel
```
vincentpierre: Method name is too close to GetSideChannel IMO. I would say let's keep only this one as it is more general.

com.unity.ml-agents/Runtime/SideChannels/EngineConfigurationChannel.cs
mmattar: Might be worth adding that as a comment to this constructor and the one for stats side channel.

com.unity.ml-agents/Runtime/SideChannels/StatsSideChannel.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+namespace MLAgents.SideChannels
+{
+    public class StatsSideChannel : SideChannel
```
mmattar: Would TensorBoardMetricsSideChannel be a more appropriate name?

com.unity.ml-agents/Runtime/SideChannels/StatsSideChannel.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// <summary>
+        /// Initializes the side channel with the provided channel ID.
+        /// </summary>
+        internal StatsSideChannel()
+        {
+            ChannelId = new Guid(k_StatsSideChannelDefaultId);
+        }
+
+        /// <summary>
+        /// Add a stat value for reporting. This will appear in the Tensorboard summary and trainer gauges.
```
mmattar: This feels like great documentation for the class, not just this method.

com.unity.ml-agents/Runtime/SideChannels/StatsSideChannel.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// You can nest stats in Tensorboard with "/".
+        /// Note that stats are only written to Tensorboard each summary_frequency steps; if a stat is
+        /// received multiple times, only the most recent version is used.
+        /// To avoid conflicts between multiple environments, only stats from worker index 0 are used.
+        /// </summary>
+        /// <param name="key"></param>
+        /// <param name="value"></param>
+        public void AddStat(string key, float value)
```
mmattar: If we rename the class based on the method above - this ought to be renamed to `AddMetricValue`

ml-agents/mlagents/trainers/env_manager.py
```diff
@@ -108,4 +109,11 @@ def _process_step_infos(self, step_infos: List[EnvironmentStep]) -> int:
                         name_behavior_id, ActionInfo.empty()
                     ),
                 )
+
+                # In order to prevent conflicts between multiple environments,
```
mmattar: This would be helpful documentation in the C# class too.

com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
```diff
@@ -537,6 +537,43 @@ public void UnregisterSideChannel(SideChannel sideChannel)
             }
         }
 
+        /// <summary>
```
mmattar: Do we expect the implementation of these two methods to be different for different ICommunicator implementations?

ml-agents-envs/mlagents_envs/side_channel/stats_side_channel.py
```diff
@@ -0,0 +1,22 @@
...
+
+class StatsSideChannel(SideChannel):
+    def __init__(self) -> None:
+        # >>> uuid.uuid5(uuid.NAMESPACE_URL, "com.unity.ml-agents/StatsSideChannel")
+        # UUID('a1d8f7b7-cec8-50f9-b78b-d3e165a78520')
+        super().__init__(uuid.UUID("a1d8f7b7-cec8-50f9-b78b-d3e165a78520"))
+
+        self.stats: Dict[str, float] = {}
+
+    def on_message_received(self, msg: IncomingMessage) -> None:
```
vincentpierre: docstriing on these methods and class?

ml-agents/mlagents/trainers/agent_processor.py
```diff
@@ -258,3 +258,8 @@ def __init__(
             self.behavior_id
         )
         self.publish_trajectory_queue(self.trajectory_queue)
+
+    def set_environment_stats(self, env_stats: Dict[str, float]) -> None:
+        for stat_name, val in env_stats.items():
+            # full_stat = f"Environment/{stat_name}"
```
vincentpierre: This looks like dead code, if it is not, maybe add an extra comment on this line about in which situations this should be considered.

ml-agents/mlagents/trainers/agent_processor.py
```diff
@@ -258,3 +258,8 @@ def __init__(
             self.behavior_id
         )
         self.publish_trajectory_queue(self.trajectory_queue)
+
+    def set_environment_stats(self, env_stats: Dict[str, float]) -> None:
```
vincentpierre: docstring on this method (even though it is not public)
Is this method only ever going to be used with TensorBoardStats ? If it is the case, the name should reflect that. If not, add a comment explaining what should call this method.

ml-agents/mlagents/trainers/env_manager.py
vincentpierre: I am rather hesitant on this. Are the stats only usable when training? Is it not possible for users to get the C# stats when using the LL-API? If it is possible to use it with LL-API, then C# should not assume how the stats are consumed (at least not without a warning of type : "When using ml-agents training ™️, the stats will...")

com.unity.ml-agents/Runtime/SideChannels/StatsSideChannel.cs
ervteng: It won't just be for TB - with this change it'll show up anywhere Python stats are written (e.g. CSV, timers, and perhaps later the Console).

ml-agents/mlagents/trainers/agent_processor.py
ervteng: 👍 
Almost want to say that the default behavior (without the enum) is `add_stat` just from the types of things MikeWise recorded (# of food eaten, etc.) and what kinds of things game developers are likely to record (score, coins collected, ranking, etc.). But hard to say.

ml-agents/mlagents/trainers/agent_processor.py
ervteng: Can we call this method `record_environment_stats`? Since under the hood we might be doing both `set` and `add` for different stats

ml-agents/mlagents/trainers/tests/test_subprocess_env_manager.py
```diff
@@ -147,7 +147,7 @@ def test_advance(self, mock_create_worker, external_brains_mock, step_mock):
         env_manager.set_agent_manager(brain_name, agent_manager_mock)
 
         step_info_dict = {brain_name: Mock()}
-        step_info = EnvironmentStep(step_info_dict, 0, action_info_dict)
+        step_info = EnvironmentStep(step_info_dict, 0, action_info_dict, {})
```
ervteng: We probably should put something in the dictionary so that the method gets tested, and similarly run `set_environment_stats` in the `test_agent_processor.py` tests

ml-agents/mlagents/trainers/agent_processor.py
chriselion: So do we still only want to keep worker 0 if the stats are being averaged? Seems like for that case, we'd want them all,

com.unity.ml-agents/Runtime/Academy.cs
chriselion: Discussed offline; should be fine since this is what MonoBehavior does for GetComponent.

com.unity.ml-agents/Runtime/SideChannels/StatsSideChannel.cs
chriselion: Yeah, I kept the name generic but called out TensorBoard in the comments. I'll also add an example of this in the "Using TensorBoard" docs.

com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
chriselion: The implementation will depend on how the SideChannels are stored by the ICommunicator implementation. There's no guarantee that another implementation would use a UUID->SideChannel dictionary.

We could make this an internal static method in SideChannel.cs if you think it's likely to be reusable.

com.unity.ml-agents/Runtime/SideChannels/StatsSideChannel.cs
mmattar: Ok, let's then document the different ways that adding a _stat_ here will bubble up on the Python side. I find the term stats odd since TB refers to metrics, but I also can't see a better alternative (esp that it does align with timer stats)

ml-agents/mlagents/trainers/agent_processor.py
chriselion: Updated this to
```
for stat_name, (val, agg_type) in env_stats.items():
    if agg_type == StatsAggregationMethod.AVERAGE:
        self.stats_reporter.add_stat(stat_name, val)
    elif agg_type == StatsAggregationMethod.MOST_RECENT:
        # In order to prevent conflicts between multiple environments,
        # only stats from the first environment are recorded.
        if worker_id == 0:
            self.stats_reporter.set_stat(stat_name, val)
```
I think that does the least-surprising thing in each case. I guess we could have a third enum value like MOST_RECENT_PRIMARY_WORKER that checks the worker_id, and make MOST_RECENT always set, but I don't think those will get used much anyway.

ml-agents/mlagents/trainers/agent_processor.py
chriselion: > Is this method only ever going to be used with TensorBoardStats
I calls the StatsReporter, which writes to multiple locations, but I think TensorBoard is the main usecase.

Renamed to record_environment_stats() and added docstring

ml-agents/mlagents/trainers/agent_processor.py
chriselion: Done

com.unity.ml-agents/Runtime/SideChannels/StatsSideChannel.cs
chriselion: I moved the comments below up here and clarified a bit.

com.unity.ml-agents/Runtime/SideChannels/StatsSideChannel.cs
chriselion: Done

ml-agents/mlagents/trainers/env_manager.py
chriselion: I moved it to a comment on StatAggregationMethod.MostRecent:
> To avoid conflicts between multiple environments, the ML Agents environment will only
> keep stats from worker index 0.

ml-agents-envs/mlagents_envs/side_channel/stats_side_channel.py
chriselion: Done

com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
```diff
@@ -537,6 +537,43 @@ public void UnregisterSideChannel(SideChannel sideChannel)
...
+        /// Returns the SideChannel of Type T if there is one registered, or null if it doesn't.
+        /// If there are multiple SideChannels of the same type registered, only the first found will be returned.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <returns></returns>
+        public T GetSideChannel<T>() where T: SideChannel
+        {
+            foreach (var sc in m_SideChannels.Values)
```
surfnerd: if there are multiple side channels of the same type but with different GUIDs, what happens?

com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
```diff
@@ -537,6 +537,43 @@ public void UnregisterSideChannel(SideChannel sideChannel)
...
+            }
+            return null;
+        }
+
+        /// <summary>
+        /// Returns all SideChannels of Type T that are registered.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <returns></returns>
+        public List<T> GetSideChannels<T>() where T: SideChannel
```
surfnerd: this makes more sense to me

com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
chriselion: Oops, I updated the comments on the Academy method, but forgot to update here (and ICommunicator). The Academy ones now say:
> /// If there are multiple SideChannels of the same type registered, the returned instance is arbitrary.

Realistically, it'll be the first one that we find.

What does MonoBehavior.GetComponent<>() do if you search for a type that has multiple instances attached? I assume it does something similar. I guess we could keep searching and warn if we find another matching instance...

surfnerd: yeah, it actually returns the first one from the order of what you see in the inspector.  As long as it's called out in docs.

ml-agents/mlagents/trainers/tests/test_subprocess_env_manager.py
chriselion: Done

chriselion:
> I wonder how you got the graph for the number of times the ball got dropped. Are you keeping trak of the total number in C#?

I hacked that as a static for testing :)

chriselion:
Mind taking another look at this? It changed a bit, mostly around the multiple aggregation types.

ml-agents-envs/mlagents_envs/side_channel/stats_side_channel.py
```diff
@@ -0,0 +1,48 @@
+from mlagents_envs.side_channel import SideChannel, IncomingMessage
+import uuid
+from typing import Dict, Tuple
+from enum import Enum
+
+
+# Determines the behavior of how multiple stats within the same summary period are combined.
+class StatsAggregationMethod(Enum):
```
chriselion: Is there a better place for this? It needs to live in ml-agents-envs (unless we move the SideChannel?) but it feels strange importing `mlagents_envs.side_channel.stats_side_channel` in a lot of places just to get the enum. @ervteng thoughts?

ervteng: Hmm, this actually seems the least obtrusive since the StatsReporter lives in `mlagents`. The other option would be in the `stats.py`, but then we'd have to move that to `mlagents_envs`.

com.unity.ml-agents/Runtime/Academy.cs
vincentpierre: Let's replace Academy.Instance.FloatProperty then. Do you want to do it in this PR or make another one?

Project/Assets/ML-Agents/Examples/FoodCollector/Scripts/FoodCollectorSettings.cs
```diff
@@ -44,5 +45,6 @@ void ClearObjects(GameObject[] objects)
     public void Update()
     {
         scoreText.text = $"Score: {totalScore}";
+        Academy.Instance.GetSideChannel<StatsSideChannel>().AddStat("TotalScore", totalScore);
```
chriselion: @vincentpierre Added to FoodCollector here. If we want to add to other examples, let's do so in a followup PR.

com.unity.ml-agents/Runtime/Academy.cs
chriselion: Let's do it in another PR.

Project/Assets/ML-Agents/Examples/FoodCollector/Scripts/FoodCollectorSettings.cs
vincentpierre: Should we store the StatsSideChannel in a private property to avoid calling GetSideChannel every update?

chriselion: Yeah, probably - I'm not to worried about the performance, but we should set a good example...

com.unity.ml-agents/Runtime/Academy.cs
chriselion: https://jira.unity3d.com/browse/MLA-799

chriselion:
FWIW here's FoodCollector with the `contribute` flag set to true everywhere: 
image

PR #3669, opened 2020-03-22
Update ProjectSettingsOverrides.cs

sterlingcrispin:
### Proposed change(s)

Warning developers these are being changed during runtime and then changed back, I was reusing a scene and scratching my head trying to figure out why fixedDeltaTime had no effect , a warning would be helpful

### Types of change(s)

- [ X] Documentation update
- [ X] Other (please describe)


chriselion:
Sorry for the delay on this. I don't think we should be adding a console message for this, but I added a better description for ProjectSettingsOverrides in https://github.com/Unity-Technologies/ml-agents/pull/3838

Issue #3667, opened 2020-03-22
Base Port  with 3DBall example and getting started jupyter notebook


seboz123:
I am new to ml-agents and unity.
When I tried to load the environment in the getting-started notebook, I found that it wasn't working, because the environment took too long to respond. I found out, that you need to set the base_port in the instatiation of the UnityEnvironment to must be set to 5004, it is not set properly by default. Is this intended?

chriselion:
Hi,
I believe you're right; we currently default the base_port to 5005, regardless of whether an environment path is set or not. A better approach would be
*default base_port to None
*if base_port is None, use 5005 if file_name is specified, else DEFAULT_EDITOR_PORT (5004)

We have this logged in our internal tracker as MLA-709; I'll try to get a fix in for it soon...

chriselion:
PR to clean this up: https://github.com/Unity-Technologies/ml-agents/pull/3673

PR #3673, opened 2020-03-23
cleanup port logic in UnityEnvironment

chriselion:
### Proposed change(s)
@awjuliani hit this a few weeks ago, and it came up in an issue today too. If no environment was specified, we would still default to port 5005, but we should be smart enough to use 5004 (the editor port) in this case.

https://github.com/Unity-Technologies/ml-agents/issues/3667
https://jira.unity3d.com/browse/MLA-709?

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/docs/Custom-SideChannels.md
+++ b/docs/Custom-SideChannels.md
@@ -193,7 +193,7 @@ messages to the Unity environment from Python using it.
 string_log = StringLogChannel()
 
 # We start the communication with the Unity Editor and pass the string_log side channel as input
-env = UnityEnvironment(base_port=UnityEnvironment.DEFAULT_EDITOR_PORT, side_channels=[string_log])
+env = UnityEnvironment(side_channels=[string_log])
 env.reset()
 string_log.send_string("The environment was reset")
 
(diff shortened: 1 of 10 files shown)


docs/Custom-SideChannels.md
```diff
@@ -193,7 +193,7 @@ messages to the Unity environment from Python using it.
 string_log = StringLogChannel()
 
 # We start the communication with the Unity Editor and pass the string_log side channel as input
-env = UnityEnvironment(base_port=UnityEnvironment.DEFAULT_EDITOR_PORT, side_channels=[string_log])
+env = UnityEnvironment(side_channels=[string_log])
```
chriselion: Since these should now do the correct thing, I removed the explicit port (for simplicity and future-proofing) wherever I found it in the docs.

ml-agents-envs/mlagents_envs/communicator.py
```diff
@@ -8,8 +8,8 @@ def __init__(self, worker_id=0, base_port=5005):
         """
         Python side of the communication. Must be used in pair with the right Unity Communicator equivalent.
 
+        :int worker_id: Offset from base_port. Used for training multiple environments simultaneously.
```
chriselion: Rearranged to match the argument order. I think I changed the wording of this wherever I saw it (`git blame` showed it was 3 years old).

PR #3674, opened 2020-03-23
Fix editor port in Dockerfile

chriselion:
### Proposed change(s)
I noticed this while searching for hardcoded ports in https://github.com/Unity-Technologies/ml-agents/pull/3673.

5004 has been the editor port for a while now.

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -132,7 +132,7 @@ COPY ml-agents /ml-agents
 WORKDIR /ml-agents
 RUN pip install -e .
 
-# port 5005 is the port used in in Editor training.
-EXPOSE 5005
+# port 5004 is the port used in in Editor training.
+EXPOSE 5004
 
 ENTRYPOINT ["mlagents-learn"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -132,7 +132,9 @@ COPY ml-agents /ml-agents
 WORKDIR /ml-agents
 RUN pip install -e .
 
-# port 5004 is the port used in in Editor training.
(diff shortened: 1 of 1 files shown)


anupambhatnagar:
@chriselion My recommendation would be to open a broader range of ports. In my use case I often run multiple mlagents-learn with the same executable in a docker container so have a bigger range of ports to be open is useful.

Dockerfile
```diff
@@ -132,7 +132,7 @@ COPY ml-agents /ml-agents
 WORKDIR /ml-agents
 RUN pip install -e .
 
-# port 5005 is the port used in in Editor training.
-EXPOSE 5005
+# port 5004 is the port used in in Editor training.
+EXPOSE 5004
```
anupambhatnagar:
```suggestion
EXPOSE 5000-9999
```

anupambhatnagar:
```suggestion
EXPOSE 5000-9999
```

chriselion:
Is there any downside to opening up such a wide range? How about 5004-5050 to allow for the editor or several environments?

anupambhatnagar:
Downsides - not that I know of. I suggested a wider range since some common tools such as tensorboard use a default of 6006 and some other services like jupyter/spark/yarn/airflow use ports in the 8000 range.

5004-5050 sounds reasonable too.

PR #3680, opened 2020-03-24
Raise exceptions from environment subprocesses

harperj:
### Proposed change(s)

This commit surfaces exceptions from environment worker subprocesses,
and changes the SubprocessEnvManager to raise those exceptions when
caught.  Additionally TrainerController was changed to treat environment
exceptions differently than KeyboardInterrupts.  We now raise the
environment exceptions after exporting the model, so that ML-Agents will
correctly exit with a non-zero return code.

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/mlagents/trainers/subprocess_env_manager.py
+++ b/ml-agents/mlagents/trainers/subprocess_env_manager.py
@@ -3,7 +3,11 @@ from typing import Dict, NamedTuple, List, Any, Optional, Callable, Set, Tuple
 import cloudpickle
 
 from mlagents_envs.environment import UnityEnvironment
-from mlagents_envs.exception import UnityCommunicationException, UnityTimeOutException
+from mlagents_envs.exception import (
+    UnityCommunicationException,
+    UnityTimeOutException,
+    UnityEnvironmentException,
+)
 from multiprocessing import Process, Pipe, Queue
 from multiprocessing.connection import Connection
 from queue import Empty as EmptyQueueException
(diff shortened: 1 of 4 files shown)


ervteng:
Looks good to me - should we add a test that makes sure the exception is passed correctly up?

harperj:
@ervteng Good timing :-)

I was writing (and just pushed) a test showing we raise the exception from SubprocessEnvManager now.

ml-agents/mlagents/trainers/subprocess_env_manager.py
```diff
@@ -64,6 +68,9 @@ def send(self, name: str, payload: Any = None) -> None:
     def recv(self) -> EnvironmentResponse:
         try:
             response: EnvironmentResponse = self.conn.recv()
+            if response.name == "env_close":
```
mmattar: one suggestion: create a const variable for "env_close" as it's used in 2 places.

chriselion: or even better, an enum

harperj: Made `EnvironmentCommand` an enum with all of the different command types.  This required renaming the previous `EnvironmentCommand` class to `EnvironmentRequest`.  Changed references throughout accordingly 👍

mmattar: 🎉 nice.

PR #3687, opened 2020-03-25
Remove "docker target" feature

harperj:
### Proposed change(s)

The "docker target" feature and associated command-line flag
--docker-target-name were created for use with the now-deprecated
Docker setup.  This feature redirects the paths used by learn.py
for the environment and config files to be based from a directory
other than the current working directory.  Additionally it wrapped
the environment execution with xvfb-run.

This change removes the "docker target" feature because:

* Renaming the paths doesn't fix any problem.  Absolute paths can
  already be passed for configs and environment executables.
* Use of xserver, Xvfb, or xvfb-run are independent of mlagents-learn
  and can be used outside of the mlagents-learn call.  Further, xvfb-run
  is not the only solution for software rendering.
* The feature is confusing, since it doesn't actually involve the Docker tool
   at all.

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [x] Breaking change
- [x] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/docs/Using-Docker.md
+++ b/docs/Using-Docker.md
@@ -95,7 +95,6 @@ docker run -it --name <container-name> \
            -p 5005:5005 \
            -p 6006:6006 \
            <image-name>:latest \
-           --docker-target-name=unity-volume \
            <trainer-config-file> \
            --env=<environment-name> \
            --train \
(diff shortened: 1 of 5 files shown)


docs/Using-Docker.md
```diff
@@ -95,7 +95,6 @@ docker run -it --name <container-name> \
            -p 5005:5005 \
```
harperj: Despite this doc being deprecated, I've updated it in case users still find it and want to try it out.  This should also help us transition to a new Dockerfile in the future.

PR #3689, opened 2020-03-25
Test against older python and C# versions

chriselion:
### Proposed change(s)

Extends yamato tests to use older versions of python and C# codebase. 

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/tests/yamato/yamato_utils.py
+++ b/ml-agents/tests/yamato/yamato_utils.py
@@ -1,7 +1,6 @@
 import os
 import subprocess
 import yaml
-import time
 
 
 def get_unity_executable_path():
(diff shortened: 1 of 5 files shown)


chriselion:
Confirmed that 0.14.0 versions of C# and python failed (although not quite the way I would have expected):
https://github.com/Unity-Technologies/ml-agents/pull/3693
https://github.com/Unity-Technologies/ml-agents/pull/3694

chriselion:
@vincentpierre followup from the meeting - 0.14 used the old format for the communication protocol versioning, so there's no expectation that it'll work with master (or 0.15):
https://github.com/Unity-Technologies/ml-agents/blob/0.14.0/ml-agents-envs/mlagents_envs/environment.py#L54

ml-agents/tests/yamato/yamato_utils.py
```diff
@@ -49,21 +49,62 @@ def run_standalone_build(base_path: str, verbose: bool = False) -> int:
...
+        subprocess.check_call(
+            f"git checkout {csharp_version} -- {csharp_dir}", shell=True
+        )
+
+
+def undo_git_checkout():
+    """
+    Clean up the git working directory.
+    """
+    subprocess.check_call("git reset HEAD .", shell=True)
```
chriselion: Number of times I accidentally reverted changes to this file while testing these commands: 2

surfnerd: I created a git utility class while working at nvidia in order to be able to mock out the git calls for testing.  saved my life.

chriselion: YAGNI for now, but it if we need to do more git logic, I'll look into it.

PR #3690, opened 2020-03-25
[refactor] Run Trainers in separate threads

ervteng:
### Proposed change(s)
This began as a fix to SAC's `num_updates` functionality. and evolved into a bit more than that.

#### Changes to SAC and why they were necessary 

Currently SAC updates each time advance() is called, unless `train_interval` is specified, in which case it updates every train_interval steps. `num_updates` refers to the number of batches sampled and used per updates. So with these two parameters you can control how often updates happen relative to steps.

The problem becomes when you scale the number of areas or envs. Now all of a sudden we have a much greater number of steps per update, and the parameter doesn’t scale. 

This PR introduces a `steps_per_update` parameter to SAC. SAC will keep track of the number of steps it receives when a trajectory comes in, and do the appropriate number of updates. HOWEVER, this produced a bad user experience, as the game would randomly "freeze" to do training. 

#### Further changes

To get around this problem, this PR also puts the Trainers' advance() calls in a separate thread. We do this by making AgentManagerQueue a thread-safe blocking Queue, and semi-blocking (get with timeout) on the queues in advance()

This has the added benefit of (as measured on 3DBall, single environment):

- 15% step throughput improvement on PPO
- 13% step throughput improvement on SAC

**Note that this slightly breaks the on-policyness of PPO**; steps taken during a buffer update will use the in-between parameterizations (e.g. a step taken after the first batch will have that parameterization, after the 2nd batch will have the next, and so on). I've made threading disableable easily with a flag in `TrainerController` so we could turn it on/off if we see significant degradation in performance.

~TODO: test Ghost Trainer, change documentation.~ Test with larger number of envs. 
~Future: Switch to full multiprocess (requires a rethink of the StatsReporter shared between the trainer and the AgentProcessor).~ 

### Types of change(s)

- [x] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


docs/Training-SAC.md
```diff
@@ -106,17 +105,16 @@ there may not be any new interesting information between steps, and `train_inter
 
 Typical Range: `1` - `5`
 
-### Number of Updates
+### Steps Per Update
 
-`num_update` corresponds to the number of mini batches sampled and used for training during each
-training event. In SAC, a single "update" corresponds to grabbing a batch of size `batch_size` from the experience
-replay buffer, and using this mini batch to update the models. Typically, this can be left at 1.
+`steps_per_update` corresponds to the number agent steps (actions) taken for each mini-batch sampled and used during training. In SAC, a single "update" corresponds to grabbing a batch of size `batch_size` from the experience
```
andrewcoh, line 110: number of agent steps

docs/Training-SAC.md
```diff
@@ -106,17 +105,16 @@ there may not be any new interesting information between steps, and `train_inter
...
+replay buffer, and using this mini batch to update the models. Typically, this should be greater than 1.
 However, to imitate the training procedure in certain papers (e.g.
 [Kostrikov et. al](http://arxiv.org/abs/1809.02925), [Blondé et. al](http://arxiv.org/abs/1809.02064)),
 we may want to update N times with different mini batches before grabbing additional samples.
-We can change `train_interval` and `num_update` to N to accomplish this.
+We can change `steps_per_update` to lower than 1 to accomplish this.
 
-Typical Range: `1`
+Typical Range: `10` - `20`
```
andrewcoh, line 117: What's the reasoning behind this range?  It's not atypical to perform an update on every step of an agent.  Granted, that doesn't scale if there are ten agents and every single one takes a step per timestep in which case the minimum is 10.  Maybe this should be written as a function of the number of agents/envs?

ervteng: This was to keep in-line with our previous SAC behavior, which was one update per env step (~10 agent steps). From empirical trials I haven't noticed much of a sample efficiency improvement pushing this down to 1 - in fact it caused a huge wallclock time degradation. This was even with Snoopy Pop, the slowest env we have. 

This number should be independent of # of agents/envs unlike `num_updates`. Maybe we should lower the range to `1-20`?

andrewcoh:
How did you test the GhostTrainer?

ervteng:
> How did you test the GhostTrainer?

Ran it for half an hour and the ELO went up a bit. Am running a full cloud run of all envs now.

ml-agents/mlagents/trainers/sac/trainer.py
```diff
@@ -217,12 +224,11 @@ def _is_ready_update(self) -> bool:
     @timed
     def _update_policy(self) -> None:
         """
-        If train_interval is met, update the SAC policy given the current reward signals.
-        If reward_signal_train_interval is met, update the reward signals from the buffer.
+        Update the SAC policy and reward signals until the steps_per_update ratio
```
andrewcoh, line 227: Would "until the steps_per_update ratio is met" be more appropriate as a docstring for the _update_sac_policy function?

ml-agents/mlagents/trainers/sac/trainer.py
```diff
@@ -265,9 +271,8 @@ def update_sac_policy(self) -> None:
             int(self.trainer_parameters["batch_size"] / self.policy.sequence_length), 1
         )
 
-        num_updates = self.trainer_parameters["num_update"]
         batch_update_stats: Dict[str, list] = defaultdict(list)
-        for _ in range(num_updates):
+        while self.step / self.update_steps > self.steps_per_update:
```
andrewcoh, line 275: Is the new hyperparameter enforced as "every time a steps_per_update num of steps elapses, we perform an update" or "after x steps elapses, we perform x/steps_per_update updates."  Judging from this line, is it the latter?

ervteng: Aren't these things the same thing?

ml-agents/mlagents/trainers/agent_processor.py
```diff
@@ -231,20 +232,23 @@ def __init__(self, behavior_id: str, maxlen: int = 1000):
...
         self.behavior_id = behavior_id
 
     def empty(self) -> bool:
-        return len(self.queue) == 0
+        return self.queue.empty()
 
-    def get_nowait(self) -> T:
+    def get(self, block: bool = True, timeout: float = None) -> T:
+        """
+        Blocking get
```
harperj, line 243: This could use a little elaboration, I think.

ml-agents/mlagents/trainers/sac/trainer.py
```diff
@@ -84,15 +84,22 @@ def __init__(
...
+        # Don't count buffer_init_steps in steps_per_update ratio, but also don't divide-by-0
+        self.update_steps = max(1, self.trainer_parameters["buffer_init_steps"])
+        self.reward_signal_update_steps = max(
+            1, self.trainer_parameters["buffer_init_steps"]
+        )
+
+        self.steps_per_update = (
+            trainer_parameters["steps_per_update"]
+            if "steps_per_update" in trainer_parameters
             else 1
```
harperj, line 97: Noting your docs above about recommended ranges, would something in that range be a better default?  Even better, this could be put into a constant somewhere.

ml-agents/mlagents/trainers/trainer/rl_trainer.py
```diff
@@ -131,9 +131,12 @@ def _maybe_write_summary(self, step_after_process: int) -> None:
         if step_after_process >= self.next_summary_step and self.get_step != 0:
             self._write_summary(self.next_summary_step)
 
-    def advance(self) -> None:
+    def advance(self, empty_queue: bool = False) -> None:
         """
         Steps the trainer, taking in trajectories and updates if ready.
+        Will block and wait briefly if there are no trajectories.
+        :param empty_queue: Whether or not to empty the queue when called. For synchronous
```
harperj, line 138: Why not just always empty the queue when called?

ervteng: It's possible (e.g. in the case of very fast environments, or with enough parallel envs) that the queue fills faster than we can load it into the replay buffer. In that scenario the policy will never update since it'll be stuck in this loop. The opposite won't happen since the put() on the env_manager side is blocking, so if the queue is full the envs will stop while the trainers process the trajectories. 

We could possibly put the loop on the outside (i.e. include the policy update in the get loop). I'd still need some sort of option to disable the blocking behavior in the unthreaded case, though.

harperj: @ervteng Makes sense. I did notice there was a limit on how many times you'll call `get` so I thought that issue might be taken care of in that way.

ervteng: Actually I think you're right - it would be. But it would severely rate-limit the number of updates in that case, e.g. if the queue maxlen is 1000, and its always full, we'd update at most with 1/1000 trajectories. Perhaps this issue isn't a huge deal, especially if the max queue length is small.

harperj: I was just keeping an eye on potentially minimizing changes to our existing interfaces if possible.  Up to you on whether you want to worry about that.  The implementation as it is looks good for me.

docs/Migrating.md
```diff
@@ -12,22 +12,24 @@ The versions can be found in
...
+* The `play_against_current_self_ratio` self-play trainer hyperparameter has been renamed to `play_against_latest_model_ratio`
 * The Jupyter notebooks have been removed from the repository.
 * `Academy.FloatProperties` was removed.
 * `Academy.RegisterSideChannel` and `Academy.UnregisterSideChannel` were removed.
+* `num_updates` and `train_interval` for SAC have been replaced with `steps_per_update`.
+
 
 ### Steps to Migrate
+* Replace the `--load` flag with `--resume` when calling `mlagents-learn`, and don't use the `--train` flag as training
+ will happen by default. To run without training, use `--inference`.
```
mmattar, line 26: To run without training, use `--inference`. --> To run with inference instead of training, use `--inference`.

ml-agents/mlagents/trainers/trainer/rl_trainer.py
```diff
@@ -142,16 +146,18 @@ def advance(self) -> None:
                 # being emptied, the trajectories in the queue are on-policy.
                 for _ in range(traj_queue.maxlen):
```
harperj, line 147: Could this be `traj_queue.qsize`?

ml-agents/mlagents/trainers/trainer/rl_trainer.py
```diff
@@ -142,16 +146,18 @@ def advance(self) -> None:
...
                     try:
-                        t = traj_queue.get_nowait()
+                        t = traj_queue.get(block=not empty_queue, timeout=0.05)
                         self._process_trajectory(t)
```
harperj, line 150: should we still process the trajectory if we have exceeded our buffer size?

ml-agents/mlagents/trainers/trainer/rl_trainer.py
ervteng: 👍 changed.

ml-agents/mlagents/trainers/trainer/rl_trainer.py
ervteng: That's a possibility - I guess the question is whether it is better to dump trajectories or exceed the buffer size. I'm leaning towards exceeding the buffer size. @andrewcoh do you have any preferences?

Ultimately we should change this loop so that we stop getting from the queue if we need to update.

ml-agents/mlagents/trainers/trainer/trainer.py
```diff
@@ -157,12 +157,14 @@ def get_policy(self, name_behavior_id: str) -> TFPolicy:
...
 
     @abc.abstractmethod
-    def advance(self) -> None:
+    def advance(self, empty_queue: bool = False) -> None:
         """
         Advances the trainer. Typically, this means grabbing trajectories
         from all subscribed trajectory queues (self.trajectory_queues), and updating
         a policy using the steps in them, and if needed pushing a new policy onto the right
         policy queues (self.policy_queues).
+        :param empty_queue: Whether or not to empty the queue when called. For synchronous
```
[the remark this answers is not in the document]
ervteng: Removed empty_queue and am using `qsize` to empty the queue every time.

ml-agents/mlagents/trainers/sac/trainer.py
harperj: @andrewcoh @ervteng did you work this out?

docs/Training-PPO.md
```diff
@@ -300,6 +300,16 @@ This option is provided in case you want to initialize different behaviors from
 in most cases, it is sufficient to use the `--initialize-from` CLI parameter to initialize
 all models from the same run.
 
+### (Optional) Advanced: Disable Threading
+
+By default, PPO model updates can happen while the environment is being stepped. To disable this
+behavior, for instance to maintain strict
```
andrewcoh, line 306: The point of this sentence could be a bit clearer. "To disable this, do this." Then, "One may want to do this to maintain the strict on-policyness of PPO..." etc.

docs/Training-SAC.md
```diff
@@ -106,17 +105,18 @@ there may not be any new interesting information between steps, and `train_inter
...
-training event. In SAC, a single "update" corresponds to grabbing a batch of size `batch_size` from the experience
-replay buffer, and using this mini batch to update the models. Typically, this can be left at 1.
-However, to imitate the training procedure in certain papers (e.g.
-[Kostrikov et. al](http://arxiv.org/abs/1809.02925), [Blondé et. al](http://arxiv.org/abs/1809.02064)),
-we may want to update N times with different mini batches before grabbing additional samples.
-We can change `train_interval` and `num_update` to N to accomplish this.
+`steps_per_update` corresponds to the number of agent steps (actions) taken for each mini-batch sampled and used during training. In SAC, a single "update" corresponds to grabbing a batch of size `batch_size` from the experience
+replay buffer, and using this mini batch to update the models. Typically, this should be greater
+than 1. Note that setting `steps_per_update` lower will improve sample efficiency (reduce the number of steps required to train)
+but increase the CPU time spent performaing updates. For most environments where steps are fairly fast (e.g. our example
```
andrewcoh, line 113: performaing -> performing

docs/Training-SAC.md
```diff
@@ -106,17 +105,18 @@ there may not be any new interesting information between steps, and `train_inter
...
+environments) `steps_per_update` equals the number of agents in the scene is a good balance. For slow environments (steps
```
andrewcoh, line 114:
```suggestion
environments) `steps_per_update` equal to the number of agents in the scene is a good balance. For slow environments (steps
```

docs/Training-SAC.md
andrewcoh, line 113:
```suggestion
but increase the CPU time spent performing updates. For most environments where steps are fairly fast (e.g. our example
```

docs/Training-PPO.md
ervteng: Updated the docs - hopefully the new one is clearer

ml-agents/mlagents/trainers/sac/trainer.py
andrewcoh: @ervteng I don't think they are the same thing. If (1) is enforced then (2) is true, but not necessarily is the inverse true. Not sure it's a big deal.

Issue #3531, opened 2020-02-28
Multiagent Bug in Gym-Unity 14.1


sahandrez:
Hello, 

There seems to be a bug with multiagent environments in gym-unity release 14.1. 
Some of the example environments work fine (e.g. 3D ball) while some of them (e.g. WallJump) fail with this error:
```
"The environment does not have the expected amount of agents. "
gym_unity.envs.UnityGymException: The environment does not have the expected amount of agents. Some agents did not request decisions at the same time.
```
It is possible that the issue is originated in the Unity environment side, but I can't seem to find the reason behind it. Single Agent environments work fine. 

I'm using the provided notebooks [here](https://github.com/Unity-Technologies/ml-agents/blob/latest_release/notebooks/getting-started-gym.ipynb).

Thanks, 
Sahand

ervteng:
Hi @sahandrez, the multi-agent gym currently doesn't support environments with multiple types of behaviors (WallJump has two), or on-demand decisions (e.g. Bouncer). This is due to how gym expects all data per step to be from the same observation/action space.

sahandrez:
Hi @ervteng, 
I see. Thanks. I'm having another problem after fixing what you mentioned. Now in our custom environment, we are getting this error: 
```
File "\workspace\ml-agents\gym-unity\gym_unity\envs\__init__.py", line 584, in mark_agent_done
    gym_index = self._agent_id_to_gym_index.pop(agent_id)
KeyError: 7
```
Any idea what might have caused it? Is is caused possibly because some agents are not sending done signals? Thanks.

ervteng:
Hi @sahandrez, are the agents done at different times/do they use on demand decisions? 

Could you post the Inspector for your agents? Would be really helpful for us to debug. Thanks!

sahandrez:
Hi @ervteng, 
Thanks. We solved our problem. There was a bug in our environment which was not flagged in V0.13 but with the new multiagent system in V0.14 we figured it out.

batu:
Hello @sahandrez ,
Can you tell me how you fixed the problem? I am facing the same KeyError issue in `mark_agent_done`

sahandrez:
Hi @batu, 
Sorry I no longer have access to the code. You can ask [@ReinforcedMan](https://github.com/ReinforcedMan) for more info.

RealZST:
Hi @batu ,
[...]

PR #3700, opened 2020-03-26
handle multiple dones in a single step

chriselion:
### Proposed change(s)

Fix two bugs in the gym interface that could happen if an Agent called EndEpisode() multiple times in the same step:
1) The bookkeeping would try to remove the "transient" agent before it was added, which would fail. We also need to avoid adding it since it's already done.
2) For a single Agent, the checks in `_sanitize_info()` was too strict - we could have the case where `step_result.n_agents()=3 self._n_agents=1 n_extra_agents=2` because there we 2 "done" agents in the step.

This would show up with stack traces:
```
  File "test_gym.py", line 38, in <module>
    observation, reward, done, info = multi_env.step(multi_env.action_space.sample())
  File "/Users/chris.elion/code/ml-agents/gym-unity/gym_unity/envs/__init__.py", line 222, in step
    step_result = self._step()
  File "/Users/chris.elion/code/ml-agents/gym-unity/gym_unity/envs/__init__.py", line 449, in _step
    return self._sanitize_info(info)
  File "/Users/chris.elion/code/ml-agents/gym-unity/gym_unity/envs/__init__.py", line 371, in _sanitize_info
    "The number of agents in the scene does not match the expected number."
gym_unity.envs.UnityGymException: The number of agents in the scene does not match the expected number.
```
with multiagent=False, or 
```
  File "test_gym.py", line 32, in <module>
    observations, rewards, dones, info = multi_env.step(actions)
  File "/Users/chris.elion/code/ml-agents/gym-unity/gym_unity/envs/__init__.py", line 222, in step
    step_result = self._step()
  File "/Users/chris.elion/code/ml-agents/gym-unity/gym_unity/envs/__init__.py", line 445, in _step
    self.agent_mapper.mark_agent_done(agent_id, reward)
  File "/Users/chris.elion/code/ml-agents/gym-unity/gym_unity/envs/__init__.py", line 531, in mark_agent_done
    gym_index = self._agent_id_to_gym_index.pop(agent_id)
KeyError: 2
```
with multiagent=True

https://forum.unity.com/threads/errors-when-using-gym-unity-0-14-1-and-0-15-0.852586/

https://github.com/Unity-Technologies/ml-agents/issues/3531#issuecomment-592663967

https://jira.unity3d.com/browse/MLA-805

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/gym-unity/gym_unity/envs/__init__.py
+++ b/gym-unity/gym_unity/envs/__init__.py
@@ -364,9 +364,8 @@ class UnityEnv(gym.Env):
 
     def _sanitize_info(self, step_result: BatchedStepResult) -> BatchedStepResult:
         n_extra_agents = step_result.n_agents() - self._n_agents
-        if n_extra_agents < 0 or n_extra_agents > self._n_agents:
+        if n_extra_agents < 0:
             # In this case, some Agents did not request a decision when expected
-            # or too many requested a decision
             raise UnityGymException(
                 "The number of agents in the scene does not match the expected number."
             )
@@ -386,6 +385,10 @@ class UnityEnv(gym.Env):
         # only cares about the ordering.
         for index, agent_id in enumerate(step_result.agent_id):
             if not self._previous_step_result.contains_agent(agent_id):
+                if step_result.done[index]:
+                    # If the Agent is already done (e.g. it ended its epsiode twice in one step)
+                    # Don't try to register it here.
+                    continue
                 # Register this agent, and get the reward of the previous agent that
                 # was in its index, so that we can return it to the gym.
                 last_reward = self.agent_mapper.register_new_agent_id(agent_id)
@@ -528,8 +531,12 @@ class AgentIdIndexMapper:
         """
         Declare the agent done with the corresponding final reward.
         """
-        gym_index = self._agent_id_to_gym_index.pop(agent_id)
-        self._done_agents_index_to_last_reward[gym_index] = reward
+        if agent_id in self._agent_id_to_gym_index:
+            gym_index = self._agent_id_to_gym_index.pop(agent_id)
+            self._done_agents_index_to_last_reward[gym_index] = reward
+        else:
+            # Agent was never registered in the first place (e.g. EndEpisode called multiple times)
+            pass
 
     def register_new_agent_id(self, agent_id: int) -> float:
         """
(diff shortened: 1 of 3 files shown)


gym-unity/gym_unity/envs/__init__.py
```diff
@@ -528,8 +531,12 @@ def mark_agent_done(self, agent_id: int, reward: float) -> None:
         """
         Declare the agent done with the corresponding final reward.
         """
-        gym_index = self._agent_id_to_gym_index.pop(agent_id)
-        self._done_agents_index_to_last_reward[gym_index] = reward
+        if agent_id in self._agent_id_to_gym_index:
+            gym_index = self._agent_id_to_gym_index.pop(agent_id)
```
chriselion: This is where the reported exceptions are. We could also make a check before the call to `mark_agent_done`, but it seemed a bit cleaner to do it here.

gym-unity/gym_unity/envs/__init__.py
```diff
@@ -386,6 +385,10 @@ def _sanitize_info(self, step_result: BatchedStepResult) -> BatchedStepResult:
         # only cares about the ordering.
         for index, agent_id in enumerate(step_result.agent_id):
             if not self._previous_step_result.contains_agent(agent_id):
+                if step_result.done[index]:
```
chriselion: Just making the mark_agent_done() change without this means that we'll register the agent and run out of "slots" later.

gym-unity/gym_unity/envs/__init__.py
```diff
@@ -364,9 +364,8 @@ def _check_agents(self, n_agents: int) -> None:
 
     def _sanitize_info(self, step_result: BatchedStepResult) -> BatchedStepResult:
         n_extra_agents = step_result.n_agents() - self._n_agents
-        if n_extra_agents < 0 or n_extra_agents > self._n_agents:
+        if n_extra_agents < 0:
```
chriselion: I'm not 100% it's safe to remove this, but it's what was triggering in the single agent case (with 2 "done" agents)

chriselion:
Bypassing C# yamato tests.

PR #3705, opened 2020-03-28
[refactor] CLI changes

ervteng:
### Proposed change(s)

- Make `--train` the default, and deprecate it (with warning message)
- Add `--inference` flag for inference mode
- Replace `--load` with `--resume`
- Block training if the model files already exist. Require `--force` to overwrite these files.
- Tensorboard writer deletes old summaries if overwriting. This prevents the double-back-around plots we sometimes get if we use the same run-id.

~TODO: Doc changes, changelog~

MLA-787

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


ml-agents/mlagents/trainers/trainer_util.py
```diff
@@ -190,3 +190,33 @@ def _load_config(fp: TextIO) -> Dict[str, Any]:
             "Error parsing yaml file. Please check for formatting errors. "
             "A tool such as http://www.yamllint.com/ can be helpful with this."
         ) from e
+
+
+def handle_existing_directories(
+    model_path: str, summary_path: str, resume: bool, force: bool
```
ervteng: note that `summary_path` isn't used yet but with consolidating all summaries into the same directory will be checked as well.

ml-agents/mlagents/trainers/stats.py
```diff
@@ -198,8 +198,21 @@ def _maybe_create_summary_writer(self, category: str) -> None:
                 basedir=self.base_dir, category=category
             )
             os.makedirs(filewriter_dir, exist_ok=True)
+            self._delete_all_events_files(filewriter_dir)
```
harperj: Wouldn't this be a problem when using `--resume`?

ervteng: Added a fix - the TB writer now takes in a param that tells it to not delete if `--resume` is passed.

ml-agents/mlagents/trainers/stats.py
```diff
@@ -199,8 +202,22 @@ def _maybe_create_summary_writer(self, category: str) -> None:
                 basedir=self.base_dir, category=category
             )
             os.makedirs(filewriter_dir, exist_ok=True)
+            if self._clear_past_data:
+                self._delete_all_events_files(filewriter_dir)
             self.summary_writers[category] = tf.summary.FileWriter(filewriter_dir)
 
+    def _delete_all_events_files(self, directory_name: str) -> None:
```
harperj: Minor: could add some logging to make the user aware this is happening.

ervteng: Added

PR #3716, opened 2020-04-01
Make the DecisionRequester public and customizable.

surfnerd:
### Proposed change(s)
Make the DecisionRequest public and add a delegate to its API to allow for logic customization in calling `RequestDecision` and/or `RequestAction`.
I chose to use a delegate instead of a virtual method to make it easier for users to implement Agent subclasses.  With the delegate, they can keep all of their Agent logic in one class, and easily hook into the EnvironmentStep loop in order to ensure that their calls to RequestAction or RequestDecision are handled in the appropriate step.

This pull request also changes how we handle logging our standalone build.  We now collect the standalone build log as an artifact and will cat the log if the build fails or if the verbose flag was set to `true`.

[MLA-764](https://jira.unity3d.com/browse/MLA-764)

### Types of change(s)
- [x] Code refactor
~- [ ] Bug fix~
~- [ ] New feature~
~- [ ] Breaking change~
~- [ ] Documentation update~
~- [ ] Other (please describe)~

### Checklist


com.unity.ml-agents/Runtime/DecisionRequester.cs
```diff
@@ -19,6 +20,26 @@ internal class DecisionRequester : MonoBehaviour
                  "of 5 means that the Agent will request a decision every 5 Academy steps.")]
         public int DecisionPeriod = 5;
 
+        /// <summary>
+        /// A delegate that can be used to customize the logic of when your Agent gets its
+        /// <see cref="Agent.RequestDecision"/> and <see cref="Agent.RequestAction"/> called.
+        /// If this value is not get, the default logic will use <see cref="DecisionPeriod"/> and
```
chriselion, line 26:
```suggestion
        /// If this value is not set, the default logic will use <see cref="DecisionPeriod"/> and
```

surfnerd: ah, I will move that sentence down to the actual member as well.  Thanks.

com.unity.ml-agents/Runtime/DecisionRequester.cs
```diff
@@ -29,37 +29,31 @@ internal class DecisionRequester : MonoBehaviour
...
-
         Agent m_Agent;
-        int m_Offset;
 
         internal void Awake()
         {
-            m_Offset = offsetStep ? gameObject.GetInstanceID() : 0;
             m_Agent = gameObject.GetComponent<Agent>();
-            Academy.Instance.AgentSetStatus += MakeRequests;
+            Debug.Assert(m_Agent != null, "Agent component was not found on this gameObject and is required.");
```
chriselion, line 37: Should was also (or instead) add `[RequireComponent(typeof(Agent))]` to the class?

chriselion:
Looks good, just needs a test.

com.unity.ml-agents/Runtime/DecisionRequester.cs
surfnerd: I briefly thought about doing exactly that and decided not to.  I'll add it.

ml-agents/tests/yamato/yamato_utils.py
```diff
@@ -95,6 +104,8 @@ def checkout_csharp_version(csharp_version):
     if csharp_version is None:
         return
     csharp_dirs = ["com.unity.ml-agents", "Project"]
+    subprocess.check_call(f"rm -rf Project/Library", shell=True)
+    subprocess.check_call(f"rm -rf com.unity.ml-agents", shell=True)
```
chriselion, line 108: nit: move this into the loop below

PR #3718, opened 2020-04-02
Rename monster cube and change fire to red

ervteng:
### Proposed change(s)

Rename monster cube and change fire to red.

dragongif


awjuliani:
🙌

PR #3722, opened 2020-04-02
Fix save snapshot bug in ghost trainer

andrewcoh:
### Proposed change(s)

This PR fixes the bug that the ghost trainer did not load the latest snapshot before saving a checkpoint and then reload the current snapshot to resume training.

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/mlagents/trainers/ghost/trainer.py
+++ b/ml-agents/mlagents/trainers/ghost/trainer.py
@@ -301,13 +301,22 @@ class GhostTrainer(Trainer):
     def save_model(self, name_behavior_id: str) -> None:
         """
         Forwarding call to wrapped trainers save_model
+        Loads the latest snapshot then reloads current snapshot.
         """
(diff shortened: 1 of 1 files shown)


ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -301,13 +301,22 @@ def end_episode(self):
     def save_model(self, name_behavior_id: str) -> None:
         """
         Forwarding call to wrapped trainers save_model
+        Loads the latest snapshot then reloads current snapshot.
```
ervteng, line 304: Can we make the comment "Loads the latest policy weights, saves it, then reloads the current policy weights before resuming training." Just so we don't forget exactly what this code does 3 months from now

PR #3724, opened 2020-04-02
Add CompletedEpisodes counter

chriselion:
### Proposed change(s)

Adds `Agent.CompletedEpisodes` property, and uses this in the ModelOverrider.

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ModelOverrider.cs
+++ b/Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ModelOverrider.cs
@@ -86,7 +86,11 @@ namespace MLAgentsExamples
         {
             if (m_MaxEpisodes > 0)
             {
-                if (m_NumSteps > m_MaxEpisodes * m_Agent.maxStep)
+                // For Agents without maxSteps, exit as soon as we've hit the target number of episodes.
+                // For Agents that specify maxStep, also make sure we've gone at least that many steps.
+                // Since we exit as soon as *any* Agent hits its target, the maxSteps condition keeps us running
+                // a bit longer in case there's an early failure.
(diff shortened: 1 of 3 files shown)


Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ModelOverrider.cs
```diff
@@ -86,7 +86,11 @@ void FixedUpdate()
         {
             if (m_MaxEpisodes > 0)
             {
-                if (m_NumSteps > m_MaxEpisodes * m_Agent.maxStep)
+                // For Agents without maxSteps, exit as soon as we've hit the target number of episodes.
+                // For Agents that specify maxStep, also make sure we've gone at least that many steps.
+                // Since we exit as soon as *any* Agent hits its target, the maxSteps condition keeps us running
+                // a bit longer in case there's an early failure.
+                if (m_Agent.CompletedEpisodes >= m_MaxEpisodes && m_NumSteps > m_MaxEpisodes * m_Agent.maxStep)
```
surfnerd, line 93: should this only be enabled with some sort of define around it?  It seems a bit dubious without one.

surfnerd: like if we're only doing this on cloud build we could surround it with `UNITY_CLOUD_BUILD`.  Or have a custom build define for it on the cloudbuild configurations.

chriselion: It's in the Examples directory, with a note that says "This is used internally to validate the file after training is done." above. I don't think we need to make it any safer.

PR #3726, opened 2020-04-03
Wrapped trainer has internal policy in ghost trainer

andrewcoh:
### Proposed change(s)

Fixing potential bug in symmetric games introduced by team change.  The fix creates an internal policy to be used by the wrapped trainer for training so the wrapped trainer always has the latest model.  This simplifies the save/export functions in the ghost trainer. 

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/mlagents/trainers/ppo/trainer.py
+++ b/ml-agents/mlagents/trainers/ppo/trainer.py
@@ -70,7 +70,7 @@ class PPOTrainer(RLTrainer):
         self._check_param_keys()
         self.load = load
         self.seed = seed
-        self.policy: NNPolicy = None  # type: ignore
+        self.policy: TFPolicy = None  # type: ignore
 
     def _check_param_keys(self):
         super()._check_param_keys()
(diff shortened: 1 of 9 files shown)


ml-agents/mlagents/trainers/ppo/trainer.py
```diff
@@ -253,14 +253,12 @@ def add_policy(
                     self.__class__.__name__
                 )
             )
-        if not isinstance(policy, NNPolicy):
-            raise RuntimeError("Non-NNPolicy passed to PPOTrainer.add_policy()")
-        self.policy = policy
+        self.policy = self.create_policy(brain_parameters)
         self.optimizer = PPOOptimizer(self.policy, self.trainer_parameters)
```
ervteng, line 257: Almost want to say the add_policy method should go in `rl_trainer` with the Optimizer as an object (e.g. `self.optimizer_class = PPOOptimizer` and then in the method `self.optimizer = self.optimizer_class(self.policy)`. Since it's duplicated here and in SAC.

andrewcoh:
Closing this for another solution. Please see [#3735 ](https://github.com/Unity-Technologies/ml-agents/pull/3735)

PR #3745, opened 2020-04-06
check demonstration version before loading

chriselion:
### Proposed change(s)
Currently the demonstration version is always set to 1:
https://github.com/Unity-Technologies/ml-agents/blob/d4810e28d16a13efb580337eab40db7b99cd9a8c/com.unity.ml-agents/Runtime/Demonstrations/Demonstration.cs#L36

If we decide to change the version later, we can either make the loading code handle both versions, or drop support for version 1 (hopefully the former). In the meantime, we should make sure that any attempt to load a new version in old code is unsupported.

https://jira.unity3d.com/browse/MLA-840

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/mlagents/trainers/demo_loader.py
+++ b/ml-agents/mlagents/trainers/demo_loader.py
@@ -23,7 +23,7 @@ from google.protobuf.internal.encoder import _EncodeVarint  # type: ignore
 
 
 INITIAL_POS = 33
-SUPPORTED_DEMONSTRATION_VERSIONS = frozenset([1])
+SUPPORTED_DEMONSTRATION_VERSIONS = frozenset([0, 1])
 
 
 @timed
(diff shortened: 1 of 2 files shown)


mmattar:
> or drop support for version 1

If we ship v1.0 of the SDK with the Demo version set to 1, doesn't this mean that we can't drop support for it? Otherwise, that would be a major breaking change.

chriselion:
Correct. But I don't think support for it would be a major burden, we'd just need to check the version and switch which loading code we call.

ml-agents/mlagents/trainers/demo_loader.py
```diff
@@ -22,6 +22,10 @@
 from google.protobuf.internal.encoder import _EncodeVarint  # type: ignore
 
 
+INITIAL_POS = 33
```
ervteng, line 25: Does this mean we should fix INITIAL_POS for all future versions?

chriselion: Depends - I think the current meta proto should stay as it is. But if we wanted to add more data in a v2 format, we could have a v2 metadata that's in file after the v1 metadata.

PR #3749, opened 2020-04-07
Add ModelOverrider to all of the Agent prefabs to enable Barracuda Inference with specified .nn model file

xiaomaogy:
### Proposed change(s)

Added the ModelOverrider.cs to all of the Agent prefabs, so that Barracuda inference with a specified .nn model file loaded is enabled. 

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ProjectSettingsOverrides.cs
```diff
@@ -43,6 +43,8 @@ public void Awake()
             Physics.defaultSolverIterations = solverIterations;
             Physics.defaultSolverVelocityIterations = solverVelocityIterations;
 
+            // Make sure the Academy singleton is initialized first, since it will create the SideChannels.
+            var academy = Academy.Instance;
```
anupambhatnagar, line 47: are we still using Academy? I thought it was deprecated.

xiaomaogy: I'm not super sure about this. It comes from an earlier commit from @chriselion.

chriselion: Academy was changed from a MonoBehaviour to a singleton, but it is not deprecated and will not be any time soon.

The line below (SideChannelUtils.GetSideChannel) tries to access a SideChannel that is created when the Academy is initialized. Since the Academy initialization is lazy, it's hard to guarantee that it has been done at this point (it depends on whether Agents are initialized before or after this). Thus, this line ensures that the Academy is initialized if it wasn't already, and has no effect if it already is.

Feel free to move as much as you think is relevant into the comment.

Project/ProjectSettings/EditorBuildSettings.asset
```diff
@@ -4,5 +4,8 @@
 EditorBuildSettings:
   m_ObjectHideFlags: 0
   serializedVersion: 2
-  m_Scenes: []
+  m_Scenes:
```
chriselion, line 7: Please revert the change to this file.

xiaomaogy: Thanks for the reminder, reverted.

PR #3751, opened 2020-04-07
[MLA-825] Add default values for SideChannel IncomingMessages methods

chriselion:
### Proposed change(s)

To assist with backwards compatibility across side channels, this adds a default value when reading from IncomingMessages. This will let us add data to the messages in the future, but still allow reading from older versions.

Note that this only falls back to the default if it's at the end of the buffer. Trying to read an int when there's only 1 byte left will still raise/throw, but I think this is the correct behavior in that case.

https://jira.unity3d.com/browse/MLA-825

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/com.unity.ml-agents/Runtime/SideChannels/IncomingMessage.cs
+++ b/com.unity.ml-agents/Runtime/SideChannels/IncomingMessage.cs
@@ -1,5 +1,5 @@
 using System.Collections.Generic;
-using System.Runtime.CompilerServices
+using System.Runtime.CompilerServices;
 using System;
 using System.IO;
 using System.Text;
--- a/com.unity.ml-agents/Runtime/SideChannels/IncomingMessage.cs
+++ b/com.unity.ml-agents/Runtime/SideChannels/IncomingMessage.cs
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Runtime.CompilerServices
 using System;
 using System.IO;
 using System.Text;
--- a/com.unity.ml-agents/Runtime/SideChannels/IncomingMessage.cs
+++ b/com.unity.ml-agents/Runtime/SideChannels/IncomingMessage.cs
@@ -26,38 +26,43 @@ namespace MLAgents.SideChannels
         }
 
         /// <summary>
-        /// Read a boolan value from the message.
+        /// Read a boolean value from the message.
         /// </summary>
         /// <returns></returns>
-        public bool ReadBoolean()
+        public bool ReadBoolean(bool defaultValue = false)
         {
-            return m_Reader.ReadBoolean();
+            return CanReadMore() ? m_Reader.ReadBoolean() : defaultValue;
         }
 
         /// <summary>
         /// Read an integer value from the message.
         /// </summary>
         /// <returns></returns>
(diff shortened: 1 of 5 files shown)


com.unity.ml-agents/Runtime/SideChannels/IncomingMessage.cs
```diff
@@ -26,38 +26,47 @@ public IncomingMessage(byte[] data)
...
+            return CanReadMore() ? m_Reader.ReadSingle() : defaultValue;
         }
 
         /// <summary>
         /// Read a string value from the message.
         /// </summary>
+        /// <param name="defaultValue">Default value to use if the end of the message is reached.</param>
         /// <returns></returns>
-        public string ReadString()
+        public string ReadString(string defaultValue = default)
```
chriselion, line 63: Should we default to `""` instead?

com.unity.ml-agents/Runtime/SideChannels/IncomingMessage.cs
```diff
@@ -66,9 +75,15 @@ public string ReadString()
         /// <summary>
         /// Reads a list of floats from the message. The length of the list is stored in the message.
         /// </summary>
+        /// <param name="defaultValue">Default value to use if the end of the message is reached.</param>
         /// <returns></returns>
-        public IList<float> ReadFloatList()
+        public IList<float> ReadFloatList(IList<float> defaultValue = default)
```
chriselion, line 80: Not sure if there's a better default here.

com.unity.ml-agents/Runtime/SideChannels/IncomingMessage.cs
```diff
@@ -97,5 +112,14 @@ public void Dispose()
             m_Reader?.Dispose();
             m_Stream?.Dispose();
         }
+
+        /// <summary>
+        /// Whether or not there is more data left in the stream that can be read.
+        /// </summary>
+        /// <returns></returns>
+        bool CanReadMore()
```
surfnerd, line 120: could this be inlined?

surfnerd: [ [MethodImpl(MethodImplOptions.AggressiveInlining)]](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.methodimplattribute.-ctor?view=netframework-4.8#System_Runtime_CompilerServices_MethodImplAttribute__ctor)
```suggestion
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        bool CanReadMore()
```

com.unity.ml-agents/Runtime/SideChannels/IncomingMessage.cs
```diff
@@ -1,4 +1,5 @@
 using System.Collections.Generic;
+using System.Runtime.CompilerServices
```
surfnerd, line 2:
```suggestion
using System.Runtime.CompilerServices;
```

PR #3765, opened 2020-04-10
Change Agent.Heuristic to take a float[] argument

chriselion:
### Proposed change(s)

Changes the signature of Agent.Heuristic() to take a `float[]` instead of returning one. This eliminate a common source of user error, where the returned array was the wrong size.

### Useful links (Github issues, JIRA tickets, ML-Agents forum threads etc.)

https://jira.unity3d.com/browse/MLA-768
https://github.com/Unity-Technologies/ml-agents/issues/3719

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [x] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist
- [ ] Added tests that prove my fix is effective or that my feature works
- [x] Updated the [changelog](https://github.com/Unity-Technologies/ml-agents/blob/master/com.unity.ml-agents/CHANGELOG.md) (if applicable)
- [x] Updated the [documentation](https://github.com/Unity-Technologies/ml-agents/tree/master/docs) (if applicable)
- [x] Updated the [migration guide](https://github.com/Unity-Technologies/ml-agents/blob/master/docs/Migrating.md) (if applicable)

PR #3773, opened 2020-04-14
Improvements to Learning-Environment-Create-New.md

mmattar:
- Changed the ordered list to use "1."
- Trimmed down text
- Removed reference to materials as those are in the Example Envs project

### Proposed change(s)

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [X] Documentation update
- [ ] Other (please describe)

### Checklist


docs/Learning-Environment-Create-New.md
```diff
@@ -105,31 +85,36 @@ component later in the tutorial.
...
+* `CollectObservations(VectorSensor sensor)`
+* `OnActionReceived(float[] vectorAction)`
+
+We overview each of these in more detail in the dedicated subsections below.
 
 ### Initialization and Resetting the Agent
 
-When the Agent reaches its target, its episode ends and the `OnEpisodeBegin()`
-method moves the target to a random location. In addition, if the Agent rolls
+When the Agent (Sphere) reaches its target (Cube), its episode ends and the `OnEpisodeBegin()`
```
chriselion, line 113: This paragraph doesn't actually talk about "Initialization" even though it's in the section title. Maybe mention that `Initialization` is called at the start of simulation too?

docs/Learning-Environment-Create-New.md
```diff
@@ -230,23 +194,17 @@ public override void CollectObservations(VectorSensor sensor)
...
-response to the observation input and then sees what kind of rewards it gets as
-a result.
+receives actions and assigns the reward.
+
+#### Actions
+
+To solve the task of moving towards the target, the Agent (Sphere) needs to be able to
+move in the `x` and `z` directions. As such, we will provide 2 actions to the agent.
+The first determines the force applied along the x-axis; the second
+determines the force applied along the z-axis. (If we allowed the Agent to move in three dimensions, then we would need a third action.
```
chriselion, line 207: nit: line length

docs/Learning-Environment-Create-New.md
```diff
@@ -230,23 +194,17 @@ public override void CollectObservations(VectorSensor sensor)
...
-receives the actions from the Brain and assigns the reward.
-
-### Actions
-
-The decision of the Brain comes in the form of an action array passed to the
-`OnActionReceived()` function. The number of elements in this array is determined by
-the `Vector Action` `Space Type` and `Space Size` settings of the
-agent's Brain. The RollerAgent uses the continuous vector action space and needs
-two continuous control signals from the Brain. Thus, we will set the Brain
-`Space Size` to 2. The first element,`action[0]` determines the force
```
chriselion, line 243: I think we should keep some of the instructions that are removed here, namely setting the action space to continuous and the space size to 2.

chriselion: Never mind, guess it's covered below.

docs/Learning-Environment-Create-New.md
```diff
@@ -339,91 +293,94 @@ to connect everything together in the Unity Editor. This involves
...
+the heuristic will generate an action corresponding to the values of the "Horizontal" and "Vertical"
+input axis (which correspond to the keyboard arrow keys):
 
 ```csharp
-    public override void Heuristic(float[] actionsOut)
-    {
-        actionsOut[0] = Input.GetAxis("Horizontal");
-        actionsOut[1] = Input.GetAxis("Vertical");
-    }
+public override float[] Heuristic()
```
chriselion, line 317: Old code was correct, see https://github.com/Unity-Technologies/ml-agents/pull/3765

docs/Learning-Environment-Create-New.md
```diff
@@ -473,7 +420,7 @@ localPosition rather than position, so that our position reference is in referen
 to the prefab TrainingArea's location, and not global coordinates.
 
 1. Replace all references of `this.transform.position` in RollerAgent.cs with `this.transform.localPosition`.
-2. Replace all references of `Target.position` in RollerAgent.cs with `Target.localPosition`.
+1. Replace all references of `Target.position` in RollerAgent.cs with `Target.localPosition`.
```
chriselion, line 423: Should we just start with these in the example, instead of changing them later? There's been some confusion before because the examples conflict with the best practices that we call out elsewhere.

mmattar: Makes sense.

docs/Learning-Environment-Create-New.md
mmattar: Updated.

docs/Learning-Environment-Create-New.md
mmattar: Fixed

docs/Learning-Environment-Create-New.md
mmattar: Good catch.

docs/Learning-Environment-Create-New.md
```diff
@@ -435,45 +434,22 @@ well the Agent is achieving the task. In this example, the maximum reward an
...
 
-In many of the [example environments](Learning-Environment-Examples.md), many copies of
-the training area are instantiated in the scene. This generally speeds up training,
-allowing the environment to gather many experiences in parallel. This can be achieved
-simply by instantiating many Agents which share the `Behavior Parameters`. Use the following steps to
-parallelize your RollerBall environment.
+In many of the [example environments](Learning-Environment-Examples.md), many
+copies of the training area are instantiated in the scene. This generally speeds
+up training, allowing the environment to gather many experiences in parallel.
+This can be achieved simply by instantiating many Agents which share the
```
chriselion, line 442: > which share the `Behavior Parameters`

This isn't really accurate; they should have identical `Behavior Parameters` but they're not actually "shared" .

mmattar: Good catch - fixed it.

PR #3776, opened 2020-04-14
Improvements to Training-ML-Agents

mmattar:
- Removed duplicate documentation
- Moved CLI descriptions to learn.py
- Reorganized "Training with mlagents-learn" into 5 sub-sections

### Proposed change(s)

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [X] Documentation update
- [ ] Other (please describe)

### Checklist


ml-agents/mlagents/trainers/learn.py
```diff
@@ -123,41 +138,50 @@ def _create_parser():
...
     )
     argparser.add_argument(
         "--base-port",
         default=UnityEnvironment.BASE_ENVIRONMENT_PORT,
         type=int,
-        help="Base port for environment communication",
+        help="The starting port for environment communication. Each concurrent Unity environment "
+        "instance will get assigned a port sequentially, starting from the base-port. Each instance "
+        "will use the port (base_port + worker_id), where the worker_id is sequential IDs given to "
+        "each instance from 0 to (num_envs - 1). Note that When training using the Editor rather "
```
chriselion, line 151:
```suggestion
        "each instance from 0 to (num_envs - 1). Note that when training using the Editor rather "
```

mmattar:
@ervteng - @chriselion has already reviewed/approved but wanted a quick pass from you.

docs/Training-ML-Agents.md
```diff
@@ -1,196 +1,116 @@
...
+
+These artifacts (except the `.nn` file) are updated throughout the training process and finalized
+when training completes or is interrupted.
+
+#### Debugging
 
-**Note:** The default port TensorBoard uses is 6006. If there is an existing session
-running on port 6006 a new session can be launched on an open port using the --port
-option.
+If you enable the `--debug` flag in the command line, the trainer metrics are logged to a CSV file
```
ervteng, line 75: Unfortunately none of these things are correct anymore, I would just remove the whole Debugging section. 

The CSV is still created under the `summaries/` directory, but is slated to be deprecated and replaced with a JSON file, so it's probably OK if we don't mention it.

docs/Training-ML-Agents.md
```diff
@@ -1,196 +1,116 @@
...
+#### Stopping and Resuming Training
 
-To interrupt training and save the current progress, hit Ctrl+C once and wait for the
-model to be saved out.
+To interrupt training and save the current progress, hit `Ctrl+C` once and wait for the
+model(s) to be saved out.
 
-### Loading an Existing Model
+To resume a previously interrupted or completed training run, use the `--resume` flag and
+make sure to specify the previously used `run-id`.
```
ervteng, line 92: Should be consistent when referring to the run ID in the docs? I think we use a mix of `<run-identifier>`, `run-id`, "run-id", and "run ID".  I'm biased towards using `<run-identifier>` when talking about CLI commands and "run ID" or "run identifier" in paragraphs, but OK with the other options as well.

docs/Training-ML-Agents.md
mmattar: So should we remove now and bring back once we add the JSON file?

mmattar: And also, remove the --debug option from the CLI?

docs/Training-ML-Agents.md
mmattar: Good point. I made a quick fix - let me know if this is what you had in mind.

docs/Training-ML-Agents.md
ervteng: +1 for remove now and bring back once we add the JSON file. 

The `--debug` CLI option is still there, but it doesn't really help the user in debugging training (it's mostly for us), nor does it log the CSV. So it probably doesn't go in this doc. Is it OK if we just leave it to the `--help` menu to describe?

docs/Training-ML-Agents.md
ervteng: 👍

PR #3797, opened 2020-04-17
Clean up and fix save and load in ghost

andrewcoh:
### Proposed change(s)

A recent change [#3735 ](https://github.com/Unity-Technologies/ml-agents/pull/3735) created a bug in the save/load feature.  The wrapped trainer was correctly loading the policy but the initial ghosted policies were not receiving the loaded model due to the ordering of the function calls.  Specifically, the wrapped trainer must add_policy so that the optimizer can load the weights of the checkpoint before the ghosted policies can receive those weights.

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/mlagents/trainers/ghost/trainer.py
+++ b/ml-agents/mlagents/trainers/ghost/trainer.py
@@ -326,6 +326,7 @@ class GhostTrainer(Trainer):
         """
         policy = self.trainer.create_policy(parsed_behavior_id, brain_parameters)
         policy.create_tf_graph()
+        policy.initialize_or_load()
         policy.init_load_weights()
         team_id = parsed_behavior_id.team_id
         self.controller.subscribe_team_id(team_id, self)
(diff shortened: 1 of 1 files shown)


ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -335,15 +336,14 @@ def create_policy(
             internal_trainer_policy = self.trainer.create_policy(
                 parsed_behavior_id, brain_parameters
             )
-            internal_trainer_policy.create_tf_graph()
```
andrewcoh, line 338: This was being called twice in both the optimizer and here.

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -335,15 +336,14 @@ def create_policy(
...
+            self.trainer.add_policy(parsed_behavior_id, internal_trainer_policy)
```
andrewcoh, line 339: This is moved so that it occurs before the ghost policy receives its weights.

ml-agents/mlagents/trainers/ghost/trainer.py
```diff
@@ -326,6 +326,7 @@ def create_policy(
         """
         policy = self.trainer.create_policy(parsed_behavior_id, brain_parameters)
         policy.create_tf_graph()
+        policy.initialize_or_load()
```
andrewcoh, line 329: In the case of symmetric games (soccer/tennis), this ensures that both policies are initialized from the checkpoint.

PR #3815, opened 2020-04-21
[refactor] Allow full RunOptions to be specified in trainer configuration YAML

ervteng:
### Proposed change(s)

Rather than have the YAML stored as part of RunOptions, make the YAML a representation of RunOptions, including engine options and other parameters. Anything not specified in the YAML config will be set to the default values; anything specified in the CLI will override values found in the YAML. 

This removes the need for a separate run_experiment entry point. 

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


ml-agents/mlagents/trainers/learn.py
```diff
@@ -43,6 +43,31 @@
 logger = logging_util.get_logger(__name__)
 
 
+class DetectDefault(argparse.Action):
```
ervteng, line 46: I'm not super happy about this method, but couldn't find another way to detect non-default parameters without removing the `default=` from the `add_argument` calls.

harperj: Another option would be to remove the defaults / default to `None`.

ml-agents/mlagents/trainers/learn.py
```diff
@@ -268,26 +308,36 @@ def from_argparse(args: argparse.Namespace) -> "RunOptions":
           configs loaded from files.
         """
         argparse_args = vars(args)
+        full_config = {}
```
harperj, line 311: I think we should be clear on what config we are talking about.  I think a better name here would be `run_options_dict` or something.

ml-agents/mlagents/trainers/learn.py
```diff
@@ -268,26 +308,36 @@ def from_argparse(args: argparse.Namespace) -> "RunOptions":
...
+        full_config.update(argparse_args)
+        # Since argparse accepts file paths in the config options which don't exist in CommandLineOptions,
+        # these keys will need to be deleted to use the **/splat operator below.
+        full_config.pop("trainer_config_path")
```
harperj, line 315: Wonder if the path issue could be fixed upstream of this.  Maybe with a custom action..

ml-agents/mlagents/trainers/learn.py
ervteng: I think the main reason I wanted to keep the defaults was so they would show up when we do `--help`. It'll be especially important once @mmattar removes the command line options documentation. The alternative would be to put the (default=X) text in the help description itself, which isn't terribad either.

harperj: Makes sense.  I didn't think of the self-documenting side of it.  Adding them in the description text would be OK as long as the description text is generated from the actual default values :-)

ml-agents/mlagents/trainers/learn.py
ervteng: Not sure if it's what you had in mind, but I added custom action that removes the `trainer_config_path` attribute from the namespace and adds it to a class variable which can be accessed by `learn.py`. Also moved custom actions and the huge `create_parser` method to a separate file.

ml-agents/mlagents/trainers/learn.py
```diff
@@ -268,26 +86,36 @@ def from_argparse(args: argparse.Namespace) -> "RunOptions":
           configs loaded from files.
         """
         argparse_args = vars(args)
-        config_path = argparse_args["trainer_config_path"]
-        # Load YAML and apply overrides as needed
+        run_options_dict = {}
+        run_options_dict.update(argparse_args)
+        print(args)
```
harperj, line 91: Was this intentional?

ervteng: Good catch - this was for debugging, removed

PR #3817, opened 2020-04-21
[bug-fix] Bugfixes for Threaded Trainers

ervteng:
### Proposed change(s)

This PR fixes two issues with the threaded trainer. 

- When ghost-training, two threads are created for the same trainer causing a race condition on the buffer. We now only create one thread for a ghost-trainer. 
- If there is more than one trainer but no samples are produced by the environment for that trainer (e.g. WallJump's swappable brains) the idle trainer busy-waits polling the trajectory queue. Now we yield the thread. 

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/mlagents/trainers/trainer/rl_trainer.py
+++ b/ml-agents/mlagents/trainers/trainer/rl_trainer.py
@@ -152,7 +152,7 @@ class RLTrainer(Trainer):  # pylint: disable=abstract-method
                     except AgentManagerQueue.Empty:
                         break
                 if self.threaded and not _queried:
-                    # Avoid busy-waiting
+                    # Yield thread to avoid busy-waiting
                     time.sleep(0.0001)
         if self.should_still_train:
             if self._is_ready_update():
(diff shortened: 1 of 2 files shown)


ml-agents/mlagents/trainers/trainer/rl_trainer.py
```diff
@@ -142,12 +143,17 @@ def advance(self) -> None:
...
                 for _ in range(traj_queue.qsize()):
+                    _queried = True
                     try:
                         t = traj_queue.get_nowait()
                         self._process_trajectory(t)
                     except AgentManagerQueue.Empty:
                         break
+                if self.threaded and not _queried:
+                    # Avoid busy-waiting
+                    time.sleep(0.0001)
```
ervteng, line 156: @harperj the alternative solution here might be to use a blocking get with timeout when not `_queried`. (similar to old solution with `empty_queue`). I've tested both and it seems the performance is similar.

harperj: If this works to yield I think it's fine.  You could also make this `time.sleep(0)` I believe.

ervteng: IIRC time.sleep(0) only works on some platforms, annoyingly.

PR #3856, opened 2020-04-26
Release tooling

chriselion:
### Proposed change(s)
Previously, when we published to pypi (python package index), we had an extra check that the library version matched the tag being applied (this is the process that CircleCI recommends). This won’t work in the upcoming release, because we’ll be tagging `release_1` but publishing `0.16.0` libraries.
This PR adds an extra `__release_tag__` to the files that also contain the version number. and compare that against the tag being applied instead. This tag is now adjusted via the same script that we use to update the version numbers. Will document in the release checklist when it’s ready.

https://circleci.com/blog/continuously-deploying-python-packages-to-pypi-with-circleci/)

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (see above)

### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -177,6 +177,14 @@ jobs:
       directory:
         type: string
         description: Local directory to use for publishing (e.g. ml-agents)
+      username:
+        type: string
+        description: pypi username
+        default: mlagents
+      test_repository_args:
+        type: string
+        description: Optional override repository URL. Only use for tests, leave blank for "real"
+        default:
     docker:
       - image: circleci/python:3.6
     steps:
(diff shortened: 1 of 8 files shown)


chriselion:
Confirmed that `release_1_test2` tag worked on https://github.com/Unity-Technologies/ml-agents/pull/3857/files and uploaded to the test pypi: https://test.pypi.org/project/mlagents/#history

harperj:
@chriselion just to clarify, the separate release tag and version numbers are because the releases are decoupled with Python package versioning?  I'm assuming this is because of the Unity package version being different from Python in the future?

chriselion:
@harperj Yes, the plan for this release is:
Python version: `0.16.0`
C# version: `1.0.0`
Release tag: `release_1`

PR #3866, opened 2020-04-27
use python to call markdown-link-check (instead of bash magic)

chriselion:
### Proposed change(s)
Not entirely sure what was going on before, but we should have been failing some link checks and weren't. markdown-link-check only accepts one file at a time (others are ignored) but pre-commit wants to pass a few batches at a time; previously we got around this with some bash magic, but just having python make individual calls should be a lot cleaner.

This fixes the only failing case on master; will port to release branch too.

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -84,11 +84,8 @@ repos:
     hooks:
     -   id: markdown-link-check
         name: markdown-link-check
-        # markdown-link-check doesn't support multiple files on the commandline, so this hacks around that.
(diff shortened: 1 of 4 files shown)


chriselion:
@mmattar Did you say you saw a missing link a few weeks ago? This should fix checking for it.

PR #3870, opened 2020-04-27
[Bug fix] Hard reset when team changes

andrewcoh:
### Proposed change(s)

This PR enforces a full environment reset when the team is changed by the ghost trainer.  The reason for this is that due to the time horizon hyperparameter, some trajectories may still be in progress when the team is changed.  When the trajectory is finally added to the buffer, it was essentially collected by two different agents, one of which can be very very off-policy.  This PR forces a full reset when the team is changed which terminates any trajectories in progress and throws them away. 

This is accomplished by treating the GhostController as the SamplerManager or MetaCurriculum and allowing it to be a condition by which to initiate a full reset. 

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


ml-agents/mlagents/trainers/tests/test_simple_rl.py
```diff
@@ -143,6 +144,7 @@ def _check_environment_trains(
         StatsReporter.add_writer(debug_writer)
         # Make sure threading is turned off for determinism
         trainer_config["threading"] = False
+        ghost_controller = GhostController()
```
chriselion, line 147: Does this need to be passed in as an arg? Looks like from the old code, the TrainerController could the instance from the TrainerFactory (but maybe I'm missing something).

ml-agents/mlagents/trainers/ghost/controller.py
```diff
@@ -35,6 +37,17 @@ def get_learning_team(self) -> int:
         """
         return self._learning_team
 
+    @property
+    def reset(self) -> bool:
```
chriselion, line 41: nit `should_reset` (or `should_reset()` method)

ml-agents/mlagents/trainers/ghost/controller.py
```diff
@@ -26,6 +26,8 @@ def __init__(self, maxlen: int = 10):
         self._learning_team: int = -1
         # Dict from team id to GhostTrainer for ELO calculation
         self._ghost_trainers: Dict[int, GhostTrainer] = {}
+        # Signals to the trainer control to perform a hard reset
+        self._reset = False
```
chriselion, line 30: nit: Either `_should_reset` or `_changed_teams` (I think I prefer the latter).

chriselion: or I guess `_changed_training_team` is closest to the method name.

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -284,7 +287,7 @@ def reset_env_if_ready(self, env: EnvManager, steps: int) -> None:
             and (self.resampling_interval)
             and (steps % self.resampling_interval == 0)
         )
-        if meta_curriculum_reset or generalization_reset:
+        if meta_curriculum_reset or generalization_reset or self.ghost_controller.reset:
```
chriselion, line 290:
```suggestion
        ghost_controller_reset = self.ghost_controller.reset
        if meta_curriculum_reset or generalization_reset or ghost_controller_reset:
```
I think this is a little better if you were using both self-play and a curriculum or generalization. If both things wanted a reset, you'd reset two steps in a row  - the ghost controller wouldn't clear the flag because it would short-circuit the evaluation here. Unlikely to happen, but potentially annoying if it did.

config/trainer_config.yaml
```diff
@@ -135,6 +135,7 @@ Tennis:
     buffer_size: 10240
     hidden_units: 256
     time_horizon: 1000
+    threaded: false
```
chriselion, line 138: Maybe you've already done this, but it might be good to try retraining with just the reset change and just disabling threading, so that we're not conflating the effects.

ml-agents/mlagents/trainers/tests/test_simple_rl.py
andrewcoh: you aren't missing anything. I reverted that and its much cleaner

config/trainer_config.yaml
andrewcoh: I'm running both but just using this branch to do that. I won't merge until both results are in.

ml-agents/mlagents/trainers/trainer_controller.py
andrewcoh: Done.

ml-agents/mlagents/trainers/ghost/controller.py
andrewcoh: Done.

ml-agents/mlagents/trainers/ghost/controller.py
andrewcoh: Done.

ml-agents/mlagents/trainers/tests/simple_test_envs.py
```diff
@@ -232,6 +232,7 @@ def _make_batched_step(
         decision_step = DecisionSteps(m_vector_obs, m_reward, m_agent_id, action_mask)
         terminal_step = TerminalSteps.empty(self.behavior_spec)
         if done:
+            self.final_rewards[name].append(self.rewards[name])
```
andrewcoh, line 235: Had to do this for the reset.

ml-agents/mlagents/trainers/tests/test_trainer_controller.py
```diff
@@ -3,13 +3,17 @@
 
 from mlagents.tf_utils import tf
 from mlagents.trainers.trainer_controller import TrainerController
+from mlagents.trainers.ghost.controller import GhostController
 from mlagents.trainers.sampler_class import SamplerManager
 
+TRAINER_FACTORY = MagicMock()
```
andrewcoh, line 9: @ervteng is there a better way to mock this out?

chriselion: I'd say slightly cleaner to do in the functions where it's used instead of globally.

andrewcoh: Done.

chriselion:
:shipit:

PR #3885, opened 2020-04-28
[bug-fix] Fix exception thrown when quitting in-editor training from editor

ervteng:
### Proposed change(s)

An exception was being thrown when the communicator was exited from the editor, causing the trainer to throw an error and not save the model. This PR introduces a new type of exception that is only triggered when communication was exited purposefully, and changes the trainer_controller to catch this exception and quit gracefully. 

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents-envs/mlagents_envs/exception.py
+++ b/ml-agents-envs/mlagents_envs/exception.py
@@ -24,7 +24,7 @@ class UnityCommunicationException(UnityException):
 
 class UnityCommunicatorStoppedException(UnityException):
     """
-    The communicator has stopped gracefully.
+    Raised when communicator has stopped gracefully.
     """
 
     pass
--- a/ml-agents-envs/mlagents_envs/exception.py
+++ b/ml-agents-envs/mlagents_envs/exception.py
@@ -22,6 +22,14 @@ class UnityCommunicationException(UnityException):
     pass
 
 
+class UnityCommunicatorStoppedException(UnityException):
+    """
+    The communicator has stopped gracefully.
+    """
+
+    pass
+
+
 class UnityObservationException(UnityException):
     """
     Related to errors with receiving observations.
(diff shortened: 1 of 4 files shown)


ml-agents-envs/mlagents_envs/exception.py
```diff
@@ -22,6 +22,14 @@ class UnityCommunicationException(UnityException):
     pass
 
 
+class UnityCommunicatorStoppedException(UnityException):
+    """
+    The communicator has stopped gracefully.
```
harperj, line 27:
```suggestion
    Raised when communicator has stopped gracefully.
```

ml-agents/mlagents/trainers/trainer_controller.py
```diff
@@ -234,12 +235,15 @@ def start_learning(self, env_manager: EnvManager) -> None:
...
             if self.train_model:
                 self._save_model_when_interrupted()
 
-            if isinstance(ex, KeyboardInterrupt):
+            if isinstance(ex, KeyboardInterrupt) or isinstance(
+                ex, UnityCommunicatorStoppedException
+            ):
                 pass
             else:
                 # If the environment failed, we want to make sure to raise
```
harperj, line 249: We should still make sure the `.nn` is always saved, right?

ervteng: Added `finally` block

chriselion: Hang on, I'm not convinced that the `finally` will happen, since you're raising in the except.

Might be easier to do
```
if self.train_model:
    self._export_graph()
if isinstance(...):
    pass
else:
    raise ex
```

chriselion: Never mind, I'm convinced. Carry on.

PR #3927, opened 2020-05-06
Documentation fix: MonoBehaviour.OnAwake -> Awake

ReneB:
`MonoBehaviour` only calls methods that it knows of in advance. `OnAwake` is not expected to exist and is thus never called. Since there is no real IntelliSense or anything for these "magical" `MonoBehaviour` methods, it's easy to accidentally call the method OnAwake and have it never be called.

### Proposed change(s)

Fix a slight error in the Migrating documentation that might catch developers off-guard otherwise.

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Other (please describe)

### Checklist


surfnerd:
Thanks for finding this.

PR #3936, opened 2020-05-07
[refactor] Structure configuration files into classes 

ervteng:
### Proposed change(s)

This PR implements the changes described in this [design doc](https://docs.google.com/document/d/1dTp5y1MRqpgsR2XfgABSId3hcX_pVNiklchjJDi1EXo/edit?ts=5ea748cf#heading=h.v7t1ynb2u4tr). In particular, it replaces the configuration dictionary (called `trainer_params` in many files) with objects based on the `attrs` library. These are more flexible than NamedTuples and support smart defaults, validators, typing, and can be structured/unstructured from Dicts based on the `cattrs` library. 

This PR also assigns reasonable defaults to every hyperparameter, and `mlagents-learn` now allows for training without any configuration file specified (defaults to PPO with extrinsic reward signal). In the future, it will allow for defaults that are being set relative to other defaults (as done in the SelfPlaySettings).  

The core of the changes are in [settings.py](https://github.com/Unity-Technologies/ml-agents/blob/develop-attrs/ml-agents/mlagents/trainers/settings.py). 

Attrs:
https://www.attrs.org/en/stable/

Cattrs:
https://cattrs.readthedocs.io/en/latest/

This PR does not yet refactor how the settings are passed down from Trainer->Policy-> Optimizer. The other two objects still receive the entire `TrainerSettings` object. 
TODO: Improved testing for error handling in `settings.py`, ~Update docs.~ 

NOTE: The appearance of `output_path` as `default` in the configs is a bit odd, as `output_path` is not settable through the configuration (it's determined by the `run-id`). I plan to remove `output_path` in a follow-up PR. 

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [x] Breaking change
- [x] Documentation update
- [ ] Other (please describe)

### Checklist


ml-agents/mlagents/trainers/trainer_util.py
```diff
@@ -90,27 +90,10 @@ def initialize_trainer(
...
-    if init_path is not None:
-        trainer_parameters["init_path"] = os.path.join(init_path, brain_name)
-    trainer_parameters["keep_checkpoints"] = keep_checkpoints
-    if brain_name in trainer_config:
-        _brain_key: Any = brain_name
-        while not isinstance(trainer_config[_brain_key], dict):
-            _brain_key = trainer_config[_brain_key]
-        trainer_parameters.update(trainer_config[_brain_key])
-
+    print(trainer_config)
```
harperj, line 93: Remove?

ml-agents/mlagents/trainers/trainer_util.py
```diff
@@ -60,7 +60,7 @@ def generate(self, brain_name: str) -> Trainer:
 
 
 def initialize_trainer(
-    trainer_config: Any,
+    trainer_config: TrainerSettings,
```
harperj, line 63: minor but.. `trainer_settings` if we are calling the class `TrainerSettings`?

ml-agents/mlagents/trainers/trainer_util.py
```diff
@@ -1,6 +1,5 @@
 import os
-import yaml
-from typing import Any, Dict, TextIO
+from typing import Any, Dict
```
harperj, line 2: The `Any` here refers to the trainer_config param to `TrainerFactory`.  Should that be updated with the new class name?

ml-agents/mlagents/trainers/trainer/trainer.py
```diff
@@ -25,7 +26,7 @@ class Trainer(abc.ABC):
     def __init__(
         self,
         brain_name: str,
-        trainer_parameters: dict,
+        trainer_parameters: TrainerSettings,
```
harperj, line 29: trainer_parameters -> trainer_settings?

ml-agents/mlagents/trainers/trainer/trainer.py
```diff
@@ -80,7 +81,7 @@ def get_max_steps(self) -> int:
         Returns the maximum number of steps. Is used to know when the trainer should be stopped.
         :return: The maximum number of steps of the trainer
         """
-        return int(float(self.trainer_parameters["max_steps"]))
+        return int(float(self.trainer_parameters.max_steps))
```
harperj, line 84: Does this need the `int(float(_))`?  I think we should be able to manage the type coersion in the parsing logic upstream of this.

ml-agents/mlagents/trainers/trainer/rl_trainer.py
```diff
@@ -40,7 +33,8 @@ def __init__(self, *args, **kwargs):
         }
         self.update_buffer: AgentBuffer = AgentBuffer()
         self._stats_reporter.add_property(
-            StatsPropertyType.HYPERPARAMETERS, self.trainer_parameters
+            StatsPropertyType.HYPERPARAMETERS,
+            cattr.unstructure(self.trainer_parameters),
```
harperj, line 37: Minor but this feels like it leaks the abstraction of the `TrainerSettings` implementation.  Perhaps this could be implemented as `TrainerSettings.as_dict`?

ml-agents/mlagents/trainers/settings.py
```diff
@@ -0,0 +1,365 @@
...
+    swap_steps: int = 10000
+    window: int = 10
+    play_against_latest_model_ratio: float = 0.5
+    initial_elo: float = 1200.0
+
+
+@attr.s(auto_attribs=True)
+class TrainerSettings:
+    # Edit these two fields to add new trainers #
+    class TrainerType(Enum):
```
harperj, line 201: IMO this class and the associated static method should live fully outside of `TrainerSettings`.  

So...

```python
class TrainerType(Enum):
  PPO: str = "ppo"
  SAC: str = "sac"

  def to_settings(self) -> type:
    _mapping = {
      self.PPO: PPOSettings,
      self.SAC: SACSettings,
    }
    return _mapping[self]
```

I also used self-references for the members, which might arguably be less clear.  That isn't necessary.

ml-agents/mlagents/trainers/settings.py
```diff
@@ -0,0 +1,365 @@
...
+
+
+@attr.s(auto_attribs=True)
+class BehavioralCloningSettings:
+    demo_path: str
+    steps: int = 0
+    strength: float = 1.0
+    samples_per_update: int = 0
+    num_epoch: Optional[int] = None
+    batch_size: Optional[int] = None
```
harperj, line 111: Could you clarify what it means for these to be `None`?  Might be worth a comment.

ml-agents/mlagents/trainers/settings.py
```diff
@@ -0,0 +1,365 @@
...
+        except KeyError:
+            raise TrainerConfigError(f"Unknown reward signal type {key}")
+    return d_final
+
+
+def defaultdict_to_dict(d: DefaultDict) -> Dict:
+    return {key: cattr.unstructure(val) for key, val in d.items()}
+
+
+def dict_to_defaultdict(d: Dict, t: type) -> DefaultDict:
```
harperj, line 84: Seems like this really is intended for `TrainerSettings` only since `t` is unused.  Maybe you could make it a class method?

ml-agents/mlagents/trainers/settings.py
```diff
@@ -0,0 +1,365 @@
...
+            d_copy[key] = int(
+                float(val)
+            )  # In some configs, max steps was specified as a float
+        else:
+            d_copy[key] = check_and_structure(key, val, t)
+    return t(**d_copy)
+
+
+def rewardsignal_settings_to_cls(d: Mapping) -> Any:
+    if d is None:
```
harperj, line 66: Should / could we just prevent `None` from being passed here?  If not, `d` should be an `Optional[Mapping]`.

ml-agents/mlagents/trainers/settings.py
```diff
@@ -0,0 +1,365 @@
...
+from enum import Enum
+import collections
+import argparse
+
+from mlagents.trainers.cli_utils import StoreConfigFile, DetectDefault, parser
+from mlagents.trainers.cli_utils import load_config
+from mlagents.trainers.exception import TrainerConfigError
+from mlagents.trainers.models import LearningRateSchedule, EncoderType
+
+
```
harperj, line 13: In general these parsing helper methods at the top could use a little love, IMO.  They scatter the parsing / error handling a bit.  Ideally there would be a way to embed them closer to the types/fields themselves.  If that's difficult for some reason at least some comments and/or more verbose field names might be helpful

ml-agents/mlagents/trainers/trainer/trainer.py
ervteng: You're right, this is already enforced during the structuring. Removed.

ml-agents/mlagents/trainers/trainer/trainer.py
ervteng: Yep! Changed throughout

ml-agents/mlagents/trainers/trainer_util.py
ervteng: 👍

ml-agents/mlagents/trainers/trainer/rl_trainer.py
ervteng: Made a mixin that defines the `as_dict` method and gave it to TrainerSettings and RunOptions - should clean up the code in learn.py a bit.

ml-agents/mlagents/trainers/settings.py
ervteng: Split off the type enums (TrainerType and RewardSIgnalType) as separate classes.

ml-agents/mlagents/trainers/settings.py
ervteng: For the BC module, not specifying num_epoch and batch_size means that it will use the current values for these from the trainer. This logic happens in the BC module itself - I couldn't find a good way to put the logic in `settings.py`. Will add a comment

ml-agents/mlagents/trainers/settings.py
ervteng: True, these should never be None unless we have a malformed config (e.g. if user puts `env_settings: null` in their config). I now throw an error if a None (or non-Mapping) is received.

ml-agents/mlagents/trainers/settings.py
ervteng: I've moved them into their appropriate classes as static methods - the downside being that certain classes are now quite large. But I think it's an improvement over having those methods up at the top.

config/upgrade_config.py
```diff
@@ -0,0 +1,110 @@
+import attr
```
andrewcoh, line 1: How long do you think we'll need to keep this script around?

docs/Training-Configuration-File.md
```diff
@@ -26,20 +26,20 @@ choice of the trainer (which we review on subsequent sections).
...
+| `max_steps`              | (default = `500000`) Total number of experience points that must be collected from the simulation before ending the training process. <br><br>Typical range: `5e5` - `1e7`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |
```
andrewcoh, line 36: 'experience points'

ml-agents/mlagents/trainers/components/bc/module.py
```diff
@@ -36,14 +31,16 @@ def __init__(
...
+        self.model = BCModel(policy, self.current_lr, settings.steps)
         _, self.demonstration_buffer = demo_to_buffer(
-            demo_path, policy.sequence_length, policy.brain
+            settings.demo_path, policy.sequence_length, policy.brain
         )
 
-        self.batch_size = batch_size if batch_size else default_batch_size
-        self.num_epoch = num_epoch if num_epoch else default_num_epoch
+        self.batch_size = (
+            settings.batch_size if settings.batch_size else default_batch_size
```
andrewcoh, line 41: why does this need to be a tuple now?

ml-agents/mlagents/trainers/components/reward_signals/curiosity/signal.py
```diff
@@ -26,9 +20,11 @@ def __init__(
         :param encoding_size: The size of the hidden encoding layer for the ICM
```
andrewcoh, line 20: param docstring :(

config/upgrade_config.py
ervteng: Hopefully only for 1-2 releases. Or perhaps it should be released with Trainer 1.0 and removed after that.

ml-agents/mlagents/trainers/components/bc/module.py
ervteng: It's not a tuple, it's the way `black` likes to format long lines 😛

ml-agents/mlagents/trainers/ppo/trainer.py
```diff
@@ -76,9 +80,9 @@ def _check_param_keys(self):
         # Check that batch size is greater than sequence length. Else, throw
         # an exception.
         if (
-            self.trainer_parameters["sequence_length"]
-            > self.trainer_parameters["batch_size"]
-            and self.trainer_parameters["use_recurrent"]
+            self.trainer_settings["sequence_length"]
+            > self.trainer_settings["batch_size"]
+            and self.trainer_settings["use_recurrent"]
```
harperj, line 85: I think this can be:
```suggestion
            self.trainer_settings.sequence_length
            > self.trainer_settings.batch_size
            and self.trainer_settings.use_recurrent
```

ml-agents/mlagents/trainers/sac/optimizer.py
```diff
@@ -44,32 +45,35 @@ def __init__(self, policy: TFPolicy, trainer_params: Dict[str, Any]):
...
-                self.tau = trainer_params.get("tau", 0.005)
-                self.burn_in_ratio = float(trainer_params.get("burn_in_ratio", 0.0))
+                self.act_size = policy.act_size
+                policy_network_settings = policy.network_settings
+                h_size = policy_network_settings.hidden_units
+                num_layers = policy_network_settings.num_layers
+                vis_encode_type = policy_network_settings.vis_encode_type
+
+                self.tau = hyperparameters.tau
+                self.burn_in_ratio = 0.0
```
harperj, line 65: Looks like this used to be a trainer param but is not just being set to 0.0?  Any reason why?

ml-agents/mlagents/trainers/sac/trainer.py
```diff
@@ -47,75 +47,44 @@ def __init__(
...
         super()._check_param_keys()
         # Check that batch size is greater than sequence length. Else, throw
         # an exception.
         if (
-            self.trainer_parameters["sequence_length"]
-            > self.trainer_parameters["batch_size"]
-            and self.trainer_parameters["use_recurrent"]
+            self.trainer_settings["sequence_length"]
+            > self.trainer_settings["batch_size"]
+            and self.trainer_settings["use_recurrent"]
```
harperj, line 87:
```suggestion
            self.trainer_settings.sequence_length
            > self.trainer_settings.batch_size
            and self.trainer_settings.use_recurrent
```

ml-agents/mlagents/trainers/settings.py
```diff
@@ -0,0 +1,375 @@
...
+    def _set_default_hyperparameters(self):
+        return self.trainer_type.to_settings()()
+
+    network_settings: NetworkSettings = attr.ib(factory=NetworkSettings)
+    reward_signals: Dict[RewardSignalType, RewardSignalSettings] = attr.ib(
+        factory=lambda: {RewardSignalType.EXTRINSIC: RewardSignalSettings()}
+    )
+    init_path: Optional[str] = None
+    output_path: str = "default"
+    # TODO: Remove parser default and remove from CLI
```
harperj, line 194: Address before merging?

ml-agents/mlagents/trainers/sac/trainer.py
ervteng: Ah, this method isn't actually used anymore, the check is done in `settings.py` in `_check_batch_size_seq_length`. I've removed them from PPO and SAC.

ml-agents/mlagents/trainers/ppo/trainer.py
ervteng: Ah, this method isn't actually used anymore, the check is done in `settings.py` in `_check_batch_size_seq_length`. I've removed them from PPO and SAC.

ml-agents/mlagents/trainers/sac/optimizer.py
ervteng: This wasn't configurable before (also set to 0 as a constant at the top of the file). Currently the burn-in feature doesn't actually work properly with values greater than 0. Will fix or remove in a PR that just touches that.

ml-agents/mlagents/trainers/settings.py
ervteng: Done!

PR #3997, opened 2020-05-20
add debugging option to ModelOverrider, handle scene resets

chriselion:
### Proposed change(s)
The Basic scene is currently broken for inference testing. This is because we reset the scene (as of https://github.com/Unity-Technologies/ml-agents/pull/3789/files) when the episode ends; since the counts are per-ModelOverride and per-Agent, they also get reset when the scene does.

This PR adds a static counter for the number of episodes completed and steps taken; this does what we want for a single Agent, and I think it will behave reasonably for multiple agents too (although we don't have this scenario right now).

Also adds a debug field to override the commandline options used for configuring the model overrider, so that we can test it in the Editor more easily (and thus with a debugger).

https://github.com/Unity-Technologies/ml-agents/pull/3789/files
https://jira.unity3d.com/browse/MLA-1021

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ModelOverrider.cs
+++ b/Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ModelOverrider.cs
@@ -5,6 +5,9 @@ using Unity.Barracuda;
 using System.IO;
 using Unity.MLAgents;
 using Unity.MLAgents.Policies;
+#if UNITY_EDITOR
+using UnityEditor;
+#endif
 
 namespace Unity.MLAgentsExamples
 {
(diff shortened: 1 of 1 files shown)


Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ModelOverrider.cs
```diff
@@ -38,8 +41,27 @@ public class ModelOverrider : MonoBehaviour
         int m_MaxEpisodes;
 
         int m_NumSteps;
+        int m_PreviousNumSteps;
+        int m_PreviousAgentCompletedEpisodes;
```
xiaomaogy, line 45: A little bit confused about the m_ prefix, does it mean max or something else?

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ModelOverrider.cs
```diff
@@ -38,8 +41,27 @@ public class ModelOverrider : MonoBehaviour
...
 
         bool m_QuitOnLoadFailure;
+        [Tooltip("Debug values to be used in place of the command line for overriding models.")]
+        public string debugCommandLineOverride;
+
+        // Static values to keep track of completed episodes and steps across resets
+        // These are updated in OnDisable.
+        static int s_PreviousAgentCompletedEpisodes;
+        static int s_PreviousNumSteps;
```
xiaomaogy, line 54: And here s_ prefix means static?

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ModelOverrider.cs
chriselion: From the [coding standard](https://github.cds.internal.unity3d.com/unity/com.unity.coding/blob/master/UnityCoding/Packages/com.unity.coding/Coding~/Conventions/CSharpReference.cs)
```
        // |    - Use prefix + PascalCase for non-public field naming.
        // |        - Prefixes: m_ = instance field, s_ = static readwrite field, k_ = const
```
I usually think of `m` as "member"

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ModelOverrider.cs
chriselion: Yes

PR #4026, opened 2020-05-27
fix missing metafile checks

chriselion:
### Proposed change(s)
This has been broken for a while because it doesn't check the package directory. This led to shipping with a missing metafile, which was discovered when installing the package from packman.

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/utils/validate_meta_files.py
+++ b/utils/validate_meta_files.py
@@ -2,34 +2,56 @@ import os
 
 
 def main():
-    asset_path = "Project/Assets"
+    asset_paths = ["Project/Assets", "com.unity.ml-agents"]
     meta_suffix = ".meta"
     python_suffix = ".py"
+    whitelist = frozenset(
+        [
+            "com.unity.ml-agents/.editorconfig",
+            "com.unity.ml-agents/.gitignore",
+            "com.unity.ml-agents/.npmignore",
+            "com.unity.ml-agents/Tests/.tests.json",
+        ]
+    )
+    ignored_dirs = {"Documentation~"}
 
     num_matched = 0
 
     unmatched = set()
 
-    for root, dirs, files in os.walk(asset_path):
-        dirs = set(dirs)
-        files = set(files)
+    for asset_path in asset_paths:
+        for root, dirs, files in os.walk(asset_path, topdown=True):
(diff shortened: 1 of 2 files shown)


chriselion:
Testing in https://github.com/Unity-Technologies/ml-agents/pull/4027

chriselion:
Wow, didn't even need the other test, because I forgot `EnumReflectionSensor.cs.meta`

chriselion:
@surfnerd Any thoughts on keeping this like it is (manual script) vs converting it to a pre-commit hook? We can also extend the check to make sure that the corresponding file is tracked in git...

utils/validate_meta_files.py
```diff
@@ -2,34 +2,56 @@
...
-                expected = f.replace(meta_suffix, "")
-            else:
-                expected = f + meta_suffix
-
-            if expected not in combined:
-                unmatched.add(os.path.join(root, f))
-            else:
-                num_matched += 1
+    for asset_path in asset_paths:
+        for root, dirs, files in os.walk(asset_path, topdown=True):
```
surfnerd, line 23: isn't `topdown` `True` by default?

utils/validate_meta_files.py
```diff
@@ -2,34 +2,56 @@
...
+            for ignored in ignored_dirs:
+                try:
+                    dirs.remove(ignored)
+                except ValueError:
+                    pass
+
+            dirs = set(dirs)
+            files = set(files)
+
+            combined = dirs | files
```
surfnerd, line 34: that's cool

chriselion: Yeah, python sets have operator overloads. `&` gives the intersection, etc.
https://docs.python.org/3.7/library/stdtypes.html#set-types-set-frozenset

utils/validate_meta_files.py
chriselion: Good point, I can undo that.

surfnerd:
pre-commit sounds fine too.  But I like it in CI since you don't have to install any pre-commit hooks locally in order to catch it.  We could do both I guess but this seems like the best "catch-all" version.

PR #4035, opened 2020-05-28
[versioning] Save ML-Agents version in checkpoints and check on load

ervteng:
### Proposed change(s)

This PR saves the semantic version of the trainer package (`major`, `minor`, and `patch`) as 3 variables in the TF graph. It is also exported into the .nn file. This lets us check whether a checkpoint is being loaded from the same version of ML-Agents, and (in the future) check which version of ML-Agents created a particular NN file. 

Note that this is different than the existing version number in the NN, which is checked by C#. That number corresponds to the input and output tensors. It is possible, when upgrading the Trainer code, for an NN file to remain compatible with C# but not be loadable into Python for training (e.g. if the network architecture changes). 

Currently, we throw a warning if the versions don't match when a user tries to load a checkpoint.

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/mlagents/trainers/policy/tf_policy.py
+++ b/ml-agents/mlagents/trainers/policy/tf_policy.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Tuple
 import abc
 import os
 import numpy as np
@@ -14,6 +14,7 @@ from mlagents_envs.base_env import DecisionSteps
 from mlagents.trainers.models import ModelUtils
 from mlagents.trainers.settings import TrainerSettings, NetworkSettings
 from mlagents.trainers.brain import BrainParameters
+from mlagents.trainers import __version__
 
 
 logger = get_logger(__name__)
@@ -46,7 +47,6 @@ class TFPolicy(Policy):
         :param brain: The corresponding Brain for this policy.
         :param trainer_settings: The trainer parameters.
         """
-        self._version_number_ = 2
         self.m_size = 0
         self.trainer_settings = trainer_settings
         self.network_settings: NetworkSettings = trainer_settings.network_settings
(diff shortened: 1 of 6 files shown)


ml-agents/mlagents/trainers/policy/tf_policy.py
```diff
@@ -14,11 +14,17 @@
...
 from mlagents.trainers.brain import BrainParameters
+from mlagents.trainers import __version__
 
 
 logger = get_logger(__name__)
 
 
+# This is the version number of the inputs and outputs of the model, and
+# determines compatibility with inference in Barracuda.
+API_VERSION_NUMBER = 2
```
chriselion, line 25: `MODEL_FORMAT_VERSION_NUMBER`? "API" doesn't really feel right here.

ml-agents/mlagents/trainers/policy/tf_policy.py
```diff
@@ -114,6 +120,35 @@ def create_tf_graph(self):
...
         pass
 
+    @staticmethod
+    def _convert_version_string(version_string: str) -> Tuple[int, ...]:
+        """
+        Converts the version string into a Tuple of ints (major_ver, minor_ver, patch_ver).
+        :param version_string: The semantic-versioned version string (X.Y.Z).
+        :return: A Tuple containing (major_ver, minor_ver, patch_ver).
+        """
+        split_ver = version_string.split(".")[0:3]  # Remove dev tag
```
chriselion, line 130: I think you can use `distutils.version.LooseVersion` to simplify this:
```
>>> from distutils.version import LooseVersion
>>> v = LooseVersion("1.2.3.dev4")
>>> v
LooseVersion ('1.2.3.dev4')
>>> v.version
[1, 2, 3, 'dev', 4]
>>> v.version[0:3]
[1, 2, 3]
```

ervteng: Oh, that is very handy! Updated to use the LooseVersion class.

PR #4087, opened 2020-06-05
[refactor] Remove nonfunctional `output_path` option from TrainerSettings

ervteng:
### Proposed change(s)

In the settings change (#3936) `output_path` was left as a trainer configuration with the value `default`, but changing it had no effect, as the trainer always saves in `results/run_id/behavior_name`. This PR removes that option and removes it from the example YAMLs. Instead, the `results/run_id` directory is passed to the Trainer and Policy as an artifact directory. In the future, if we want this to be user-configurable, we should do it at the level of `run-id` rather than trainer, so that all trainers for that run save in the same directory

This also removes the `run_id` that was being passed to the trainer and not used. 

NOTE: Likely to cherry-pick this commit into the release branch as it avoid some confusion for our users. 

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


ervteng:
> Do you think it is safe to cherry pick such a large change into this release?

I agree it's a bit of a risk. But I think the tradeoff in terms of user experience is worth it. What do you think?

PR #4157, opened 2020-06-24
Add TargetController/OrientationCubeController Components & Bugfix

Hunter-Unity:
### Proposed change(s)

1. Move target spawning and target collision logic into standalone component for easier re-use. Includes collision callback functionality.
2. Move Orientation cube logic into standalone component for easier re-use.
3. Crawler & Walker environments now use these new components & have been retrained to confirm it works.
4. Walker's reward function slightly modified. The facing reward scalar went from .01 to .02, and the head height reward scalar went from .01 to .005. The reason is to prevent walker from running/walking backwards. The reason for the change in head height reward is that Walker will run faster without the reward but it runs on it's tippy toes. So instead of removing this reward completely, i've just reduced it.
5. Walker Bug Fix: The Walker prefab had a missing reference to the Orientation Cube. The fix for this is included in this PR. Learns better now.
6. TargetController adds logic to detect if target falls off a platform and will respawn the target if it falls.
7. Removes TargetContact logic from the JointDriveController

*The Worm env will be updated with these changes in a separate PR with an updated version of the ragdoll

### Types of change(s)

- [x] Bug fix
- [x] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


Project/Assets/ML-Agents/Examples/Crawler/Scripts/CrawlerAgent.cs
```diff
@@ -98,10 +109,11 @@ public void CollectObservationBodyPart(BodyPart bp, VectorSensor sensor)
     /// </summary>
     public override void CollectObservations(VectorSensor sensor)
     {
-        sensor.AddObservation(Quaternion.FromToRotation(body.forward, orientationCube.transform.forward));
-
+        sensor.AddObservation(body.rotation);
```
awjuliani, line 112: Why did this observation change?

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
```diff
@@ -0,0 +1,143 @@
...
+            }
+        }
+
+        void Update()
+        {
+            if (respawnIfFallsOffPlatform)
+            {
+                if (transform.position.y < m_startingPos.y - fallDistance)
+                {
+                    Debug.LogError($"{transform.name} Fell Off Platform");
```
awjuliani, line 67: Do we want to log this as an error?

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
```diff
@@ -0,0 +1,143 @@
...
+                triggerIsTouching = true;
+                onTriggerEnterEvent.Invoke(col);
+            }
+        }
+
+        private void OnTriggerStay(Collider col)
+        {
+            if (col.CompareTag(tagToDetect))
+            {
+                triggerIsTouching = true;
```
awjuliani, line 129: `triggerIsTouching` and `colliderIsTouching` are both bools that can be set in the inspector, but it seems they aren't actually used for anything. They are just turned on and off. Can we remove them?

Project/Assets/ML-Agents/Examples/Crawler/Scripts/CrawlerAgent.cs
```diff
@@ -71,11 +65,28 @@ public override void Initialize()
...
+        orientationCube.UpdateOrientation(body, target.transform);
+    }
+
     /// <summary>
     /// Add relevant information on each body part to observations.
     /// </summary>
     public void CollectObservationBodyPart(BodyPart bp, VectorSensor sensor)
     {
+        //GROUND CHECK
         sensor.AddObservation(bp.groundContact.touchingGround ? 1 : 0); // Whether the bp touching the ground
```
vincentpierre, line 90: I think there is a AddObservation for boolean. Maybe is this instead of casting to int ?

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
```diff
@@ -0,0 +1,143 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Moves target to a random position within specified radius.
+        /// </summary>
+        public void MoveTargetToRandomPosition()
+        {
+            var newTargetPos = m_startingPos + (Random.insideUnitSphere * spawnRadius);
```
vincentpierre, line 78: Is it a 3D sphere or a 2D sphere ? I think we want it to be 2D. Or it might spawn at `m_startingPos - spawnRadius * Vector3.up`

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
```diff
@@ -0,0 +1,143 @@
...
+        public TriggerEvent onTriggerExitEvent = new TriggerEvent();
+
+        [System.Serializable]
+        public class CollisionEvent : UnityEvent<Collision>
+        {
+        }
+
+        [Header("Collision Callbacks")] public bool colliderIsTouching;
+        public CollisionEvent onCollisionEnterEvent = new CollisionEvent();
+        public CollisionEvent onCollisionStayEvent = new CollisionEvent();
```
vincentpierre, line 48: Why do we expose those ? Do we expect users to subscribe to these events ?

Project/Assets/ML-Agents/Examples/Walker/Scripts/WalkerAgent.cs
```diff
@@ -101,10 +115,11 @@ public void CollectObservationBodyPart(BodyPart bp, VectorSensor sensor)
     /// </summary>
     public override void CollectObservations(VectorSensor sensor)
     {
-        sensor.AddObservation(Quaternion.FromToRotation(hips.forward, orientationCube.transform.forward));
-        sensor.AddObservation(Quaternion.FromToRotation(head.forward, orientationCube.transform.forward));
+        sensor.AddObservation(hips.rotation);
+        sensor.AddObservation(head.rotation);
```
vincentpierre, line 119: I thought you tested and `Quaternion.FromToRotation(hips.forward, orientationCube.transform.forward)` worked better than `hips.rotation` ?

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
```diff
@@ -0,0 +1,143 @@
...
+        public TriggerEvent onTriggerStayEvent = new TriggerEvent();
+        public TriggerEvent onTriggerExitEvent = new TriggerEvent();
+
+        [System.Serializable]
+        public class CollisionEvent : UnityEvent<Collision>
+        {
+        }
+
+        [Header("Collision Callbacks")] public bool colliderIsTouching;
+        public CollisionEvent onCollisionEnterEvent = new CollisionEvent();
```
vincentpierre, line 47: This is the only event we use and I think we should register the Callback to TouchedTarget via code rather than via UI

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
Hunter-Unity: yeah they're not currently used. was more for potential future use. I can remove.

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
Hunter-Unity: i this you're right i'll doublecheck.

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
Hunter-Unity: yeah they show as events in the editor so the user won't have to subscribe/unsubscribe via code. just drag in the object you want to send the callback to and choose the method.

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
Hunter-Unity: I think the ultimate purpose of this script is to handle as many of our current "Target" situations as possible instead of ad-hoc implementations. if we leave the other callbacks we can use this script for other things like Reacher which works with a OnTriggerStay() right now. This script could replace the ReacherGoal.cs.

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
Hunter-Unity: no i don't think so i changed it to a Log()

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
Hunter-Unity: changed. i removed both properties. will add in the future if needed

Project/Assets/ML-Agents/Examples/Crawler/Scripts/CrawlerAgent.cs
Hunter-Unity: changed. agreed, no reason to cast to an int.

Project/Assets/ML-Agents/Examples/Crawler/Scripts/CrawlerAgent.cs
Hunter-Unity: this was an oversight from some testing i was doing. i changed it back to the previous observations.

Project/Assets/ML-Agents/Examples/Walker/Scripts/WalkerAgent.cs
Hunter-Unity: this was an oversight from some testing i was doing. i changed it back to the previous observations.

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/OrientationCubeController.cs
```diff
@@ -0,0 +1,25 @@
...
+                 dirVector == Vector3.zero
+                     ? Quaternion.identity
+                     : Quaternion.LookRotation(dirVector); //get our look rot to the target
+ 
+             //UPDATE ORIENTATION CUBE POS & ROT
+             transform.SetPositionAndRotation(rootBP.position, lookRot);
+         }
+     }
+ }
+
```
awjuliani, line 25: Looks like you need an extra line here.

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
vincentpierre: That is still not a uniform circle, probability of spawning at the center of the circle will be a lot higher than the probability of spawning at the edges (because it is a squished sphere). But I guess it does not matter...

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
vincentpierre: Is the user expected to use / modify these callbacks themselves? There is no tutorial on how to do so, and the use case seems pretty limited to me.

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
vincentpierre: I like this script, the thing I would like is 
```
        [Header("Collision Callbacks")]
        public CollisionEvent onCollisionEnterEvent = new CollisionEvent();
        public CollisionEvent onCollisionStayEvent = new CollisionEvent();
        public CollisionEvent onCollisionExitEvent = new CollisionEvent();
```
and 
```
        [Header("Trigger Callbacks")]
        public TriggerEvent onTriggerEnterEvent = new TriggerEvent();
        public TriggerEvent onTriggerStayEvent = new TriggerEvent();
        public TriggerEvent onTriggerExitEvent = new TriggerEvent();
```
Keep only one (onCollisionEnterEvent), make it private / hide in inspector. Add a public method to subscribe a callback to this event. Have the Agent subscribe to the callback via code (since it already has a reference to the "TargetController"). This simplifies the inspector UI and makes it so that a user can understand what happens by looking at the code.

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
Hunter-Unity: oh ok i understand what you are saying now. yes you are correct we are doing a random point in a sphere then squishing the Y. the reason i didn't use Random.InsideUnitCircle is because a Random.InsideUnitCircle returns a Vector2 so then you have to "remap" that Vector2 to a Vector3 and then set the Y position. it adds one more line of code and yields a very similar result.

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
Hunter-Unity: I'm not necessarily against doing this, but it goes against my original design goal which is to reduce the amount of code that needs to be in the Agent script. We'd need to Subscribe/Unsubscribe via code in each agent script. Exposing these events in the editor gets rid of all of that and gives us a general use component

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/OrientationCubeController.cs
Hunter-Unity: changed. there was a single space in there that was throwing off the formatting. fixed

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
Hunter-Unity: The user will use the standard event editor workflow. Choose the event you want. Drag in the object that needs to receive the callback and choose the method you want to call. We could separate out the collision/trigger events into it's own script. That's how i do it in other projects. 
Screen Shot 2020-07-02 at 12 19 19 PM

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/TargetController.cs
vincentpierre: All right then

PR #4201, opened 2020-07-08
[bug-fix] Make StatsReporter thread-safe

ervteng:
### Proposed change(s)

It was possible that the StatsReporter global dictionary could be mutated while writing out stats. This would happen occasionally at very low values of `summary_freq`. 

This PR adds an `RLock` to all StatsReporter functions that modify a global variable, which prevents this issue. 

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/com.unity.ml-agents/CHANGELOG.md
+++ b/com.unity.ml-agents/CHANGELOG.md
@@ -36,6 +36,8 @@ empty string). (#4155)
 - Fixed an error when setting `initialize_from` in the trainer confiiguration YAML to
 `null`. (#4175)
 - Fixed issue with FoodCollector, Soccer, and WallJump when playing with keyboard. (#4147, #4174)
+- Fixed a crash in StatsReporter when using threaded trainers with very frequent summary writes
+(#4201)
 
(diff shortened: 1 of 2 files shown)


com.unity.ml-agents/CHANGELOG.md
```diff
@@ -36,6 +36,8 @@ empty string). (#4155)
 - Fixed an error when setting `initialize_from` in the trainer confiiguration YAML to
 `null`. (#4175)
 - Fixed issue with FoodCollector, Soccer, and WallJump when playing with keyboard. (#4147, #4174)
+- Fixed a crash in StatsReporter when using threaded trainers with very frequent summary writes
+(#4201)
```
chriselion, line 40:
```suggestion
- Fixed a rare crash in StatsReporter when using threaded trainers (#4201)
```
Very frequent summary writes weren't a requirement for the crash, it just increased the probability enough so that we could reproduce it reliably.

PR #4223, opened 2020-07-13
update com.unity.ml-agents.extensions to Apache 2.0 license

chriselion:
### Proposed change(s)
Change com.unity.ml-agents.extensions to use Apache 2.0 license.

We initially thought UCL would be easier, but there's concern that migrating code from one to the other could be problematic.

### DIFF
--- a/com.unity.ml-agents.extensions/LICENSE.md
+++ b/com.unity.ml-agents.extensions/LICENSE.md
@@ -1,9 +1,204 @@
-com.unity.ml-agents.extensions copyright © 2020 Unity Technologies
+com.unity.ml-agents.extensions copyright © 2020 Unity Technologies ApS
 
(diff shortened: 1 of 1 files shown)


com.unity.ml-agents.extensions/LICENSE.md
```diff
@@ -1,9 +1,204 @@
-com.unity.ml-agents.extensions copyright © 2020 Unity Technologies
+com.unity.ml-agents.extensions copyright © 2020 Unity Technologies ApS
```
chriselion, line 1: This is copied from com.unity.ml-agents/LICENSE.md (with the package name changed).

unityjeffrey:
hey @surfnerd - any issue to use Apache 2.0 if we don't plan on putting extensions package through package review?

surfnerd:
no

PR #4224, opened 2020-07-14
[MLA-1138] joint observations

chriselion:
### Proposed change(s)
* Add IJointExtractor. This is an interface that provides observations from a single joint (either old-style Joint, or part of an ArticulationBody).
* Get the joint "positions" from ArticulationBody. These are actually angles for revolute and spherical joints.
* Get joint forces for both ArticulationBody and Rigidbody Joints. These are forces applied by the solver, and indicate that a constraint is at its limit.

https://jira.unity3d.com/browse/MLA-1138

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/com.unity.ml-agents.extensions/Runtime/Sensors/ArticulationBodyJointExtractor.cs
+++ b/com.unity.ml-agents.extensions/Runtime/Sensors/ArticulationBodyJointExtractor.cs
@@ -1,0 +1,130 @@
+#if UNITY_2020_1_OR_NEWER
+
+using System.Collections.Generic;
+using UnityEngine;
+using Unity.MLAgents.Sensors;
+
+namespace Unity.MLAgents.Extensions.Sensors
+{
+    public class ArticulationBodyJointExtractor : JointExtractor
+    {
+        ArticulationBody m_Body;
+
+        public ArticulationBodyJointExtractor(ArticulationBody body)
+        {
+            m_Body = body;
+        }
+
+        public int NumObservations(PhysicsSensorSettings settings)
+        {
+            return NumObservations(m_Body, settings);
+        }
+
+        public static int NumObservations(ArticulationBody body, PhysicsSensorSettings settings)
+        {
+            if (body == null || body.isRoot)
+            {
+                return 0;
+            }
+
+            var totalCount = 0;
+            if (settings.UseJointPositions)
+            {
+                switch (body.jointType)
+                {
+                    case ArticulationJointType.RevoluteJoint:
+                    case ArticulationJointType.SphericalJoint:
+                        // Two floats per angular component
+                        totalCount += 2 * body.dofCount;
+                        break;
+                    case ArticulationJointType.FixedJoint:
+                        break; // TODO (none?)
+                    case ArticulationJointType.PrismaticJoint:
+                        // One linear component
+                        totalCount += body.dofCount;
+                        break;
+                }
+            }
+
+            return totalCount;
+        }
+
+        public int Write(PhysicsSensorSettings settings, ObservationWriter writer, int offset)
+        {
+            if (m_Body == null || m_Body.isRoot)
+            {
+                return 0;
+            }
+
+            var currentOffset = offset;
+
+            // Write joint positions
+            if (settings.UseJointPositions)
+            {
+                switch (m_Body.jointType)
+                {
+                    case ArticulationJointType.RevoluteJoint:
+                    case ArticulationJointType.SphericalJoint:
+                        // All joint positions are angular
+                        for (var dofIndex = 0; dofIndex < m_Body.dofCount; dofIndex++)
+                        {
+                            var jointRotationRads = m_Body.jointPosition[dofIndex];
+                            writer[currentOffset++] = Mathf.Sin(jointRotationRads);
+                            writer[currentOffset++] = Mathf.Cos(jointRotationRads);
+                        }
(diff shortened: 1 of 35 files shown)


com.unity.ml-agents.extensions/Runtime/Sensors/ArticulationBodyJointExtractor.cs
```diff
@@ -0,0 +1,144 @@
...
+            {
+                return 0;
+            }
+
+            var totalCount = 0;
+            if (settings.UseJointPositions)
+            {
+                switch (body.jointType)
+                {
+                    case ArticulationJointType.RevoluteJoint:
```
vincentpierre, line 35: What happens in this case? Is this TODO later?

com.unity.ml-agents.extensions/Runtime/Sensors/PhysicsSensorSettings.cs
```diff
@@ -40,6 +40,10 @@ public struct PhysicsSensorSettings
         /// </summary>
         public bool UseLocalSpaceLinearVelocity;
 
+        public bool UseJointPositions;
```
vincentpierre, line 43: Not really joint positions. UseJointAngles seems more appropriate to me.

chriselion: "positions" is what ArticulationBody uses (even though it refers to translations or angles). How about `UseJointPositionsAndAngles`?

chriselion:
Yeah, probably the robot arm is the best testbed. To keep the number of permutations small, how about runs with
* local-space transforms only
* model-space transforms only
* local + model
* local + model + joint positions/angles
?

com.unity.ml-agents.extensions/Runtime/Sensors/ArticulationBodyJointExtractor.cs
chriselion: That should "fall through" to `totalCount += 2 * body.dofCount` right? Or is my understanding of switch statements obsolete?

I believe (should verify with the physics folks) that:
RevoluteJoint: 1 angular dof
SphericalJoint: 3 angular dof
FixedJoint: 0 dof
PrismaticJoint: 1 linear dof

so both RevoluteJoint and SphericalJoint are basically the same (just different values for body.dofCount)

chriselion:
Experiment results, picking the "middle" run for each variant (judged subjectively via eyeball):
image
So they're more or less equivalent (within noise).

I still have hope that running this on a more complex system like Walker or Crawler will have some benefits.

com.unity.ml-agents.extensions/Runtime/Sensors/ArticulationBodyJointExtractor.cs
surfnerd: fall through is what will happen, maybe a comment to display intent could help

com.unity.ml-agents.extensions/Runtime/Sensors/ArticulationBodyJointExtractor.cs
```diff
@@ -0,0 +1,144 @@
...
+            {
+                return 0;
+            }
+
+            var totalCount = 0;
+            if (settings.UseJointPositionsAndAngles)
+            {
+                switch (body.jointType)
+                {
+                    case ArticulationJointType.RevoluteJoint:
```
surfnerd, line 35:
```suggestion
                    case ArticulationJointType.RevoluteJoint:    // purposely fall through to the SphericalJoint case
```

com.unity.ml-agents.extensions/Runtime/Sensors/ArticulationBodyJointExtractor.cs
```diff
@@ -0,0 +1,144 @@
...
+                    case ArticulationJointType.SphericalJoint:
+                        // Two floats per angular component
+                        totalCount += 2 * body.dofCount;
+                        break;
+                    case ArticulationJointType.FixedJoint:
+                        break; // TODO (none?)
```
surfnerd, line 41: is this a TODO, or is it just a no-op since the joint is fixed?

com.unity.ml-agents.extensions/Runtime/Sensors/IJointExtractor.cs
```diff
@@ -0,0 +1,10 @@
+using Unity.MLAgents.Sensors;
+
+namespace Unity.MLAgents.Extensions.Sensors
+{
+    public interface IJointExtractor
```
surfnerd, line 5: nit: Comments

com.unity.ml-agents.extensions/Runtime/Sensors/ArticulationBodyJointExtractor.cs
chriselion: It's a no-op. updated.

com.unity.ml-agents.extensions/Runtime/Sensors/ArticulationBodyJointExtractor.cs
chriselion: Added some more comments.

PR #4239, opened 2020-07-16
[MLA-427] add pyupgrade to pre-commit and run

chriselion:
### Proposed change(s)
Run pyupgrade in pre-commit. This mostly fixes:
* numbered .format() arguments
* old-style super() calls
* old-style inheritance from `object`

There's another option that will convert `str.format()` to f-strings, but saving that for another PR.

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -39,6 +39,13 @@ repos:
         # flake8-tidy-imports is used for banned-modules, not actually tidying
         additional_dependencies: [flake8-comprehensions==3.2.2, flake8-tidy-imports==4.1.0, flake8-bugbear==20.1.4]
 
+-   repo: https://github.com/asottile/pyupgrade
+    rev: v2.7.0
+    hooks:
+    -   id: pyupgrade
+        args: [--py3-plus]
+        exclude: .*barracuda.py
+
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v2.5.0
     hooks:
--- a/gym-unity/setup.py
+++ b/gym-unity/setup.py
@@ -23,7 +23,7 @@ class VerifyVersionCommand(install):
         tag = os.getenv("CIRCLE_TAG")
 
         if tag != EXPECTED_TAG:
-            info = "Git tag: {0} does not match the expected tag of this app: {1}".format(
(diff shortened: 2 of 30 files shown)


ml-agents-envs/mlagents_envs/environment.py
```diff
@@ -357,10 +357,9 @@ def set_action_for_agent(
         expected_shape = (spec.action_size,)
         if action.shape != expected_shape:
             raise UnityActionException(
-                f"The Agent {0} with BehaviorName {1} needs an input of dimension "
```
chriselion, line 360: This combination of an f-string with numbered .format() exposed a bug in pyupgrade: https://github.com/asottile/pyupgrade/issues/329

gym-unity/setup.py
```diff
@@ -23,7 +23,7 @@ def run(self):
         tag = os.getenv("CIRCLE_TAG")
 
         if tag != EXPECTED_TAG:
-            info = "Git tag: {0} does not match the expected tag of this app: {1}".format(
+            info = "Git tag: {} does not match the expected tag of this app: {}".format(
```
ervteng, line 26: should we convert this to an actual f-string? `f"Git tag:...`

chriselion: I was saving that for the next PR. Adding `--py36-plus` should convert these automatically. The only ones I rewrote were the ones in environment.py (because one of them was broken).

.pre-commit-config.yaml
```diff
@@ -39,6 +39,13 @@ repos:
         # flake8-tidy-imports is used for banned-modules, not actually tidying
         additional_dependencies: [flake8-comprehensions==3.2.2, flake8-tidy-imports==4.1.0, flake8-bugbear==20.1.4]
 
+-   repo: https://github.com/asottile/pyupgrade
+    rev: v2.7.0
+    hooks:
+    -   id: pyupgrade
+        args: [--py3-plus]
```
ervteng, line 46: Should we do `--py36-plus` or would that be too strict?

PR #4240, opened 2020-07-16
Update changelog for .nn checkpoints

chriselion:
### Proposed change(s)
There was no changelog entry for https://github.com/Unity-Technologies/ml-agents/pull/4127

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/com.unity.ml-agents/CHANGELOG.md
+++ b/com.unity.ml-agents/CHANGELOG.md
@@ -11,8 +11,8 @@ and this project adheres to
 ### Major Changes
 
 ### Minor Changes
-Model checkpoints are now also saved as .nn files during training. (#4127)
-Model checkpoint info is saved in TrainingStatus.json after training is concluded (#4127)
+- Model checkpoints are now also saved as .nn files during training. (#4127)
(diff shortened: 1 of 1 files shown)


com.unity.ml-agents/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to
 ### Major Changes
 
 ### Minor Changes
+Model checkpoints are now also saved as .nn files during training. (#4127)
```
harperj, line 14: minor, but in the past I believe we used a bulleted list
```suggestion
* Model checkpoints are now also saved as .nn files during training. (#4127)
```

sankalp04, line 14: +1. I will add to this, mentioning that TrainingStatus.json will output the ML-Agents model checkpoint info.

chriselion: Yup, you're right. Looks like we're using `-` for bullets though.

PR #4248, opened 2020-07-18
[MLA-1172] Avoid calling training_behaviors each step

chriselion:
### Proposed change(s)

Still WIP. Need feedback on the callback vs. interface.

https://jira.unity3d.com/browse/MLA-1172

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/mlagents/trainers/env_manager.py
+++ b/ml-agents/mlagents/trainers/env_manager.py
@@ -78,6 +78,7 @@ class EnvManager(ABC):
     @property
     @abstractmethod
     def training_behaviors(self) -> Dict[BehaviorName, BehaviorSpec]:
+        # TODO change this to a function and make it sound slower
         pass
 
     @abstractmethod
@@ -92,13 +93,14 @@ class EnvManager(ABC):
             self._process_step_infos(self.first_step_infos)
             self.first_step_infos = None
         # Get new policies if found. Always get the latest policy.
-        for brain_name in self.training_behaviors:
+        # TODO rename brain_name
+        for brain_name, agent_manager in self.agent_managers.items():
             _policy = None
             try:
                 # We make sure to empty the policy queue before continuing to produce steps.
                 # This halts the trainers until the policy queue is empty.
                 while True:
-                    _policy = self.agent_managers[brain_name].policy_queue.get_nowait()
+                    _policy = agent_manager.policy_queue.get_nowait()
             except AgentManagerQueue.Empty:
                 if _policy is not None:
                     self.set_policy(brain_name, _policy)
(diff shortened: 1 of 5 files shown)


ml-agents/mlagents/trainers/env_manager.py
```diff
@@ -109,6 +118,18 @@ def advance(self):
         return num_step_infos
 
     def _process_step_infos(self, step_infos: List[EnvironmentStep]) -> int:
+        behavior_ids: Set[str] = set()
+        # TODO nested set comprehension?
+        for step_info in step_infos:
+            behavior_ids |= set(step_info.name_behavior_ids)
```
harperj, line 124: Couldn't we use this to keep `training_behaviors` up to date (cached), and avoid having the callback?

chriselion: That was the approach I wanted to take originally. But at least at the moment, we have to create the trainers (in the trainer_controller) before we get to the StepInfo:
https://github.com/Unity-Technologies/ml-agents/blob/b0be09a0e7cc55a292d63642cfd596e4d2da9c83/ml-agents/mlagents/trainers/trainer_controller.py#L176-L183

harperj: I think moving:
```python
     external_brain_behavior_ids = set(env_manager.training_behaviors.keys()) 
     new_behavior_ids = external_brain_behavior_ids - last_brain_behavior_ids 
     self._create_trainers_and_managers(env_manager, new_behavior_ids) 
     last_brain_behavior_ids = external_brain_behavior_ids 
```

into `TrainerController.advance` after the `env.advance()` call would solve the ordering issue?

harperj: (not against changing the abstraction either, just commenting because I put a little thought into this previously)

chriselion: Yeah, I considered that too. It would mean the first step that you get a new behavior, you won't have a trainer to process them, so you'd end up dropping the StepInfo

https://github.com/Unity-Technologies/ml-agents/blob/6dc68df348380152197bc4add9b6073b35d79145/ml-agents/mlagents/trainers/env_manager.py#L114-L120

(we can silence the warning, but still lose some information).

It's not the end of the world, but the callback seems like the lesser of two evils.

harperj: Ah yeah, I remember this issue.  Seems like the alternative is that the EnvManager has the responsibility for creating the AgentManager, but at the moment the AgentManager depends on the Trainer.

PR #4282, opened 2020-07-29
Pass artifactory url for all pip calls in yamato

chriselion:
### Proposed change(s)
Not sure if there's a good way to DRY this up. It should be redundant eventually when the CI images have a `pip.ini`/`pip.conf` file that make this the default.

### DIFF
--- a/.yamato/protobuf-generation-test.yml
+++ b/.yamato/protobuf-generation-test.yml
@@ -14,8 +14,7 @@ test_mac_protobuf_generation:
       python3 -m venv venv
       . venv/bin/activate
       pip install --upgrade pip --index-url https://artifactory.prd.it.unity3d.com/artifactory/api/pypi/pypi/simple
-      pip install grpcio-tools==1.13.0  --progress-bar=off --index-url https://artifactory.prd.it.unity3d.com/artifactory/api/pypi/pypi/simple
-      pip install mypy-protobuf==1.16.0 --progress-bar=off --index-url https://artifactory.prd.it.unity3d.com/artifactory/api/pypi/pypi/simple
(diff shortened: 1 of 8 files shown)


.yamato/protobuf-generation-test.yml
```diff
@@ -13,9 +13,8 @@ test_mac_protobuf_generation:
       nuget install Grpc.Tools -Version $GRPC_VERSION -OutputDirectory protobuf-definitions/
       python3 -m venv venv
       . venv/bin/activate
-      pip install --upgrade pip
-      pip install grpcio-tools==1.13.0  --progress-bar=off
-      pip install mypy-protobuf==1.16.0 --progress-bar=off
+      pip install --upgrade pip --index-url https://artifactory.prd.it.unity3d.com/artifactory/api/pypi/pypi/simple
+      pip install grpcio==1.28.1 grpcio-tools==1.13.0 protobuf==3.11.3 six==1.14.0 mypy-protobuf==1.16.0  --progress-bar=off --index-url https://artifactory.prd.it.unity3d.com/artifactory/api/pypi/pypi/simple
```
chriselion, line 17: I got the versions here from an older passing run of the proto generation test:
https://yamato.prd.cds.internal.unity3d.com/jobs/497-ml-agents/tree/master/.yamato%252Fprotobuf-generation-test.yml%2523test_mac_protobuf_generation/2066047/job/(log:Execution)

chriselion: Turns out that wasn't the cause of the delta (but still worth doing).

ml-agents-envs/mlagents_envs/communicator_objects/unity_to_external_pb2_grpc.py
```diff
@@ -4,7 +4,7 @@
 from mlagents_envs.communicator_objects import unity_message_pb2 as mlagents__envs_dot_communicator__objects_dot_unity__message__pb2
 
 
-class UnityToExternalProtoStub:
+class UnityToExternalProtoStub(object):
```
chriselion, line 7: Follow up jira: https://jira.unity3d.com/browse/MLA-1215 
We should have run the proto check test when these were modified.

PR #4294, opened 2020-08-01
[examples] Behavior Name correction

ChristianCoenen:
### Proposed change(s)

Stumbled across this one when testing file sizes with the parameter `save_replay_buffer` to `true` for the VisualPushBlock environment. Wondered that a folder named 'VisualHallway' was created and in there was no buffer file.

Occurred because it searched for the 'VisualHallway' config which logically couldn't be found in the VisualPushBlock yaml file. So it used default values.

So the important change is the one in line 870. When I opened the prefab it automatically updated some other stuff. I think this is because some scripts were updated in the last months without opening the prefab to update it as well. So when I opened it, it just updated a bunch of variable names that seems to have changed. I am quite sure its save to commit, but you can also just see this as a bug report with a suggested fix and do it yourself, or update my commit.

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/Project/Assets/ML-Agents/Examples/PushBlock/Prefabs/PushBlockVisualArea.prefab
+++ b/Project/Assets/ML-Agents/Examples/PushBlock/Prefabs/PushBlockVisualArea.prefab
@@ -859,17 +859,18 @@ MonoBehaviour:
   m_Name: 
   m_EditorClassIdentifier: 
   m_BrainParameters:
-    vectorObservationSize: 0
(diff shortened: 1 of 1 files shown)


chriselion:
I did a spot check with `git grep --text m_BehaviorName` and it looks like this is the only prefab like this.

PR #4298, opened 2020-08-03
don't save model twice on exit

chriselion:
### Proposed change(s)
Followup from https://github.com/Unity-Technologies/ml-agents/pull/4127

After that PR, exiting during training would save twice, once when the checkpoint is saved, and once for the "final" model.

This tweaks the logic so that RLTrainer.save_model() doesn't call the model when calling _checkpoint().

https://github.com/Unity-Technologies/ml-agents/pull/4127

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/mlagents/trainers/policy/policy.py
+++ b/ml-agents/mlagents/trainers/policy/policy.py
@@ -152,7 +152,9 @@ class Policy:
         pass
 
     @abstractmethod
-    def checkpoint(self, checkpoint_path: str, settings: SerializationSettings) -> None:
+    def checkpoint(
+        self, checkpoint_path: str, settings: Optional[SerializationSettings]
+    ) -> None:
         pass
 
     @abstractmethod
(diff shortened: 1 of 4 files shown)


dongruoping:
This will result in inconsistency of all checkpoints having both .ckpt and .nn files except the last one only having .ckpt while its .nn file is with a different name under different path. This could work but might cause some confusion when someone look at the result folder or when writing scripts to track the checkpointing in cloud training.

harperj:
Another way to handle this is to have the final model simply checkpoint the same as all of the others, and then copy the `.nn` file to its resulting location.  What do you think about that as an alternative?

chriselion:
Yeah, I like the copying idea better. I'll try to get to it this week unless someone else gets to it first.

PR #4315, opened 2020-08-06
Integrate IActuators into ML-Agents core code.

surfnerd:
### Proposed change(s)

This PR is an attempt to integrate Actuators into the core of ML-Agents code without breaking backward compatibility.  

The biggest architectural change (aside from the Actuator abstraction) is splitting the action buffer between the types of actions that are available for use.  This propagates all the way down to the IPolicy interface and makes for some unflattering hoop-jumping in order to keep Heuristic backward compatible. 

Two follow up PRs will be made:
1. Change environments.
2. Documentation (markdown) changes. 

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


com.unity.ml-agents/Runtime/Agent.deprecated.cs
```diff
@@ -0,0 +1,38 @@
+using System;
```
surfnerd, line 1: Holder for deprecated methods.  Commented them out, for now, otherwise, we can't compile.

com.unity.ml-agents/Runtime/Actuators.meta
```diff
@@ -1,8 +1,3 @@
-fileFormatVersion: 2
-guid: 26733e59183b6479e8f0e892a8bf09a4
-folderAsset: yes
-DefaultImporter:
-  externalObjects: {}
-  userData: 
-  assetBundleName: 
-  assetBundleVariant: 
+﻿fileFormatVersion: 2
```
chriselion, line 1: Is this from different editor versions?

com.unity.ml-agents/Runtime/Agent.cs
```diff
@@ -281,6 +295,24 @@ internal struct AgentParameters
         /// </summary>
         internal VectorSensor collectObservationsSensor;
 
+        /// <summary>
+        /// List of IActuators that this Agent will delegate actions to if any exist.
+        /// </summary>
+        ActuatorManager m_Actuators;
+
+        /// <summary>
+        /// DiscreteVectorActuator which is used by default if no other sensors exist on this Agent. This VectorSensor will
```
chriselion, line 304:
```suggestion
        /// VectorActuator which is used by default if no other sensors exist on this Agent. This VectorSensor will
```
?

com.unity.ml-agents/Runtime/Agent.cs
```diff
@@ -281,6 +295,24 @@ internal struct AgentParameters
...
+        /// delegate its actions to <see cref="IActionReceiver.OnActionReceived"/> by default in order to keep backward compatibility
```
chriselion, line 305: Maybe I need to read more, but it sends to Agent.OnActionReceived, right?

com.unity.ml-agents/Runtime/Agent.cs
```diff
@@ -281,6 +295,24 @@ internal struct AgentParameters
         /// </summary>
         internal VectorSensor collectObservationsSensor;
 
+        /// <summary>
+        /// List of IActuators that this Agent will delegate actions to if any exist.
+        /// </summary>
+        ActuatorManager m_Actuators;
```
chriselion, line 301: m_ActuatorManager?

image

com.unity.ml-agents/Runtime/Agent.cs
```diff
@@ -918,11 +993,11 @@ void SendInfoToBrain()
             {
                 if (m_PolicyFactory.BrainParameters.VectorActionSpaceType == SpaceType.Discrete)
                 {
-                    CollectDiscreteActionMasks(m_ActionMasker);
+                    m_Actuators.WriteActionMask();
```
chriselion, line 996: Not sure we need to it just yet, but we might want to remove the "if(spaceType == Discrete)" around this and make sure it gets handled in WriteActionMask instead.

chriselion: Probably a good sanity check is to see where else in Agent we check the space type for anything. It will probably either need to go away now, or at least soon.

com.unity.ml-agents/Runtime/Agent.cs
```diff
@@ -1049,7 +1133,7 @@ public virtual void CollectDiscreteActionMasks(DiscreteActionMasker actionMasker
         /// three values in the action array to use as the force components. During
         /// training, the agent's  policy learns to set those particular elements of
         /// the array to maximize the training rewards the agent receives. (Of course,
-        /// if you implement a <seealso cref="Heuristic"/> function, it must use the same
+        /// if you implement a <seealso cref="Heuristic(float[])"/> function, it must use the same
```
chriselion, line 1136: cref the (float[], int[]) version?

com.unity.ml-agents/Runtime/Actuators.meta
surfnerd: I'll check and see.

com.unity.ml-agents/Runtime/Agent.cs
surfnerd: yep

com.unity.ml-agents/Runtime/Agent.cs
surfnerd: Yeah, Agent Implements that interface method.

com.unity.ml-agents/Runtime/Agent.cs
surfnerd: makes sense

com.unity.ml-agents/Runtime/Agent.deprecated.cs
chriselion: > we can't compile

Is that because of the warnings-as-errors setting? I thought the obsolete warning was excluded.

com.unity.ml-agents/Runtime/Agent.cs
```diff
@@ -801,22 +830,41 @@ void ResetData()
...
+            switch (m_PolicyFactory.BrainParameters.VectorActionSpaceType)
+            {
+                case SpaceType.Continuous:
+                    #pragma warning disable CS0618
+                    Heuristic(continuousActionsOut);
+                    #pragma warning restore CS0618
+                    Array.Clear(discreteActionsOut, 0, discreteActionsOut.Length);
+                    break;
+                case SpaceType.Discrete:
+                    var convertedOut = Array.ConvertAll(discreteActionsOut, x => (float)x);
```
chriselion, line 859: This will allocate twice, right (once here and once for convertedBackToInt)? Maybe better to keep the int[] around and manually convert?

com.unity.ml-agents/Runtime/Agent.cs
```diff
@@ -48,15 +50,30 @@ internal struct AgentInfo
...
+                start = continuousActions.Length;
+            }
+            if (start >= storedVectorActions.Length)
+            {
+                return;
+            }
+
+            if (discreteActions != null)
+            {
+                Array.Copy(discreteActions, 0, storedVectorActions, start, discreteActions.Length);
```
chriselion, line 74: How does the copy work here for mixed types? Is it a memcpy, or does it actually do the int->float conversion?

chriselion: Never mind, [found it](https://docs.microsoft.com/en-us/dotnet/api/system.array.copy?view=netcore-3.1).
> performs type casting and boxing as required

com.unity.ml-agents/Runtime/Agent.cs
```diff
@@ -1099,12 +1180,34 @@ public virtual void CollectDiscreteActionMasks(DiscreteActionMasker actionMasker
...
+        /// <param name="actions">
+        /// Struct containing the buffers of actions to be executed at this step.
         /// </param>
-        public virtual void OnActionReceived(float[] vectorAction) {}
+        public virtual void OnActionReceived(ActionBuffers actions)
+        {
+            // Copy the actions into our local array and call the original method for
+            // backward compatibility.
+            // For now we need to check which array has the actions in them in order to pass it back to the old method.
+            if (actions.ContinuousActions.Length > 0)
```
chriselion, line 1191: Would it be cleaner and/or easier to test if we made this a utility method of ActionBuffers?

com.unity.ml-agents/Runtime/Agent.deprecated.cs
surfnerd: warnings-as-errors

com.unity.ml-agents/Runtime/Agent.cs
surfnerd: actually, I think I can copy back into the float array. For some reason, I just can't copy from float -> int

surfnerd: sorry, other way around.

com.unity.ml-agents/Runtime/Agent.cs
surfnerd: i wanted to keep it here to keep the context of "this is here strictly for backward compatibility", and I'm willing to move it if you feel strongly about it.

com.unity.ml-agents/Runtime/Actuators/IActionReceiver.cs
```diff
@@ -57,12 +66,40 @@ public override int GetHashCode()
                 return (ContinuousActions.GetHashCode() * 397) ^ DiscreteActions.GetHashCode();
             }
         }
+
+        /// <summary>
+        /// Packs the continuous and discrete actions into one float array.  The array passed into this method
+        /// must have a Length that is greater than or equal to the sum of the Lengths of
```
chriselion, line 72: Should we check this early? Or just rely on Array.Copy() to do the bounds check?

com.unity.ml-agents/Runtime/DiscreteActionMasker.cs
```diff
@@ -15,19 +14,13 @@ namespace Unity.MLAgents
     /// may be illegal. For example, if an agent is adjacent to a wall or other obstacle
     /// you could mask any actions that direct the agent to move into the blocked space.
     /// </remarks>
-    public class DiscreteActionMasker
+    public class DiscreteActionMasker : IDiscreteActionMask
```
chriselion, line 17: Just checking - this isn't considered a breaking change, right? I've seen some chatter about things the look harmless because the old code still compiles, but the assemblies would be incompatible.

com.unity.ml-agents/Runtime/Policies/BarracudaPolicy.cs
```diff
@@ -55,10 +60,21 @@ public void RequestDecision(AgentInfo info, List<ISensor> sensors)
         }
 
         /// <inheritdoc />
-        public float[] DecideAction()
+        public ref readonly ActionBuffers DecideAction()
```
chriselion, line 63: How is this going to work with multiple actuators and mixed action space types? Will we do all the continuous actions, then all the discrete ones?

com.unity.ml-agents/Runtime/Policies/BarracudaPolicy.cs
```diff
@@ -45,6 +49,7 @@ internal class BarracudaPolicy : IPolicy
         {
             var modelRunner = Academy.Instance.GetOrCreateModelRunner(model, brainParameters, inferenceDevice);
             m_ModelRunner = modelRunner;
+            m_SapceType = brainParameters.VectorActionSpaceType;
```
chriselion, line 52: 1) typo: m_SapceType -> m_SpaceType
2) Should we be dealing in ActionSpecs here instead? Would adding a method to BrainParameters to return an equivalent ActionSpec be helpful anywhere else?

com.unity.ml-agents/Runtime/Policies/BarracudaPolicy.cs
chriselion: Never mind, makes more sense after re-reading ExecuteActions

com.unity.ml-agents/Runtime/Policies/BarracudaPolicy.cs
```diff
@@ -55,10 +60,21 @@ public void RequestDecision(AgentInfo info, List<ISensor> sensors)
...
-            return m_ModelRunner?.GetAction(m_AgentId);
+            var actions = m_ModelRunner?.GetAction(m_AgentId);
+            if (m_SapceType == SpaceType.Continuous)
+            {
+                m_LastActionBuffer = new ActionBuffers(actions, Array.Empty<int>());
+                return ref m_LastActionBuffer;
+            }
+
+            // Need to use ConvertAll since you cannot copy a float[] int an int[].
+            m_LastActionBuffer = new ActionBuffers(ActionSegment<float>.Empty, actions == null ? ActionSegment<int>.Empty
```
chriselion, line 74: Move this to a static method e.g. `ActionBuffers.FromDiscreteActions(float[])`?

com.unity.ml-agents/Runtime/Policies/RemotePolicy.cs
```diff
@@ -13,6 +14,8 @@ internal class RemotePolicy : IPolicy
     {
         int m_AgentId;
         string m_FullyQualifiedBehaviorName;
+        SpaceType m_SpaceType;
+        ActionBuffers m_LasActionBuffer;
```
chriselion, line 18:
```suggestion
        ActionBuffers m_LastActionBuffer;
```

com.unity.ml-agents/Runtime/Policies/RemotePolicy.cs
```diff
@@ -34,10 +38,17 @@ public void RequestDecision(AgentInfo info, List<ISensor> sensors)
...
         {
             m_Communicator?.DecideBatch();
-            return m_Communicator?.GetActions(m_FullyQualifiedBehaviorName, m_AgentId);
+            var actions = m_Communicator?.GetActions(m_FullyQualifiedBehaviorName, m_AgentId);
+            if (m_SpaceType == SpaceType.Continuous)
+            {
+                m_LasActionBuffer = new ActionBuffers(actions, Array.Empty<int>());
+                return ref m_LasActionBuffer;
+            }
+            m_LasActionBuffer = new ActionBuffers(Array.Empty<float>(), Array.ConvertAll(actions ?? Array.Empty<float>(), x => (int)x));
```
chriselion, line 50: Same comment about a static factory method.

com.unity.ml-agents/Runtime/Actuators/IActionReceiver.cs
surfnerd: I can add a debug assert.  How does that sound?

com.unity.ml-agents/Runtime/Agent.cs
surfnerd: oh, I misunderstood this comment.  I'll fix it.

com.unity.ml-agents/Runtime/DiscreteActionMasker.cs
surfnerd: good question.  I wonder if we can check the version bump with the validation suite somehow

surfnerd: My intuition would be that since the default behavior didn't change, and the method's signature that uses this type is concretely defined, therefore, it should be ok.  Would be good to know the answer to this question concretely.

com.unity.ml-agents/Runtime/Policies/BarracudaPolicy.cs
surfnerd: Could be used in creating the vector actuator.

com.unity.ml-agents/Runtime/Actuators/IActionReceiver.cs
chriselion: Sounds good.

com.unity.ml-agents/Runtime/Agent.deprecated.cs
chriselion: Hmm... we can save this for later, but should try to get it figured out.

com.unity.ml-agents/Runtime/Agent.deprecated.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+    public partial class Agent
+    {
+        // [Obsolete("CollectDiscreteActionMasks has been deprecated. Please use WriteDiscreteActionMask instead.", false)]
+        public virtual void CollectDiscreteActionMasks(DiscreteActionMasker actionMasker)
+        {
+        }
+
+        // [Obsolete("The Heuristic(float[]) method has been deprecated.  Please use Heuristic(float[], int[]) instead.")]
+        public virtual void Heuristic(float[] continuousActionsOut)
```
chriselion, line 14: Thinking about this a bit more, maybe we should keep the Heuristic behavior the same until we actually have a way to specify continuous and discrete actions in BrainParameters. Even if you use `Heuristic(float[], int[])` you can't actually use both outputs yet.

If you still want to go ahead with it, I think you should keep the old argument name since these aren't necessarily continuous.

surfnerd: Just saw the name change.  Was a mistake.  I'll change it back to `actionsOut`.  

We can keep this method around as long as we like and not obsolete it.  Does that work for you?

chriselion: Sounds good!

PR #4323, opened 2020-08-07
Add Saver

dongruoping:
### Proposed change(s)

Add Saver to handle all save/load/checkpoint/export work.
Main chenge:
- Remove `checkpoint`/`save` functions in policy. `model_path` and `load` option is also removed.
- Remove model loading logics from policy and optimizer
- New BaseSaver and TFSaver class
- Remove model_path and brain_name from `SerializationSettings` and the whole definition is moved to `settings.py`.
- Move `model_serialization.py` under `tf/`

Flow of saving/loading:
1. Each Trainer is initialized with a Saver with its training settings.
2. When calling `add_policy`, register the policy and optimizer to the saver.
3. The saver will store the first policy registered and save/export this model when the training is done.
4. In saver, when registering a policy, it will also initialize/load the model (unless specified not to).
5. During training, Trainer calls the saver to do checkpoint/export.

This PR is separated from [#4296](https://github.com/Unity-Technologies/ml-agents/pull/4296). Same changes for Torch will be added to develop-add-fire.

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


ml-agents/mlagents/trainers/saver/saver.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class BaseSaver(abc.ABC):
+    """This class is the base class for the Saver"""
+
+    def __init__(self):
+        pass
+
+    @abc.abstractmethod
+    def register(self, module: object, init_or_load: bool) -> None:
```
chriselion, line 13:
```suggestion
    def register(self, module: Any, init_or_load: bool) -> None:
```
use `typing.Any` for unknown types. But if there are multiple types that could get registered, I think one method for each, like `register_policy()`, `register_optimizer()` is clearer.

ml-agents/mlagents/trainers/saver/saver.py
```diff
@@ -0,0 +1,39 @@
+# # Unity ML-Agents Toolkit
+import abc
+from mlagents.trainers.policy import Policy
+
+
+class BaseSaver(abc.ABC):
```
chriselion, line 6: I think "BaseSerializer" is a better name for this since it both saves and loads (I know the TF class is called Saver, but it has a bad name).

ml-agents/mlagents/trainers/saver/tf_saver.py
```diff
@@ -0,0 +1,145 @@
...
+        self._keep_checkpoints = trainer_settings.keep_checkpoints
+        self.load = load
+
+        # Currently only support saving one policy. This is the one to be saved.
+        self.policy: Optional[TFPolicy] = None
+
+    def register(self, module, init_or_load=True):
+        if isinstance(module, TFPolicy):
+            if self.policy is None:
+                self.policy = module
```
chriselion, line 37: Should it be an error if `self.policy` is not None? What about if `self.policy != module`?

ml-agents/mlagents/trainers/ppo/trainer.py
```diff
@@ -200,16 +199,17 @@ def create_policy(
...
         )
 
         return policy
 
     def add_policy(
-        self, parsed_behavior_id: BehaviorIdentifiers, policy: Policy
+        self,
+        parsed_behavior_id: BehaviorIdentifiers,
+        policy: Policy,
+        register_saver: bool = True,
```
chriselion, line 212: Is register_saver ever false?

ml-agents/mlagents/trainers/saver/tf_saver.py
```diff
@@ -0,0 +1,145 @@
...
+                self.graph = self.policy.graph
+                self.sess = self.policy.sess
+                with self.graph.as_default():
+                    self.tf_saver = tf.train.Saver(max_to_keep=self._keep_checkpoints)
+            if init_or_load:
+                self.initialize_or_load(module)
```
chriselion, line 43: I think I'd prefer the initialize_or_load() calls to be separate. Either that, or make the abstract method internal (`_initialize_or_load`) and make it clear from the register() docstring that _initialize_or_load will be called.

ml-agents/mlagents/trainers/saver/tf_saver.py
chriselion: Also, should initialize_or_load() be called if self.policy wasn't None? Right now it looks like we could call initialize_or_load on a policy twice.

ml-agents/mlagents/trainers/saver/tf_saver.py
```diff
@@ -0,0 +1,145 @@
...
+        self.load = load
+
+        # Currently only support saving one policy. This is the one to be saved.
+        self.policy: Optional[TFPolicy] = None
+
+    def register(self, module, init_or_load=True):
+        if isinstance(module, TFPolicy):
+            if self.policy is None:
+                self.policy = module
+                self.graph = self.policy.graph
```
chriselion, line 38: nit: initialize self.graph, self.sess, and self.tf_saver to None in `__init__()`

ml-agents/mlagents/trainers/saver/saver.py
dongruoping: I want to remain the flexibility to accept things other than policy and optimizer in case in future we have some new things needed to be saved. I'll change it to `Any`.

ml-agents/mlagents/trainers/ppo/optimizer.py
```diff
@@ -295,6 +295,9 @@ def _create_ppo_optimizer_ops(self):
         self.grads = self.tf_optimizer.compute_gradients(self.loss)
         self.update_batch = self.tf_optimizer.minimize(self.loss)
 
+    def initialize(self):
```
ervteng, line 298: if this method is only called in the __init__ do we need to make it a method? Can we just call `self.policy.initialize()` in the `__init__()`?

ml-agents/mlagents/trainers/saver/saver.py
dongruoping: Hmm but we use the term "Serializing" for model exporting. "Checkpoint" could be a better word here. I chose Saver over Checkpointer just because it sounds better.

ml-agents/mlagents/trainers/saver/saver.py
```diff
@@ -0,0 +1,39 @@
...
+    """This class is the base class for the Saver"""
+
+    def __init__(self):
+        pass
+
+    @abc.abstractmethod
+    def register(self, module: Any, init_or_load: bool) -> None:
```
chriselion, line 13: I still think you should be explicit about the type of thing you're registering; checking the type of an object and branching on that is generally an anti-pattern. 

You can make them non-abstract and do nothing by default, in case one of the implementations doesn't need to do anything (like it appears TFSave ignores optimizers)

ml-agents/mlagents/trainers/saver/saver.py
chriselion: "serialize" generally means converting between object instances and persistent formats. This also loads, so I disagree that Saver sounds better.

ml-agents/mlagents/trainers/saver/tf_saver.py
dongruoping: To the first response, it's not an error. This line correspond to the warning in `rl_trainer.py`: `Trainer has multiple policies, but default behavior only saves the first.`. There might be cases we have multiple policies and others are just ignored for checkpointing. I can put some message here like `Registering multiple policies while the default behavior only saves the first. This model will not be saved.`

dongruoping: It shouldn't be called twice. During add_policy, the policy is registered to the saver. At the same time, call initialize_or_load on that policy. Can you point me out where could it be called initialize_or_load twice?

ml-agents/mlagents/trainers/ppo/trainer.py
dongruoping: this is for unittest

ml-agents/mlagents/trainers/saver/tf_saver.py
ervteng: Hmm, there's one catch with this logic that is a bit ugly. `initialize_or_load` really should only be called once the Optimizer is created/registered, since only at that time is the entire graph created and ready for loading. I think it works in this case since both the register calls happen in add_policy after the optimizer is created. Maybe we can get away with an `if isinstance(TFOptimizer)` call `_initialize_or_load()`? Or move it to a separate call in the trainer (i.e. called after both `register` calls happen).

dongruoping: Can do that. Then in trainer.py it'll be like 
```
self.saver.register(self.policy)
self.saver.initialize_or_load(self.policy)
self.saver.register(self.optimizer)
```
(looks a bit weird not loading the optimizer but it's fine)

`initialize_or_load` is made a public function for ghost trainer. In that case, this function is called explicitly without registering.

ml-agents/mlagents/trainers/saver/saver.py
ervteng: What if we have separate `register_policy` and `register_optimizer` methods? That should make the typing explicit.

The branching on `TFPolicy` is just for TensorFlow since the policy object holds the graph and sess. In PyTorch we can treat the registered classes in the same way. In that world we could make an interface type (`SerializableModel` or similar) and set the typing to that here.

dongruoping:
@andrewcoh can you help me check if this code is doing correctly with ghost trainer? (especially the init_or_load part)

ml-agents/mlagents/trainers/saver/saver.py
dongruoping: The awkwardness here might come from the fact that this is designed to work for both TF and Torch which are doing checkpointing in very different way. In Torch, everything that has a `.get_modules()` function can be registered here and optimizer is not skipped. 

Will make it non-abstract.

ml-agents/mlagents/trainers/saver/saver.py
dongruoping: By better I mean "saver" and "checkpointer". The problem is, the term "serialize" has been saved for the use of exporting model to onnx/barracuda by us (model_serialization.py does **tf graph -> frozen_graph.pb -> .nn/.onnx**). Here the saver is more referring to checkpointint tf (**save/load .ckpt files**) and calls external export function for exporting when needed. So I would separate these two and not using the term serialize for the whole saver.

Or I can use serialize here and modify model_serialization.py to other names like model_export.py

ml-agents/mlagents/trainers/ppo/optimizer.py
dongruoping: legacy code when fixing tests. removed.

ml-agents/mlagents/trainers/saver/tf_saver.py
dongruoping: done.

ml-agents/mlagents/trainers/saver/saver.py
dongruoping: So do we want this function to have fixed input of only accepting policy and optimizer?
This is totally fine for tensorflow, and it's ok for current pytorch.
If we're not going to add other type of objects in pytorch I can change this to `register_policy` and `register_optimizer`.

ml-agents/mlagents/trainers/saver/tf_saver.py
ervteng: You wouldn't need the `self.policy` in `initialize_or_load` right? Since the saver object has already saved the policy object internally when `register(self.policy)` was called.

dongruoping: `initialize_or_load` takes the module as input argument.

This is mainly for ghost trainer where the policy in GhostTrainer is not part of the actual trainer (and thus not registered) but need to be initialized.

ml-agents/mlagents/trainers/saver/saver.py
dongruoping: Add specified type as Union[Policy, Optimizer] for this. This should clear the concern of unclear type.

ml-agents/mlagents/trainers/saver/tf_saver.py
dongruoping: Modified this. Now it takes an optional input argument `policy`. Initialize the input policy if specified, otherwise initialize self.policy.
Now it should look better.

ml-agents/mlagents/trainers/saver/saver.py
chriselion: OK, I'm fine with the name as it is.

ml-agents/mlagents/trainers/ppo/trainer.py
chriselion: Is is just to avoid extra work during the tests? Can we use a mock/patch there instead, so that we don't need to complicate the interface here?

ml-agents/mlagents/trainers/saver/saver.py
```diff
@@ -0,0 +1,44 @@
...
+        :param output_filepath: path (without suffix) for the model file(s)
+        :param brain_name: Brain name of brain to be trained.
+        """
+        pass
+
+    @abc.abstractmethod
+    def initialize_or_load(self, policy):
+        """
+        If there is an initialize path, load from that. Else, load from the set model path.
+        If load is set to True, don't reset steps to 0. Else, do. This allows a user to,
```
chriselion, line 41: Please update the docstring explaining the effect of passing the policy.
This also refers to initialize and model paths, and load, which are all part of the TFSaver implementation, but not the base class.

ml-agents/mlagents/trainers/saver/saver.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class BaseSaver(abc.ABC):
+    """This class is the base class for the Saver"""
+
+    def __init__(self):
+        pass
+
+    @abc.abstractmethod
+    def register(self, module: Any) -> None:
```
chriselion, line 13: Please add a docstring

ml-agents/mlagents/trainers/saver/tf_saver.py
chriselion: > Can you point me out where could it be called initialize_or_load twice?

This looks OK now. In the original version that I commented on, the `initialize_or_load` call was outside of the `if self.policy is None` block

ml-agents/mlagents/trainers/saver/saver.py
ervteng: To make it a bit cleaner, what if we do (in the TF implementation):
```
def register(self, module: Union[Policy, Optimizer]):
    if isinstance(module, Optimizer):
        self._register_optimizer(module)
    elif isinstance(module, Policy):
        self._register_optimizer(module)
    else: 
        # throw some error
```
The two helper methods can have Optimizer and Policy input types, and are probably not needed in the Torch implementation.

ml-agents/mlagents/trainers/saver/tf_saver.py
ervteng: Ah got it. Seems good if @andrewcoh agrees w.r.t. the ghost trainer

ml-agents/mlagents/trainers/saver/saver.py
dongruoping: Yes as this is probably not needed in the Torch, I'll put it in TFSaver.

ml-agents/mlagents/trainers/ppo/trainer.py
dongruoping: No it actually breaks some tests. For example, `test_trainer_increment_step` in `test_ppo.py`, we create a mock policy and called `trainer.add_policy(behavior_id, policy_mock)`. This will trigger the trainer to register it to Saver and the Saver will try to access its .graph and .sess and do `with self.policy.graph.as_default(): ...` which resulted in error. 
Creating an empty tf.Graph() and attach to the mock doesn't work either. TF will complain "No variables to save".

ml-agents/mlagents/trainers/saver/saver.py
dongruoping: done

ml-agents/mlagents/trainers/saver/saver.py
dongruoping: done

ml-agents/mlagents/trainers/ppo/trainer.py
chriselion: Right, that's why you can use mocks so that the saver doesn't do any registration.

In your example, changing the test_trainer_increment_step to 
```
@mock.patch.object(RLTrainer, "create_saver")
@mock.patch("mlagents.trainers.ppo.trainer.PPOOptimizer")
def test_trainer_increment_step(ppo_optimizer, mock_create_saver):
```
will fix the problem. This will cause RLTrainer.create_saver to return a Mock object, so the calls to register won't have any side effects (unless you configure the mock to have a side effect).

You can also do things like
```
    mock_saver = mock.Mock()
    mock_create_saver.return_value = mock_saver
    ...
    mock_saver.register.assert_called()
    mock_saver.initialize_or_load.assert_called()
```
if you want to sanity-check the logic in the test.

If there's a legitimate case for non-test code to not register, then it's OK to keep the flag. But if *only* tests use it, I think you need to remove it and switch the tests to use Mocks.

dongruoping: That solved the problem, thanks. Removed the redundant input argument.

ml-agents/mlagents/trainers/saver/saver.py
```diff
@@ -0,0 +1,65 @@
...
+from typing import Any
+
+
+class BaseSaver(abc.ABC):
+    """This class is the base class for the Saver"""
+
+    def __init__(self):
+        pass
+
+    def register(self, module: Any) -> None:
```
ervteng, line 12: Should `register` be an abstract method as well?

ml-agents/mlagents/trainers/saver/saver.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        Register the modules to the Saver.
+        The Saver will store the module and include it in the saved files
+        when saving checkpoint/exporting graph.
+        :param module: the module to be registered
+        """
+        pass
+
+    def _register_policy(self, policy):
```
ervteng, line 21: Should we add typing to policy and optimizer for these two (I'm guessing the generic `Policy` and `Optimizer`).

dongruoping: definition in tf_saver will have typing conflicts with super class

PR #4334, opened 2020-08-11
Adding rank to ml-agents

anupambhatnagar:
### Proposed change(s)

Added util functions to access rank in ml-agents. This will be used in the distributed training package.

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/mlagents/tf_utils/globals.py
+++ b/ml-agents/mlagents/tf_utils/globals.py
@@ -5,7 +5,7 @@ _rank: Optional[int] = None
 
 def get_rank() -> Optional[int]:
     """
-    Returns the rank (in the MPI sense) of the current process.
+    Returns the rank (in the MPI sense) of the current node.
     For local training, this will always be None.
     If this needs to be used, it should be done from outside ml-agents.
     :return:
--- a/ml-agents/mlagents/tf_utils/globals.py
+++ b/ml-agents/mlagents/tf_utils/globals.py
@@ -1,8 +1,13 @@
 from typing import Optional
 
 _rank: Optional[int] = None
-_local_rank: Optional[int] = None
 
 
 def get_rank() -> Optional[int]:
+    """
+    Returns the rank (in the MPI sense) of the current process.
+    For local training, this will always be None.
+    If this needs to be used, it should be done from outside ml-agents.
+    :return:
+    """
     return _rank
--- a/ml-agents/mlagents/tf_utils/globals.py
+++ b/ml-agents/mlagents/tf_utils/globals.py
@@ -6,7 +6,3 @@ _local_rank: Optional[int] = None
 
 def get_rank() -> Optional[int]:
     return _rank
-
-
-def broadcast_variables() -> bool:
-    return True if _rank is not None else False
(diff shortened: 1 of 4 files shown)


ml-agents/mlagents/tf_utils/globals.py
```diff
@@ -0,0 +1,13 @@
+try:
+    import horovod.tensorflow as hvd
```
chriselion, line 2: I don't think we should try to import horovod anywhere. Something like

``` python
_rank: Optional[int] = None
_local_rank: Optional[int] = None

def get_rank() -> Optional[int]:
  return _rank
```
and maybe
``` python
def set_rank(new_rank: Optional[int]) -> None:
  if _rank is not None and new_rank is not None:
    raise RuntimeError("Rank has already been set")
  _rank = new_rank
```
We can call set_rank (or just assign _rank directly) either from the distributed code, or provide an initializer via plugin.

chriselion:
Aren't there some other places you need to check the rank, like generate_session_config() and TFPolicy.save_model()

ml-agents/mlagents/trainers/stats.py
```diff
@@ -84,32 +85,51 @@ def __init__(self):
...
-                "Reward: {:0.3f}"
-                ". Std of Reward: {:0.3f}. {}".format(
-                    category,
-                    step,
-                    time.time() - self.training_start_time,
-                    stats_summary.mean,
-                    stats_summary.std,
-                    is_training,
+            if self.rank is not None:
+                logger.info(
```
chriselion, line 102: See https://github.com/Unity-Technologies/ml-agents/pull/4336 to reduce the amount of redundant code here.

ml-agents/mlagents/tf_utils/globals.py
anupambhatnagar: we don't need a set rank as horovod will assign rank to the workers and that should not be overridden

ml-agents/mlagents/trainers/stats.py
anupambhatnagar: changed

ml-agents/mlagents/trainers/policy/tf_policy.py
```diff
@@ -255,6 +257,9 @@ def initialize_or_load(self):
             self._load_graph(self.model_path, reset_global_steps=reset_steps)
         else:
             self._initialize_graph()
+        # broadcast initial weights from worker-0
+        if broadcast_variables():
+            self.sess.run(hvd.broadcast_global_variables(0))
```
anupambhatnagar, line 262: @chriselion do you have a workaround to avoid the hvd call here? One way would be to subclass the TFPolicy, but I'm trying to avoid that.

chriselion: How about something like this?
```python
from typing import Callable

class TFPolicy(Policy):
  ...
  broadcast_global_variables: Callable[[int], None] = lambda x: None
  ...
  def initialize_or_load(self):
    ...
    TFPolicy.broadcast_global_variables(0)
```

Then when we do the distributed setup, you can set 
```python
TFPolicy.broadcast_global_variables = hvd.broadcast_global_variables
```
(so it basically acts as a "function pointer" with a no-op default)

anupambhatnagar: thanks, I'll try this.

PR #4367, opened 2020-08-17
update demonstration meta files

chriselion:
### Proposed change(s)
These keep getting updated locally. Looks like their path is wrong.


surfnerd:
thank goodness

PR #4371, opened 2020-08-17
[MLA-1267] Account for actuators in training and inference.

chriselion:
### Proposed change(s)
* Use ActionSpec instead of BrainParameters for training and inference.
* Update BarracudaParameterLoader to use ActuatorComponents
* Update the Basic scene to use an actuator

I added ActionSpec.CheckNotHybrid() to validate that only of continuous or discrete is non-zero. We should eventually remove this when we allow hybrid action spaces, but for now, I tried to use it whenever I needed to assume it's one or the other.

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


com.unity.ml-agents/Runtime/Actuators/ActuatorManager.cs
```diff
@@ -99,10 +101,40 @@ internal void ReadyActuatorsForExecution(IList<IActuator> actuators, int numCont
             var discreteActions = numDiscreteBranches == 0 ? ActionSegment<int>.Empty : new ActionSegment<int>(new int[numDiscreteBranches]);
 
             StoredActions = new ActionBuffers(continuousActions, discreteActions);
-            m_DiscreteActionMask = new ActuatorDiscreteActionMask(actuators, sumOfDiscreteBranches, numDiscreteBranches);
+            m_CombinedActionSpec = CombineActionSpecs(actuators);
+            m_DiscreteActionMask = new ActuatorDiscreteActionMask(actuators, sumOfDiscreteBranches, numDiscreteBranches, m_CombinedActionSpec.BranchSizes);
             m_ReadyForExecution = true;
         }
 
+        internal static ActionSpec CombineActionSpecs(IList<IActuator> actuators)
```
chriselion, line 109: I originally had it as a static method like in `ActionSpec` like `static ActionSpec Combine(IList<ActionSpec>)`, but not sure if there's a good way of going from the List of Actuators in ReadyActuatorsForExecution() to the list of ActionSpecs

com.unity.ml-agents/Runtime/Actuators/ActuatorDiscreteActionMask.cs
```diff
@@ -27,11 +27,12 @@ internal class ActuatorDiscreteActionMask : IDiscreteActionMask
         /// </summary>
         public int CurrentBranchOffset { get; set; }
 
-        internal ActuatorDiscreteActionMask(IList<IActuator> actuators, int sumOfDiscreteBranchSizes, int numBranches)
+        internal ActuatorDiscreteActionMask(IList<IActuator> actuators, int sumOfDiscreteBranchSizes, int numBranches, int[] branchSizes = null)
```
chriselion, line 30: I can make branchSizes required; that would mean changing some tests, but we can remove code from LazyInitialize.

com.unity.ml-agents/Runtime/Actuators/ActionSpec.cs
```diff
@@ -57,19 +57,32 @@ public static ActionSpec MakeContinuous(int numActions)
         /// <param name="branchSizes">The array of branch sizes for the discrete action space.  Each index
         /// contains the number of actions available for that branch.</param>
         /// <returns>An Discrete ActionSpec initialized with the array of branch sizes.</returns>
-        public static ActionSpec MakeDiscrete(int[] branchSizes)
+        public static ActionSpec MakeDiscrete(params int[] branchSizes)
```
chriselion, line 60: Just a bit syntactic sugar so you can say `MakeDiscrete(1, 2, 3)` instead of `MakeDiscrete(new[] {1, 2, 3})`. I can revert if you're not a fan.

Project/Assets/ML-Agents/Examples/Basic/Prefabs/Basic.prefab
```diff
@@ -472,6 +477,49 @@ MonoBehaviour:
   m_Script: {fileID: 11500000, guid: 3a6da8f78a394c6ab027688eab81e04d, type: 3}
   m_Name: 
   m_EditorClassIdentifier: 
+  debugCommandLineOverride: 
+--- !u!114 &748290217173149218
```
chriselion, line 481: Moved Agent and BasicSensor to prefab, added BasicActuator

com.unity.ml-agents/Runtime/Agent.cs
```diff
@@ -412,7 +412,13 @@ public void LazyInitialize()
             Academy.Instance.DecideAction += DecideAction;
             Academy.Instance.AgentAct += AgentStep;
             Academy.Instance.AgentForceReset += _AgentReset;
-            m_Brain = m_PolicyFactory.GeneratePolicy(Heuristic);
+
+            using (TimerStack.Instance.Scoped("InitializeActuators"))
```
chriselion, line 416: Need to do this before we set up the Policy

com.unity.ml-agents/Runtime/Actuators/ActuatorDiscreteActionMask.cs
surfnerd: i wanted to avoid the allocation if the mask is never used.

com.unity.ml-agents/Runtime/Policies/BrainParameters.cs
```diff
@@ -1,6 +1,7 @@
 using System;
 using UnityEngine;
 using UnityEngine.Serialization;
+using Unity.MLAgents.Actuators;
```
surfnerd, line 4: needed?

com.unity.ml-agents/Runtime/Actuators/ActuatorDiscreteActionMask.cs
chriselion: Ah, I see. I should at least be able to change CombineActionSpecs() to use Array.Empy if there are no discrete actions, but I think we always need to do the aggregation

com.unity.ml-agents/Runtime/Policies/BrainParameters.cs
chriselion: Oops, I originally had some BrainParameters->ActionSpec code here, but didn't need it.

PR #4374, opened 2020-08-17
[DO NOT MERGE] New memory abstraction and AMRL implementation

ervteng:
### Proposed change(s)

New memory layer abstraction that lets a memory define it's memory size (w.r.t. the "memory size" specified in the config). Then the ActorCritic has a `memory_size` property which the Policy uses to set it's internal `m_size`.  This allows the actual `m_size` that is used during training to be different than what is specified in the YAML. We might want to do this, for instance, if we want `memory_size: 64` to mean the Policy has a memory size of 64, but the Critic network has memory size of 64 (total of 128) - so the Policy object can store both during training, but only export 64 for inference. 

Also an implementation of https://openreview.net/forum?id=Bkl7bREtDr using this abstraction. 

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


ervteng:
Outdated - `develop-amrl` has the latest experimental code.

PR #4399, opened 2020-08-20
Grid Sensor

J-Travnik:
### Proposed change(s)

The Grid Sensor combines the generality of data extraction from Raycasts with the image processing power of Convolutional Neural Networks. The Grid Sensor can be used to collect data in the general form of a "Width x Height x Channel" matrix which can be used for training Reinforcement Learning agents or for data analysis.

### Motivation

In MLAgents there are 2 main sensors for observing information that is "physically" around the agent. 

**Raycasts**

Raycasts provide the agent the ability to see things along prespecified lines of sight, similar to LIDAR. The kind of data it can extract is open to the developer from things like:
* The type of an object (enemy, npc, etc)
* The health of a unit
* the damage-per-second of a weapon on the ground

This is simple to implement and provides enough information for most simple games. When few are used, they are computationally fast. However, there are multiple limiting factors:
* The rays need to be at the same height as the things the agent should observe
* Objects can remain hidden by line of sight and if the knowledge of those objects is crucial to the success of the agent, then this limitation must be compensated for by the agents networks capacity (i.e., need a bigger brain with memory)
* The order of the raycasts (one raycast being to the left/right of another) is thrown away at the model level and must be learned by the agent which extends training time. Multiple raycasts exacerbates this issue.
* Typically the length of the raycasts is limited because the agent need not know about objects that are at the other side of the level. Combined with few raycasts for computational efficiency, this means that an agent may not observe objects that fall between these rays and the issue becomes worse as the objects reduce in size.

**Camera**

The Camera provides the agent with either a grayscale or an RGB image of the game environment. It goes without saying that there non-linear relationships between nearby pixels in an image. It is this intuition that helps form the basis of Convolutional Neural Networks (CNNs) and established the literature of designing networks that take advantage of these relationships between pixels. Following this established literature of CNNs on image based data, the MLAgent's Camera Sensor provides a means by which the agent can include high dimensional inputs (images) into its observation stream.
However the Camera Sensor has its own drawbacks as well.
* It requires render the scene and thus is computationally slower than alternatives that do not use rendering
* It has yet been shown that the Camera Sensor can be used on a headless machine which means it is not yet possible (if at all) to train an agent on a headless infrastructure.
* If the textures of the important objects in the game are updated, the agent needs to be retrained.
* The RGB of the camera only provides a maximum of 3 channels to the agent.

These limitations provided the motivation towards the development of the Grid Sensor and Grid Observations as described below.

### Contribution

An image can be thought of as a matrix of a predefined width (W) and a height (H) and each pixel can be thought of as simply an array of length 3 (in the case of RGB), `[Red, Green, Blue]` holding the different channel information of the color (channel) intensities at that pixel location. Thus an image is just a 3 dimensional matrix of size WxHx3. A Grid Observation can be thought of as a generalization of this setup where in place of a pixel there is a "cell" which is an array of length N representing different channel intensities at that cell position. 
From a Convolutional Neural Network point of view, the introduction of multiple channels in an "image" isn't a new concept. In fact the original inspiration for the Grid Sensor came from [MinAtar](https://arxiv.org/abs/1903.03176) which introduced a small suite of environments analogous to the [Atari Learning Environment](https://arxiv.org/abs/1207.4708) but where the representations were 10x10xn binary state representations. The distinction of Grid Observations is what the data within the channels represents. Instead of limiting the channels to color intensities, the channels within a cell of a Grid Observation generalize to any data that can be represented by a single number (float or int) such as the type of object within a cell or the value of a certain property.

Additionally, this PR also modifies the rpc_utils.py script to accept multiple pngs as was demonstrated [in this unity hack week](https://github.com/Unity-Technologies/ml-agents/compare/hw20-segmentation#diff-45a39226f4c34fcfc49904fa1bc3aaeaR55).

See docs/Grid-Sensor.md for further documentation.

### Types of change(s)

- [x] New feature
- [x] Code refactor
- [x] Documentation update

### Checklist
- [x] Updated the [changelog](https://github.com/Unity-Technologies/ml-agents/blob/master/com.unity.ml-agents.extensions/CHANGELOG.md) 

### Other comments

The Grid Sensor was developed collaboration between [Eidos Montreal](https://eidosmontreal.com/en#!) and [Matsuko](https://www.matsuko.com/).

**Developers**

- Jaden Travnik
- Charles Pearson
- Martin Certicky
- Erik Gajdos
- Romain Trachel
- Alexandre Peyrot


ml-agents-envs/mlagents_envs/rpc_utils.py
```diff
@@ -43,24 +46,54 @@ def behavior_spec_from_proto(
...
-    with hierarchical_timer("image_decompress"):
-        image_bytearray = bytearray(image_bytes)
-        image = Image.open(io.BytesIO(image_bytearray))
-        # Normally Image loads lazily, this forces it to do loading in the timer scope.
-        image.load()
-    s = np.array(image, dtype=np.float32) / 255.0
-    if gray_scale:
+    image_bytearray = bytearray(image_bytes)
+
+    if expected_channels == 1:
```
chriselion, line 59: @harperj Do you mind taking a look at this part? I made some changes to your hackweek code, because I was worried that `split()` might produce too many pieces in an edge case.

I'm still not sure the "right" way to do this; see https://stackoverflow.com/questions/63585016/determining-the-size-of-a-png-from-raw-bytes

ml-agents-envs/mlagents_envs/rpc_utils.py
```diff
@@ -43,24 +46,57 @@ def behavior_spec_from_proto(
...
+        buffer = io.BytesIO(image_bytearray[bytes_read:])
+        with hierarchical_timer("image_decompress"):
+            image = Image.open(buffer)
+            image.load()
+        image_arrays.append(np.array(image, dtype=np.float32) / 255.0)
+
+        # Look for the next header, starting from the current stream location
+        try:
+            offset = buffer.getvalue().index(PNG_HEADER, buffer.tell())
+            bytes_read += offset
```
harperj, line 87: Given this code, I don't think we could ever hit the `while` condition.

ml-agents-envs/mlagents_envs/rpc_utils.py
```diff
@@ -43,24 +46,57 @@ def behavior_spec_from_proto(
...
+        except ValueError:
+            # Didn't find the header, so must be at the end.
+            break
+
+    img = np.concatenate(image_arrays, axis=2)
+    # We can drop additional channels since they may need to be added to include
+    # numbers of observation channels not divisible by 3.
+    actual_channels = list(img.shape)[2]
+    if actual_channels > expected_channels:
```
harperj, line 96: What if the actual channels are less than the expected channels?

ml-agents-envs/mlagents_envs/rpc_utils.py
chriselion: Good point, I'll change that to a `while True:`

ml-agents-envs/mlagents_envs/rpc_utils.py
chriselion: There's a `if list(obs.shape) != list(img.shape):` check in the calling code (line 125 below). I can add something similar here if you want.

DevProject/ProjectSettings/ProjectSettings.asset
```diff
@@ -103,15 +103,21 @@ PlayerSettings:
   xboxOneMonoLoggingLevel: 0
   xboxOneLoggingLevel: 1
   xboxOneDisableEsram: 0
+  xboxOneEnableTypeOptimization: 0
```
chriselion, line 106: TODO revert

Project/ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +1 @@
-m_EditorVersion: 2018.4.24f1
+m_EditorVersion: 2018.4.22f1
```
chriselion, line 1: TODO revert

com.unity.ml-agents.extensions/CHANGELOG.md
```diff
@@ -5,4 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
- * Initial version
\ No newline at end of file
+ * Initial version
```
chriselion, line 8: TODO move to main changelog

docs/Grid-Sensor.md
```diff
@@ -0,0 +1,223 @@
+# Summary
```
chriselion, line 1: TODO: move to com.unity.ml-agents.extensions/Documentation~

com.unity.ml-agents.extensions/CHANGELOG.md
chriselion: Will do after merging PR, to avoid conflicts

ml-agents-envs/mlagents_envs/rpc_utils.py
harperj: LGTM

ml-agents-envs/mlagents_envs/rpc_utils.py
harperj: 👍 forgot about that check.  I don't think we gain much by adding another check.

ml-agents-envs/mlagents_envs/tests/test_rpc_utils.py
```diff
@@ -176,16 +190,28 @@ def proto_from_steps_and_action(
...
+def test_process_pixels_multi_png():
+    height = 128
+    width = 64
+    num_channels = 7
+    in_array = np.random.rand(height, width, num_channels)
+    byte_arr = generate_compressed_data(in_array)
+    out_array = process_pixels(byte_arr, num_channels)
+    assert out_array.shape == (height, width, num_channels)
+    assert np.sum(in_array - out_array) / np.prod(in_array.shape) < 0.01
+    assert (in_array - out_array < 0.01).all()
```
harperj, line 208: Very minor, but numpy has a method for this:
```python
np.allclose(in_array, out_array)
```

chriselion: Copied from other tests, but I'll give that a try.

chriselion: Actually, all the other tests were kinda wonky because they weren't taking the absolute value of the difference :/  Update them all to use `np.allclose` with the same tolerance.

chriselion:
Note: cancelled the CircleCI tests since their equivalents are running on github actions.

chriselion:
Will merge as soon as yamato tests pass on https://github.com/Unity-Technologies/ml-agents/pull/4409

PR #4403, opened 2020-08-21
Add torch_utils class, auto-detect CUDA availability 

ervteng:
### Proposed change(s)

This PR adds a torch_utils class (similiar to tf_utils) and requires importing `torch` from there. This lets us do a couple things, the first being detect if CUDA is available and set the default tensor type appropriately. Requiring importing `torch` from here ensures that this will be set before any torch functions are used at all. The PR also adds an `is_available()` method to `torch_utils`, and throws a nicer error if torch isn't available when you passed `--torch`. 

In the future we can also set the number of torch threads, etc. from here. 

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


.pylintrc
```diff
@@ -2,6 +2,8 @@
 # Add files or directories to the ignore list. They should be base names, not
 # paths.
 ignore=CVS
+generated-members=torch.*
```
ervteng, line 5: I'm pretty sure this isn't the best way to do this, but without it, pylint will complain about `torch` not having the right members everywhere. I believe this is because `torch` _could_ be None, though it never actually happens.

setup.cfg
```diff
@@ -21,3 +21,4 @@ ignore =
 
 banned-modules = tensorflow = use mlagents.tf_utils instead (it handles tf2 compat).
                  logging = use mlagents_envs.logging_util instead
+                 torch = use mlagents.torch_utils istead (handles GPU detection).
```
chriselion, line 24:
```suggestion
                 torch = use mlagents.torch_utils instead (handles GPU detection).
```

dongruoping:
Only tensors are created on GPU but the models are still on CPU.
Should it also be addressed in this PR?

dongruoping:
Another point missing here is that when running Torch with CuDNN, if we want to get completely reproducible results, we'll also need to set these besides setting seeds for torch and numpy:
```
torch.backends.cudnn.deterministic = True
torch.backends.cudnn.benchmark = False
```
while these could potentially hurt the running performance.

ervteng:
> Another point missing here is that when running Torch with CuDNN, if we want to get completely reproducible results, we'll also need to set these besides setting seeds for torch and numpy:
> 
> ```
> torch.backends.cudnn.deterministic = True
> torch.backends.cudnn.benchmark = False
> ```
> 
> while these could potentially hurt the running performance.

How much of a performance hit do we expect by setting these flags?

dongruoping:
> How much of a performance hit do we expect by setting these flags?
Not sure, it could be case by case, depending on how much the model changes during training and how much optimization cudnn does.
I haven't really try to quantify the impact, but it's a general warning on PyTorch doc.

dongruoping:
`ml-agents/mlagents/trainers/policy/torch_policy.py: line 61` should be removed. It reset the default tensor type which is properly set in utils to non-cuda.

PR #4404, opened 2020-08-21
Fix NNCheckpointManager for Torch

dongruoping:
### Proposed change(s)

Currently NNCheckpointManager only tracks .nn files while in PyTorch the exported file is .onnx.
Fix it to track .onnx files when using PyTorch.

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


dongruoping:
I wonder if we should also change the name `NNCheckpointManager` since it's not for .nn only anymore

ervteng:
> I wonder if we should also change the name `NNCheckpointManager` since it's not for .nn only anymore

They're still neural nets 😛 

But yeah, we can name it maybe just CheckpointManager, or PolicyCheckpointManager.

PR #4446, opened 2020-09-01
Refactor rank

dongruoping:
### Proposed change(s)

Refactor rank added in #4334.

Originally rank logics and files are added under tensorflow files so it won't work fo PyTorch, while this is actually not a framework-specific utility.
Move `global.py` out of tensorflow path and move the `broadcast_global_variables` into `TrainerController.py` to make it work for both framework.

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


chriselion:
I'll defer to @anupambhatnagar for the right behavior here. For the purposes of integrating with Horovod, it looks like there are separate tf and torch rank methods, and there doesn't appear to torch equivalent of `broadcast_global_variables`

anupambhatnagar:
@dongruoping  Rank is required only for distributed training which has been built keeping tensorflow usage in mind. At this point I'm unable to test this with pytorch + horovod so I suggest it stays where it is.

dongruoping:
> @dongruoping Rank is required only for distributed training which has been built keeping tensorflow usage in mind. At this point I'm unable to test this with pytorch + horovod so I suggest it stays where it is.

Got it

PR #4454, opened 2020-09-04
don't run pytest on doc or C# changes

chriselion:
### Proposed change(s)
Try to avoid running pytest on e.g. C#-only or doc changes. Any changes in these folders will still trigger the changes (and it always runs on master).

Alternatively, we could do this by `**.py` and `**.demo` files, but I wasn't sure if there were any other extensions we use in the tests.

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -2,6 +2,10 @@ name: pytest
 
 on:
   pull_request:
+    paths:  # This action will only run if the PR modifies a file in one of these directories
+    - 'ml-agents/**'
+    - 'ml-agents-envs/**'
+    - 'gym-unity/**'
   push:
     branches: [master]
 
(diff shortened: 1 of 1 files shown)


chriselion:
Made a python change on https://github.com/Unity-Technologies/ml-agents/pull/4455 and confirmed that tests are triggered (and they aren't on this PR)

.github/workflows/pytest.yml
```diff
@@ -2,6 +2,12 @@ name: pytest
 
 on:
   pull_request:
+    paths:  # This action will only run if the PR modifies a file in one of these directories
```
andrewcoh, line 5: this is awesome

andrewcoh (approved):

PR #4468, opened 2020-09-10
Update changelog before branching for release 7

chriselion:
### Proposed change(s)
Update the previous Unreleased section to 1.4.0, and add a new empty one.

### Types of change(s)
- [x] Documentation update

### Checklist

### DIFF
--- a/com.unity.ml-agents/CHANGELOG.md
+++ b/com.unity.ml-agents/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to
 #### com.unity.ml-agents (C#)
 #### ml-agents / ml-agents-envs / gym-unity (Python)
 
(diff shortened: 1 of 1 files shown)


hvpeteet (approved):
LGTM (not much to say on such a small change)

Nit / non-blocking question: do we have a markdown formatter / style guide?

chriselion:
> do we have a markdown formatter

Yes, his name is @mmattar :)

We tried prettier before https://prettier.io/blog/2017/11/07/1.8.0.html#markdown-support but it made some bad changes (i.e. not the same visual result) so didn't go far with it. If you've got one you recommend, I'd be willing to try it (ideally one that integrates with pre-commit, but we can make it work if it doesn't integrate out of the box).

PR #4471, opened 2020-09-11
[bug-fix] Set number of threads based on allocated CPU count in Docker containers

ervteng:
### Proposed change(s)

PyTorch by default sets number of threads to 1/2 available cores. However, in Docker containers, the built-in Python methods return the number of cores of the host machine, not the container. This PR reads the allocated CPU count from `cgroup` and uses that to set the num threads. 

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/mlagents/torch_utils/cpu_utils.py
+++ b/ml-agents/mlagents/torch_utils/cpu_utils.py
@@ -9,11 +9,11 @@ def get_num_threads_to_use() -> Optional[int]:
     need, but for smaller machines, we'd like to scale to less than that.
     By default, PyTorch uses 1/2 of the available cores.
     """
-    num_cpus = _get_num_cpus()
+    num_cpus = _get_num_available_cpus()
     return max(min(num_cpus // 2, 4), 1) if num_cpus is not None else None
 
 
-def _get_num_cpus() -> Optional[int]:
+def _get_num_available_cpus() -> Optional[int]:
     """
     Returns number of CPUs using cgroups if possible. This accounts
     for Docker containers that are limited in cores.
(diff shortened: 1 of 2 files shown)


chriselion:
Nice catch. I've run into this before and didn't think about it for pytorch.

ml-agents/mlagents/torch_utils/cpu_utils.py
```diff
@@ -0,0 +1,34 @@
...
+    if period > 0 and quota > 0:
+        return int(quota // period)
+    else:
+        return os.cpu_count()
+
+
+def _read_in_integer_file(filename: str) -> int:
+    try:
+        with open(filename) as f:
+            return int(f.readlines()[0])
```
chriselion, line 32: There's some code on https://bugs.python.org/issue36054 - they use the equivalent of `int( f.read().rstrip() )`

vincentpierre (approved):

ml-agents/mlagents/torch_utils/cpu_utils.py
ervteng: I like the `read().rstrip()` better that what I had, changed

chriselion (approved):

ml-agents/mlagents/torch_utils/cpu_utils.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    Gets the number of threads to use. For most problems, 4 is all you
+    need, but for smaller machines, we'd like to scale to less than that.
+    By default, PyTorch uses 1/2 of the available cores.
+    """
+    num_cpus = _get_num_cpus()
+    return max(min(num_cpus // 2, 4), 1) if num_cpus is not None else None
+
+
+def _get_num_cpus() -> Optional[int]:
```
chriselion, line 16: nit: maybe `_get_num_available_cpus()`?

chriselion (approved):

PR #4476, opened 2020-09-11
Add stacking to multi-dimensional and compressed observations

dongruoping:
### Proposed change(s)

Add observations stacking to multi-dimensional and compressed observations.
This allows stacking for visual observations and other multi-dimensional observations.

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


com.unity.ml-agents/Runtime/Sensors/CameraSensorComponent.cs
```diff
@@ -75,6 +75,11 @@ public bool Grayscale
             set { m_Grayscale = value; }
         }
 
+        [HideInInspector, SerializeField, FormerlySerializedAs("observationStacks")]
```
chriselion, line 78: nit: you shouldn't need the FormerlySerializedAs for these; that's only if an existing variable in a serialized object got renamed.

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
```diff
@@ -67,11 +63,24 @@ public StackingSensor(ISensor wrapped, int numStackedObservations)
             }
 
             // TODO support arbitrary stacking dimension
-            m_Shape[0] *= numStackedObservations;
+            m_Shape[m_Shape.Length - 1] *= numStackedObservations;
             m_StackedObservations = new float[numStackedObservations][];
-            for (var i = 0; i < numStackedObservations; i++)
+            m_StackedCompressedObservations = new byte[numStackedObservations][];
```
chriselion, line 68: nit: move this inside the `else` block (so that we don't allocate it if we don't need it). And same thing with m_StackedObservations in the `if` block

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
```diff
@@ -128,17 +147,42 @@ public string GetName()
...
+            var compressed = m_WrappedSensor.GetCompressedObservation();
+            m_StackedCompressedObservations[m_CurrentIndex] = compressed;
+
+            int bytesLength = 0;
+            foreach (byte[] compressedObs in m_StackedCompressedObservations)
+            {
+                bytesLength += compressedObs.Length;
+            }
+
+            byte[] bytes = new byte[bytesLength];
```
chriselion, line 161: nit: `bytesOut` or `outputBytes` or `concatenatedObservations`

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
```diff
@@ -67,11 +63,24 @@ public StackingSensor(ISensor wrapped, int numStackedObservations)
             }
 
             // TODO support arbitrary stacking dimension
-            m_Shape[0] *= numStackedObservations;
+            m_Shape[m_Shape.Length - 1] *= numStackedObservations;
             m_StackedObservations = new float[numStackedObservations][];
-            for (var i = 0; i < numStackedObservations; i++)
+            m_StackedCompressedObservations = new byte[numStackedObservations][];
+            m_EmptyPNG = CreateEmptyPNG();
```
chriselion, line 69: Move inside the `else` block, and check the compression type. I'm surprised this isn't breaking tests, since we'd be accessing the shape out of bounds for a vector sensor.

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
```diff
@@ -128,17 +147,42 @@ public string GetName()
...
+        public SensorCompressionType GetCompressionType()
         {
-            return SensorCompressionType.None;
+            return m_WrappedSensor.GetCompressionType();
         }
 
-        // TODO support stacked compressed observations (byte stream)
+        public byte[] CreateEmptyPNG()
+        {
+            int height = m_WrappedSensor.GetObservationShape()[0];
```
chriselion, line 182: This is fine for now, but it would be even better if we could write a PNG directly instead of creating the empty Texture. That can be a later optimization though.

dongruoping: I was thinking of doing that but didn't find a way to directly create an empty PNG file. Do you have any solutions in mind?

chriselion: No, I looked into it a bit and it might be non-trivial. It can definitely happen later.

com.unity.ml-agents/Runtime/Sensors/CameraSensor.cs
```diff
@@ -53,6 +54,18 @@ public SensorCompressionType CompressionType
             m_Name = name;
             m_Shape = GenerateShape(width, height, grayscale);
             m_CompressionType = compression;
+
+            if (m_CompressionType != SensorCompressionType.None)
+            {
+                if (m_Grayscale)
+                {
+                    m_CompressionMapping = new int[] { 1, 1, 1 };
```
chriselion, line 62: Should these be  {0, 0, 0} and {0, 1, 2}?

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
```diff
@@ -67,11 +63,40 @@ public StackingSensor(ISensor wrapped, int numStackedObservations)
...
+            {
+                m_StackedCompressedObservations = new byte[numStackedObservations][];
+                m_EmptyCompressedObservation = CreateEmptyPNG();
+                for (var i = 0; i < numStackedObservations; i++)
+                {
+                    m_StackedCompressedObservations[i] = m_EmptyCompressedObservation;
+                }
+
+                // Construct compression mapping
+                var compressibleSensor = m_WrappedSensor as ICompressibleSensor;
```
chriselion, line 86: You need to handle the case where the wrapped sensor isn't actually an ICompressibleSensor, just an ISensor. I think in that case, you should either leave the mapping empty, or the "identity" mapping (`mapping[i] == i`)

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
```diff
@@ -67,11 +63,40 @@ public StackingSensor(ISensor wrapped, int numStackedObservations)
...
+                int wrappedNumChannel = shape[2];
+                int wrappedMapLength = wrappedMapping.Length;
+                int offset;
+                m_CompressionMapping = new int[wrappedMapLength * m_NumStackedObservations];
+                for (var i = 0; i < numStackedObservations; i++)
+                {
+                    offset = wrappedNumChannel * i;
+                    for (var j = 0; j < wrappedMapLength; j++)
+                    {
+                        m_CompressionMapping[j + wrappedMapLength * i] = wrappedMapping[j] > 0 ? wrappedMapping[j] + offset : 0;
```
chriselion, line 97: Use -1 if the wrappedMapping is negative?

protobuf-definitions/proto/mlagents_envs/communicator_objects/observation.proto
```diff
@@ -19,4 +19,5 @@ message ObservationProto {
         bytes compressed_data = 3;
         FloatData float_data = 4;
     }
+    repeated int32 mapping = 5;
```
chriselion, line 22: Please rename to something more descriptive, such as `compressed_channel_mapping`.

I think you also need to document the format of this in more detail, either here or in the GetCompressionMapping() docstring.

dongruoping: Plan to put it in GetCompressionMapping()

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
dongruoping: This is inside the else statement of checking `m_WrappedSensor.GetCompressionType()` which should have a compression type other than None. My intention is to make all sensors that can have compressed observation being ICompressibleSensor, so it should always be ICompressibleSensor here.
Or should I check and raise an exception here if the sensor has `m_WrappedSensor.GetCompressionType() !=SensorCompressionType.None` but its' not an ICompressibleSensor?

com.unity.ml-agents/Runtime/Sensors/CameraSensor.cs
dongruoping: 0 is currently reserved for channels to be dropped and the mapping index start from 1. I'm thinking about changing it to -1 as drop and starting from 0 like you said.

ml-agents-envs/mlagents_envs/tests/test_rpc_utils.py
```diff
@@ -85,6 +85,7 @@ def generate_compressed_data(in_array: np.ndarray) -> bytes:
 def generate_compressed_proto_obs(in_array: np.ndarray) -> ObservationProto:
     obs_proto = ObservationProto()
     obs_proto.compressed_data = generate_compressed_data(in_array)
+    obs_proto.mapping.extend(list(range(len(in_array))))
```
chriselion, line 88: This needs some more unit tests. I think you should cover
* 1 channel, no mapping (old case)
* 3 channels, no mapping (old case)
* >3 channels, no mapping (old case)
* 1 channel, mapping
* 3 channel, mapping
* >3 channels with a non-trivial mapping (e.g. 4 channels and stacking=2)

com.unity.ml-agents/Runtime/Sensors/CameraSensor.cs
chriselion: Ah, I see, you're using 1-index in the python code. I think 0-indexing is clearer though.

ml-agents-envs/mlagents_envs/tests/test_rpc_utils.py
dongruoping: Sure I'll add those.
But are we allowing compressed observation with 1 channel? I thought it should always be uncompressed.

chriselion: > compressed observation with 1 channel

Grayscale visual observations, and those can be stacked.

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
chriselion: You shouldn't assume in general that `m_WrappedSensor.GetCompressionType() !=SensorCompressionType.None` means that it's an `ICompressibleSensor`. I think maybe a better name would change how we think about this - how about something like `ISparseChannelSensor`?

> My intention is to make all sensors that can have compressed observation being ICompressibleSensor, so it should always be ICompressibleSensor here.

I actually think this might make things harder for backwards compatibility reasons. I'll start another thread on that.

chriselion:
You're also going to need to add some safeguards to make sure that different versions of C# and python can work OK with each other, or give an error early when then can't. To do this, you'll need to increase the communication protocol to 1.2.0 and add a flag in the Capabilities proto message - see https://github.com/Unity-Technologies/ml-agents/pull/4462 for an example.

There's no problem using a new version of python with old C#, but if you have the latest C# using a non-trivial mapping and it's connect to an old version of the python code, you'll need to raise a warning and fall back to uncompressed observations, since the python code won't be able to decompress them correctly. Because of that, you might want consider _not_ using the ICompressibleSensor for Camera and RenderTexture sensor, since it will be harder to detect when it's OK to use compression.

com.unity.ml-agents/Runtime/Sensors/CameraSensor.cs
dongruoping: changed to 0-index

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
```diff
@@ -282,6 +284,19 @@ public override ISensor CreateSensor()
             this.DiffNumSideZX = (GridNumSideZ - GridNumSideX);
             this.OffsetGridNumSide = (GridNumSideZ - 1f) / 2f;
             this.DebugColors = debugColors;
+
+            if (GetCompressionType() != SensorCompressionType.None)
+            {
+                var totalCannel = 0;
```
chriselion, line 290: `totalChannels`?

com.unity.ml-agents/Runtime/Communicator/GrpcExtensions.cs
```diff
@@ -237,19 +237,32 @@ public static ObservationProto GetObservationProto(this ISensor sensor, Observat
             var shape = sensor.GetObservationShape();
             ObservationProto observationProto = null;
             var compressionType = sensor.GetCompressionType();
-            // Check capabilities if we need to concatenate PNGs
-            if (compressionType == SensorCompressionType.PNG && shape.Length == 3 && shape[2] > 3)
+            if (!s_HaveWarnedAboutTrainerCapabilities)
```
chriselion, line 240: This is incorrect - previously it would "downgrade" to uncompressed observations every time; now it does it the first time but not subsequent times, so training will break.

chriselion: Probably best to rename this and add a new flag for the compression mapping warning.

com.unity.ml-agents/Runtime/Communicator/GrpcExtensions.cs
```diff
@@ -237,19 +237,32 @@ public static ObservationProto GetObservationProto(this ISensor sensor, Observat
...
-                    compressionType = SensorCompressionType.None;
                 }
+
+                // Check capabilities if we need mapping for compessed observations
+                if (compressionType != SensorCompressionType.None)
+                {
+                    var canHandleMapping = Academy.Instance.TrainerCapabilities == null || Academy.Instance.TrainerCapabilities.CompressionMappings;
+                    if (!canHandleMapping)
+                    {
+                        Debug.LogWarning("Attached trainer doesn't support compression mapping. " +
```
chriselion, line 259: A warning is neither sufficient nor always necessary here. You should warn if the trainer can't handle it (no "potentially") and fall back to uncompressed observations, otherwise you shouldn't warn and should keep using compression.

The question is what's a mapping that the trainer can't handle. I think the ones that it can handle are:
1) null mapping (either not an `ICompressibleSensor` or `GetCompressionMapping()` returns null)
2) grayscale: mapping is all 0's
3) identity mapping: mapping[i] == i for all i

You should check for these cases here (or in a helper method) and not warn/downgrade if they're true.

com.unity.ml-agents/Runtime/Communicator/GrpcExtensions.cs
```diff
@@ -274,6 +287,7 @@ public static ObservationProto GetObservationProto(this ISensor sensor, Observat
             }
             else
             {
+                var compressibleSensor = sensor as ICompressibleSensor;
```
chriselion, line 290: As mentioned above, you cannot assume that `sensor` is an `ICompressibleSensor`.

Also, the proto code will throw if you pass it a null pointer, so you need to check that before setting.

protobuf-definitions/proto/mlagents_envs/communicator_objects/capabilities.proto
```diff
@@ -13,4 +13,7 @@ message UnityRLCapabilitiesProto {
 
     // concatenated PNG files for compressed visual observations with >3 channels.
     bool concatenatedPngObservations = 2;
+
+    // compression mapping for stacking compressed observations.
+    bool compressionMappings = 3;
```
chriselion, line 18: I guess we're not consistent about the name case ([protobuf style guide](https://developers.google.com/protocol-buffers/docs/style#message_and_field_names) says `snake_case` for field names). But we can at least be consistent for the variable names. Please `compressed_channel_mapping` or `compressedChannelMapping`.

ml-agents-envs/mlagents_envs/environment.py
```diff
@@ -59,7 +59,8 @@ class UnityEnvironment(BaseEnv):
     # Revision history:
     #  * 1.0.0 - initial version
     #  * 1.1.0 - support concatenated PNGs for compressed observations.
-    API_VERSION = "1.1.0"
+    #  * 1.2.0 - support compression mapping for stacked compressed observations.
+    API_VERSION = "1.2.0"
```
chriselion, line 63: You'll need to update this method 
https://github.com/Unity-Technologies/ml-agents/blob/master/ml-agents-envs/mlagents_envs/environment.py#L117-L121
to make sure the flag is set.

com.unity.ml-agents/Runtime/Sensors/ICompressibleSensor.cs
```diff
@@ -0,0 +1,21 @@
+namespace Unity.MLAgents.Sensors
+{
+    /// <summary>
+    /// Sensor interface for generating observations.
+    /// </summary>
+    public interface ICompressibleSensor : ISensor
```
chriselion, line 6: I still think we need a different name for this, since sensors that don't implement the interface can still be compressible. I suggested `ISparseChannelSensor` in another comment; I don't love that but I like it better than the current form.

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
chriselion: This still needs to be addressed - if wrapping a non-ICompressibleSensor, assume an identity mapping.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
chriselion: I think you can just use `ObservationPerCell` as the total number of channels.

com.unity.ml-agents/Runtime/Sensors/ICompressibleSensor.cs
```diff
@@ -0,0 +1,21 @@
...
+    {
+        /// <summary>
+        /// Returns the mapping of the channels in compressed data to the
+        /// actual channel after decompression.
+        /// The mapping is a list of interger index with the same length as
```
chriselion, line 11: I think this comment is inaccurate - for the GridSensor it has 8 channels but the mapping length is 12.

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
```diff
@@ -67,11 +63,40 @@ public StackingSensor(ISensor wrapped, int numStackedObservations)
...
+                int wrappedNumChannel = shape[2];
+                int wrappedMapLength = wrappedMapping.Length;
+                int offset;
+                m_CompressionMapping = new int[wrappedMapLength * m_NumStackedObservations];
+                for (var i = 0; i < numStackedObservations; i++)
+                {
+                    offset = wrappedNumChannel * i;
+                    for (var j = 0; j < wrappedMapLength; j++)
+                    {
+                        m_CompressionMapping[j + wrappedMapLength * i] = wrappedMapping[j] >= 0 ? wrappedMapping[j] + offset : -1;
```
chriselion, line 97: I think it should be the responsibility of the StackingSensor to pad the wrapped sensor's mapping to a multiple of 3, not the wrapped sensor.

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
```diff
@@ -67,11 +63,40 @@ public StackingSensor(ISensor wrapped, int numStackedObservations)
...
+                }
+
+                // Construct compression mapping
+                var compressibleSensor = m_WrappedSensor as ICompressibleSensor;
+                var wrappedMapping = compressibleSensor.GetCompressionMapping();
+                int wrappedNumChannel = shape[2];
+                int wrappedMapLength = wrappedMapping.Length;
+                int offset;
+                m_CompressionMapping = new int[wrappedMapLength * m_NumStackedObservations];
+                for (var i = 0; i < numStackedObservations; i++)
```
chriselion, line 92: You need to add some C# unit tests to make sure the mapping results in what you expect.

com.unity.ml-agents/Runtime/Sensors/ICompressibleSensor.cs
```diff
@@ -0,0 +1,21 @@
...
+        /// the number of layers (channels) of the sensor.
+        /// Each index indicates the actual channel the layer will go into.
+        /// Layers with the same index will be averaged, and layers with negative index will be dropped.
+        /// For example, mapping for CameraSensor using grayscale and stacking of two: [0, 0, 0, 1, 1, 1]
+        /// Mapping for GridSensor of 4 channels and stacking of two: [0, 1, 2, 3, -1, -1, 4, 5, 6, 7, -1, -1]
+        /// </summary>
+        /// <returns>Mapping of the compressed data</returns>
+        int[] GetCompressionMapping();
```
chriselion, line 19: GetCompressedChannelMapping? (similar to the proto field name).

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
dongruoping: If we don't do this in GridSensor we cannot use GridSensor on its own without stacking right?

chriselion: Not sure I understand. GridSensor works fine without stacking now.

If the wrapped sensor has a compression mapping of size N, the stacking sensor should have a one of size 
`(numStacks) * 3 * ((N + 2) / 3)`.

dongruoping: Yes GridSensor is fine now without mapping. But I mean the wrapped sensor should return the padded mapping otherwise it will be problematic without stacking (N itself should be multiple of 3). 
Because when using the sensor on its own, GetCompressionMapping() should return a mapping that's the same size as the compressed observation, which is padded. 
Also the stacked compressed observation is constructed by stacking the padded observation from wrapped sensor, and the mapping should correspond with that. 

But I agree it would be better to double check this in StackingSensor.

chriselion: > But I mean the wrapped sensor should return the padded mapping otherwise it will be problematic without stacking (N itself should be multiple of 3).

I disagree. If there are only 4 channels in the observation, then `[0, 1, 2, 3]` is an OK mapping for the wrapped sensor. It's up to the StackingSensor to pad this to `[0, 1, 2, 3, -1, -1, 4, 5, 6, 7, -1, -1 ...]`

dongruoping: The mapping is not only used by stacking sensor but can also be sent to python directly. In that case GridSensor will send [0, 1, 2, 3] to python with an observation of 6 channels which can be very confusing of whether it's an error or the rest should be dropped.

dongruoping: After second thoughts I decided to not send the mapping at all for GridSensor/CameraSensor and have the python code handle that in the old way.

My original idea is to make them all have the same interface so that they can all be handled in a clean and simple way, but if we're allowing sensors that doesn't follow this interface then it's becoming uglier than just keep it the old way and handle them differently.

By doing this I'm keeping the strong assumption that channel=1 is always equivalent to a mapping of {0, 0, 0} and all others has an identical mapping (non-stacking case).

Please let me know if there's still any backward-compat problem.

com.unity.ml-agents/Runtime/Communicator/GrpcExtensions.cs
```diff
@@ -310,7 +333,29 @@ public static UnityRLCapabilitiesProto ToProto(this UnityRLCapabilities rlCaps)
...
+            if (compressibleSensor is null)
+            {
+                return true;
+            }
+            var mapping = compressibleSensor.GetCompressedChannelMapping();
+            if (mapping == null)
+            {
+                return true;
+            }
+            var zeroMapping = new int[] { 0, 0, 0 };
```
chriselion, line 352: nit: can you just check these with a loop? This will allocate memory each step, which will impact performance.

I also think you can only treat {0,0,0} as trivial if numChannels==1, right?

com.unity.ml-agents/Runtime/Sensors/RenderTextureSensorComponent.cs
```diff
@@ -78,8 +97,15 @@ public override int[] GetObservationShape()
         {
             var width = RenderTexture != null ? RenderTexture.width : 0;
             var height = RenderTexture != null ? RenderTexture.height : 0;
+            var observationShape = new[] { height, width, Grayscale ? 1 : 3 };
+
+            var stacks = ObservationStacks > 1 ? ObservationStacks : 1;
+            if (stacks > 1)
+            {
+                observationShape[observationShape.Length - 1] *= stacks;
```
chriselion, line 105: nit: you can use `observationShape[2]` - no need to generalize since you make observationShape above.

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
```diff
@@ -128,17 +146,88 @@ public string GetName()
...
+        }
+
+        public int[] ConstructStackedCompressedChannelMapping(ISensor wrappedSenesor)
+        {
+            int[] wrappedMapping;
+            int wrappedNumChannel = wrappedSenesor.GetObservationShape()[2];
+            var compressibleSensor = m_WrappedSensor as ISparseChannelSensor;
+            if (compressibleSensor != null)
+            {
+                wrappedMapping = compressibleSensor.GetCompressedChannelMapping();
```
chriselion, line 202: Do you need to handle GetCompressedChannelMapping returning null?

I think you could do something like
```
if (compressibleSensor != null)
{
    wrappedMapping = compressibleSensor.GetCompressedChannelMapping();
}
if(wrappedMapping = null)
{
    if (wrappedNumChannel == 1) {...}
    else {}
}
```

ml-agents-envs/mlagents_envs/rpc_utils.py
```diff
@@ -147,7 +161,14 @@ def observation_to_np_array(
...
     else:
-        img = process_pixels(obs.compressed_data, expected_channels)
+        if "compressed_channel_mapping" in (f.name for f in obs.DESCRIPTOR.fields):
+            img = process_pixels(
+                obs.compressed_data,
+                expected_channels,
+                list(obs.compressed_channel_mapping),
+            )
+        else:
+            img = process_pixels(obs.compressed_data, expected_channels)
```
vincentpierre, line 171: I think you can do

```
compressed_channel_mapping = list(obs.compressed_channel_mapping) if "compressed_channel_mapping" in (f.name for f in obs.DESCRIPTOR.fields) else None
img = process_pixels(
                obs.compressed_data,
                expected_channels,
                lcompressed_channel_mapping,
            )
```
If it makes it clearer.

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
chriselion: Yeah, this sounds like a better approach.

ml-agents-envs/mlagents_envs/rpc_utils.py
chriselion: I'm not sure that checking for it in the fields is safe. What's the problem with just calling 
```
img = process_pixels(
  obs.compressed_data,
  expected_channels,
  obs.compressed_channel_mapping,
)
```
? If it's just the mypy type, there should be an interface that matches (maybe `Sequence`?).

ml-agents-envs/mlagents_envs/rpc_utils.py
```diff
@@ -116,12 +107,35 @@ def process_pixels(image_bytes: bytes, expected_channels: int) -> np.ndarray:
...
             break
 
-    img = np.concatenate(image_arrays, axis=2)
-    # We can drop additional channels since they may need to be added to include
-    # numbers of observation channels not divisible by 3.
-    actual_channels = list(img.shape)[2]
-    if actual_channels > expected_channels:
-        img = img[..., 0:expected_channels]
+    if mappings is not None and len(mappings) > 0:
+        image_arrays = np.concatenate(image_arrays, axis=2).transpose((2, 0, 1))
```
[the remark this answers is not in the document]
chriselion: Can you do something like
```
mappings = list(range(expected_channels)) if expected_channels > 0 else [0, 0, 0]
```
so that the "old case" is handled by the new code?

com.unity.ml-agents/Runtime/Communicator/GrpcExtensions.cs
dongruoping: {0,0,0} can only have numChannels==1 no?

chriselion: Hmm... I could write a ISparseChannelSensor that returns {0,0,0} for it's mapping but has shape[2] = 2. Not sure what would happen on the trainer side - maybe all 0s? It's probably another error condition that we need to check for somewhere.

Related, we should also have a check that the mapping values are less than the number of channels. For example, {0, 1, 99} with numChannels == 3 should be an error (and will probably raise an exception on python).

dongruoping: On python side, if there's a mapping sent it always checks if the mapping is the same length as the input observation channels, and raises an exception if it's not. So in your case it will raise exception. Since mapping is designed for non-trivial case I don't think it's a good idea to allow partial mapping here.

Thanks for remind, will add the check for mapping values less than the number of channels.

ml-agents-envs/mlagents_envs/rpc_utils.py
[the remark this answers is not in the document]
dongruoping: Doesn't that get into the same performance debate for C# that a mapping has to be created every time?

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
```diff
@@ -128,17 +146,88 @@ public string GetName()
...
+        public byte[] CreateEmptyPNG()
+        {
+            int height = m_WrappedSensor.GetObservationShape()[0];
+            int width = m_WrappedSensor.GetObservationShape()[1];
+            var texture2D = new Texture2D(width, height, TextureFormat.RGB24, false);
+            return texture2D.EncodeToPNG();
+        }
+
+        public int[] ConstructStackedCompressedChannelMapping(ISensor wrappedSenesor)
+        {
```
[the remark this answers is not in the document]
dongruoping: Added some comments. Let me know if anything still seems unclear.

ml-agents-envs/mlagents_envs/rpc_utils.py
dongruoping: Wasn't sure if it's safe to access a unset field but it seems fine.

ml-agents-envs/mlagents_envs/rpc_utils.py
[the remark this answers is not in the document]
chriselion: Fair point :) I'm not sure if memory allocation/GC in python is better or worse than in C#, but it's at least tolerated in python.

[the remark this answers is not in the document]
chriselion: Looks like you could use `range(expected_channels)` (although you'd have to make the mypy types a bit more generic) which should allocate less, and you could reuse `(0, 0, 0)`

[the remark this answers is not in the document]
dongruoping: I've split them into two functions now so it looks clear how it process with mapping or with expected_channels. If unifying them still looks better than this I can change it.

com.unity.ml-agents/CHANGELOG.md
```diff
@@ -13,11 +13,19 @@ and this project adheres to
  - Added the Random Network Distillation (RND) intrinsic reward signal to the Pytorch
  trainers. To use RND, add a `rnd` section to the `reward_signals` section of your
  yaml configuration file. [More information here](https://github.com/Unity-Technologies/ml-agents/blob/master/docs/Training-Configuration-File.md#rnd-intrinsic-reward)
-
 ### Minor Changes
 #### com.unity.ml-agents (C#)
+ - Stacking for compressed observations is now supported. An addtional setting
+ option `Observation Stacks` is added in editor to sensor components that support
+ compressed observations. An new class `ISparseChannelSensor` with an
```
chriselion, line 20:
```suggestion
 compressed observations. A new class `ISparseChannelSensor` with an
```

chriselion (commented):
Just a few small changes for what's there. But you're missing some unit test coverage on the new functionality. You can see what's covered here (this is from the latest CI run):
https://yamato-artifactviewer.prd.cds.internal.unity3d.com/10cef3c6-1a04-4bc2-9e8c-9ee059861eef%2Flogs%2Fupm-ci~%2Ftest-results%2Fisolation-3638200-com.unity.ml-agents%2FReport/index.htm

In particular, I think you need to:
1) Make a test that creates a dummy sensor that writes compressed observations, put a stacking sensor around that, and make sure the output bytes are expected. The dummy sensor doesn't need to output valid PNGs, you can just return {i} for the i'th step.
2) Make a test that has a non-multiple-of-3 number of channels, so that the -1 padding code will be tested. Again make a dummy sensor class that returns something like [3, 2, 1, 0] for the mapping and make sure the StackingSensor mapping is what you expect.
3) Add some test cases for IsTrivialMapping from GrpcExtensions

ml-agents-envs/mlagents_envs/rpc_utils.py
```diff
@@ -116,13 +107,55 @@ def process_pixels(image_bytes: bytes, expected_channels: int) -> np.ndarray:
...
+
+def process_images_mapping(image_arrays, mappings):
+    image_arrays = np.concatenate(image_arrays, axis=2).transpose((2, 0, 1))
+
+    if len(mappings) != len(image_arrays):
+        raise UnityObservationException(
+            f"Compressed observation and its mapping had different number of channels - "
+            f"observation had {len(image_arrays)} channels but its mapping had {len(mappings)} channels"
+        )
+    if len({m for m in mappings if m > -1}) != max(mappings) + 1:
```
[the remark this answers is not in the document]
dongruoping: In theory if there's only one grayscale image its mapping should not be [2, 2, 2]. It's defined as the mapping of the channel in png bytes and the actual channel after decompression so it should always start from 0. 
It also checks for cases like [0, 1, 199].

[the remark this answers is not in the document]
chriselion: Oh right, the entries are the destination channel. Sounds good.

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
```diff
@@ -85,11 +99,32 @@ public int Write(ObservationWriter writer)
 
             // Now write the saved observations (oldest first)
             var numWritten = 0;
-            for (var i = 0; i < m_NumStackedObservations; i++)
+            if (wrappedShape.Length == 1)
```
dongruoping, line 102: Found this bug that I didn't change the Write method properly for uncompressed obs and it's still stacking on the first dimension. Here changed it to the last dimension for 2D observations so that inference can work properly.

I left 1D unchanged (stacking on first dim) since it has to be handled differently than 2D anyway and otherwise all models have to be re-trained. Thoughts?

com.unity.ml-agents/Runtime/Sensors/ISparseChannelSensor.cs
```diff
@@ -0,0 +1,20 @@
+namespace Unity.MLAgents.Sensors
+{
+    /// <summary>
+    /// Sensor interface for generating observations.
```
chriselion, line 4: Can you update the description here? I think it's copied from ISensor

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
```diff
@@ -48,48 +61,74 @@ public StackingSensor(ISensor wrapped, int numStackedObservations)
...
+            if (m_WrappedSensor.GetCompressionType() != SensorCompressionType.None)
+            {
+                m_StackedCompressedObservations = new byte[numStackedObservations][];
+                m_EmptyCompressedObservation = CreateEmptyPNG();
+                for (var i = 0; i < numStackedObservations; i++)
+                {
+                    m_StackedCompressedObservations[i] = m_EmptyCompressedObservation;
+                }
+                m_CompressionMapping = ConstructStackedCompressedChannelMapping(wrapped);
+                m_tensorShape = new TensorShape(0, m_WrappedShape[0], m_WrappedShape[1], m_WrappedShape[2]);
```
chriselion, line 93: This needs to be set depending on `m_Shape.Length` - you can have uncompressed visual observations.

com.unity.ml-agents/Tests/Editor/Sensor/StackingSensorTests.cs
```diff
@@ -60,5 +63,147 @@ public void TestStackingReset()
...
+            // Test RGB stacked mapping with RenderTextureSensor
+            var renderTextureSensor = new RenderTextureSensor(new RenderTexture(24, 16, 0),
+                false, "renderTexture", SensorCompressionType.PNG);
+            var stackedRenderTextureSensor = new StackingSensor(renderTextureSensor, 2);
+            Assert.AreEqual(stackedRenderTextureSensor.GetCompressedChannelMapping(), new[] { 0, 1, 2, 3, 4, 5 });
+
+            // Test mapping with dummy layers that should be dropped
+            var dummySensor = new Dummy3DSensor();
+            dummySensor.Shape = new int[] { 2, 2, 4 };
+            dummySensor.Mapping = new int[] { 0, 1, 2, 3, -1, -1 };
```
chriselion, line 146: The wrapped sensor shouldn't be adding -1's to its mapping, right? I think we decided this was the responsibility of the StackingSensor. The assert below can stay the same, though.

chriselion: Actually, I guess you should try both - once where -1 is in the wrapped sensor's mapping to make sure they're preserved, and another to make sure they're added if needed.

dongruoping: yes I agree both cases can happen, added test for that.

chriselion (approved):
Looks good!

PR #4485, opened 2020-09-15
[Release 7] Improve environment finding logic on windows

chriselion:
### Proposed change(s)
Followup from a discussion with @zookae - On Windows, saving e.g. 3DBall would put the executable at `3DBall\UnityEnvironment.exe`. That would mean running with `--env=3DBall` would not fix the executable. This improves the logic by adding an extra check if the existing one fails: search for `*.exe` under the env directory, and exclude `UnityCrashHandler*.exe`

### Types of change(s)
- [x] Other (Small improvement)

### Checklist


ervteng (approved):
Tested, works 👍

PR #4502, opened 2020-09-22
WallJump - handle Agent starting before ModelOverrider

chriselion:
### Proposed change(s)
Overriding WallJump behaviors wasn't working on CI (although it was working in the editor) because the Agent.OnEnabled() would be called first, before the ModelOverrider.OnEnable() could process the commandline arguments. This changes the code to process the commandline args lazily, and also does a bit of cleanup.

Additionally, do some future proofing and allow the extension of the override to be specified multiple times. The extensions will be tried in that order, and if no extension is specified, we’ll try .nn then .onnx (basically current behavior).

https://jira.unity3d.com/browse/MLA-855

### Types of change(s)

- [x] Bug fix

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -14,7 +14,7 @@ jobs:
     - uses: actions/setup-ruby@v1
       with:
         ruby-version: '2.6'
-    - uses: actions/setup-dotnet@v1.6.0
+    - uses: actions/setup-dotnet@v1
       with:
         dotnet-version: '3.1.x'
     - run: dotnet tool install -g dotnet-format --version 4.1.131201
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -14,7 +14,7 @@ jobs:
     - uses: actions/setup-ruby@v1
       with:
         ruby-version: '2.6'
-    - uses: actions/setup-dotnet@v1
+    - uses: actions/setup-dotnet@v1.6.0
       with:
         dotnet-version: '3.1.x'
     - run: dotnet tool install -g dotnet-format --version 4.1.131201
--- a/Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ModelOverrider.cs
+++ b/Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ModelOverrider.cs
@@ -35,7 +35,7 @@ namespace Unity.MLAgentsExamples
 
         string m_BehaviorNameOverrideDirectory;
 
-        string m_OverrideExtension = "nn";
+        private List<string> m_OverrideExtensions = new List<string>();
 
         // Cached loaded NNModels, with the behavior name as the key.
         Dictionary<string, NNModel> m_CachedModels = new Dictionary<string, NNModel>();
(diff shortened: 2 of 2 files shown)


Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ModelOverrider.cs
```diff
@@ -25,7 +25,6 @@ namespace Unity.MLAgentsExamples
     public class ModelOverrider : MonoBehaviour
     {
         HashSet<string> k_SupportedExtensions = new HashSet<string> { "nn", "onnx" };
-        const string k_CommandLineModelOverrideFlag = "--mlagents-override-model";
```
chriselion, line 28: This was an obsolete way of specifying the override; we always use `mlagents-override-model-directory` now.

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ModelOverrider.cs
```diff
@@ -133,7 +125,7 @@ void GetAssetPathFromCommandLine()
                 }
             }
 
-            if (HasOverrides)
+            if (!string.IsNullOrEmpty(m_BehaviorNameOverrideDirectory))
```
chriselion, line 128: "inlined" this otherwise they'd call each other.

chriselion:
`actions/setup-dotnet@v1` currently failing, unrelated to this PR.

.github/workflows/pre-commit.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     - uses: actions/setup-ruby@v1
       with:
         ruby-version: '2.6'
-    - uses: actions/setup-dotnet@v1
+    - uses: actions/setup-dotnet@v1.6.0
```
chriselion, line 17: https://github.com/actions/setup-dotnet/issues/128

harperj (approved):
The code changes themselves here look good to me, but our discussion before make me wonder whether we should extend the logic a bit.  Rather than choosing '.nn' or '.onnx', what do you think about choosing by preference? 

For example we could default to `.nn` and fall back on `.onnx` now, or always prefer `.onnx` and fall back on `.nn`.

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/ModelOverrider.cs
```diff
@@ -192,69 +185,79 @@ public NNModel GetModelForBehaviorName(string behaviorName)
...
-            }
-            else if (!string.IsNullOrEmpty(m_BehaviorNameOverrideDirectory))
+            if (string.IsNullOrEmpty(m_BehaviorNameOverrideDirectory))
             {
-                assetPath = Path.Combine(m_BehaviorNameOverrideDirectory, $"{behaviorName}.{m_OverrideExtension}");
-            }
-
-            if (string.IsNullOrEmpty(assetPath))
-            {
-                Debug.Log($"No override for BehaviorName {behaviorName}, and no directory set.");
```
chriselion, line 207: Simplified the logic here now that we don't have individual overrides - now we can just check m_BehaviorNameOverrideDirectory directly.

Issue #4521, opened 2020-09-28
Accessing Academy.Instance when not playing leaks AcademyFixedUpdateSteppers and EnvironmentStep gets called multiple times per FixedUpdate


Phong13:
**Describe the bug**
Any code that accesses "Academy.Instance" when not "playing" will cause a new hidden gameobject to get created with an AcademyFixedUpdateStepper component. There can easily be many of these objects in a scene. THEY ALL STEP THE ENVIRONMENT!!!. 

It is easy to do this. For example I often test code by writing a context menu function like this:

    _[ContextMenu("Do Some Validation")]
    public void ValidateSomeStuff()
    {
        ...
        Academy myAcademy = Academy.Instance;
        ...
    }_

Every time I validate using the "ContextMenu" a new, hidden, AcademyFixedUpdateStepper gets created.

**To Reproduce**
Add this to a MonoBehaviour and fire it several times when not in Playmode. Each time it is fired a new AcademyFixedUpdateStepper is created:

    _[ContextMenu("Do Some Validation")]
    public void ValidateSomeStuff()
    {
        Academy myAcademy = Academy.Instance;

        // Count the number of hidden Fixed Update Steppers
        Transform[] ts = (Transform[])Resources.FindObjectsOfTypeAll(typeof(Transform));
        int num = 0;
        foreach (var t in ts)
        {
            if (t.GetComponent<AcademyFixedUpdateStepper>() != null)
            {
                num++;
            }
        }

        Debug.Log("Found: " + num + " hidden fixed update steppers");
    }_

Also add a static variable to Academy:

_public static int numTimesEnvironmentStepIsStepped;_

And increment this variable in "EnvironmentStep". Print it out in FixedUpdate.

**Console logs / stack traces**

**Screenshots**

**Environment (please complete the following information):**
- Unity 2019.4
- Windows 10
- Release 6
- 1.7.1
- Any environment

dongruoping:
In your code, a new Academy instance is instantiate every time the function is called by saying `Academy myAcademy = Academy.Instance;`, and that's why you ended up having multiple Academy instances. 

The Academy is initialized when the first time being accessed, so you should be able to call it directly instead of first creating your own `myAcademy`. Please refer to [this doc](https://github.com/Unity-Technologies/ml-agents/blob/master/docs/Learning-Environment-Design.md) for example of accessing Academy.

Phong13:
Hi dongruoping,

[...]

PR #4532, opened 2020-10-01
Clean up AcademyFixedUpdateStepper when playmode changed

dongruoping:
### Proposed change(s)

Clean up AcademyFixedUpdateStepper when playmode changed. Also add insurance check in stepper to delete itself if it's not the stepper of current Academy.

Fixing issue reported in #4521 .
JIRA ticket: https://jira.unity3d.com/browse/MLA-1410

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


com.unity.ml-agents/Runtime/Academy.cs
```diff
@@ -30,6 +30,15 @@ namespace Unity.MLAgents
     /// </summary>
     internal class AcademyFixedUpdateStepper : MonoBehaviour
     {
+        void OnEnable()
+        {
+            // Check if this stepper belongs to current Academy singleton
+            // if (!Academy.Instance.IsStepperOwner(this))
```
chriselion, line 36: Uncomment?
Just to be extra cautious, I think you should either
1) Change the condition to `if(Academy.IsInitialized && !Academy.Instance.IsStepperOwner(this))`
2) Make IsStepperOwner static and check `Academy.IsInitialized` there
Just so that we're not accidentally creating an AcademyInstance when we don't want to  (which is what got us into this in the first place).

com.unity.ml-agents/Runtime/Academy.cs
```diff
@@ -631,5 +658,13 @@ public void Dispose()
             // Reset the Lazy instance
             s_Lazy = new Lazy<Academy>(() => new Academy());
         }
+
+        /// <summary>
+        /// Check if the input AcademyFixedUpdateStepper belongs to this Academy.
+        /// </summary>
+        internal bool IsStepperOwner(AcademyFixedUpdateStepper stepper)
+        {
+            return stepper.gameObject.Equals(Academy.Instance.m_StepperObject);
```
chriselion, line 667: Sounds like you should either user ReferenceEquals, or compare GetInstanceID()
https://answers.unity.com/questions/752382/how-to-compare-if-two-gameobjects-are-the-same-1.html

com.unity.ml-agents/Runtime/Academy.cs
```diff
@@ -32,7 +32,14 @@ internal class AcademyFixedUpdateStepper : MonoBehaviour
     {
         void FixedUpdate()
         {
-            Academy.Instance.EnvironmentStep();
+            if (!Academy.IsStepperOwner(this))
```
chriselion, line 35: I think you should be a little more careful here with how you access the Academy instance so that we don't create a new one accidentally. I'd recommend:

1) Make IsStepperOwner() non-static again
2) In FixedUpdate(), either do nothing if the Academy isn't initialized, or call Destroy() (but avoid the call to IsStepperOwner).

chriselion, line 35: Also, can you add some comments on why this is necessary?

com.unity.ml-agents/CHANGELOG.md
```diff
@@ -31,6 +31,8 @@ and this project adheres to
  camera frame. The scene is able to train with the provided default config file. (#4511)
 ### Bug Fixes
 #### com.unity.ml-agents (C#)
+ - Fixed the issue of leaking AcademyFixedUpdateSteppers when accessing Academy
+ in editor mode. (#4521)
```
chriselion, line 35:
```suggestion
 - Fixed a bug where accessing the Academy outside of play mode would cause the Academy to get stepped multiple times when in play mode. (#4532)
```
Changed the wording a bit, and the made it refer to this PR (not the original issue). Feel free to tweak the wording if you want.

chriselion (approved):
One nit on the changelog. 

Can you also put a link to the jira issue in the PR summary?

Otherwise looks good!

dongruoping:
> Can you also put a link to the jira issue in the PR summary?

Updated in the summary

PR #4543, opened 2020-10-06
Add updated Dockerfile and CI build

harperj:
### Proposed change(s)

* Updates the ML-Agents Docker image based on things we have learned from training in the cloud.
* Automatically pushes the image to our own GCP projects on each release.

https://jira.unity3d.com/browse/MLA-510

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -3,7 +3,6 @@ name: build and push Docker image
 on:
   push:
     tags:
-      - "release_[0-9]+_test[0-9]+"
       - "release_[0-9]+"
 
 jobs:
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -16,5 +16,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
+        with:
+          version: '270.0.0'
+          service_account_email: ${{ secrets.GCP_STG_EMAIL }}
+          service_account_key: ${{ secrets.GCP_STG_PRIVATE_KEY }}
+          export_default_credentials: true
+      - name: Set up docker to authenticate with GCP
+        run: gcloud auth configure-docker
       - name: Build docker image
         run: docker build --tag ml-agents:${GITHUB_SHA} .
--- a/.github/workflows/build-docker.yaml
+++ b/.github/workflows/build-docker.yaml
@@ -25,4 +25,8 @@ jobs:
       - name: Set up docker to authenticate with GCP
         run: gcloud auth configure-docker
       - name: Build docker image
-        run: docker build --tag ml-agents:${GITHUB_SHA} .
+        run: docker build --tag ml-agents:${GITHUB_SHA} --build-arg SHA=${GITHUB_SHA} .
+      - name: Tag docker image (stg)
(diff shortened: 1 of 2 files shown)


harperj:
For @chriselion -- does this belong in the CHANGELOG? I don't believe we consider the Dockerfile officially supported any longer.

chriselion:
Don't think it needs to go in the changelog.

If you want this to run on the upcoming release, you should get it on that branch today; the plan is to release tomorrow morning.

.github/workflows/build-docker.yaml
```diff
@@ -0,0 +1,47 @@
...
+        run: docker build --tag ml-agents:${GITHUB_SHA} --build-arg SHA=${GITHUB_SHA} .
+      - name: Tag docker image (stg)
+        run: |
+          docker tag ml-agents:${GITHUB_SHA} gcr.io/unity-ai-ml-agents-stg/ml-agents:${GITHUB_SHA}
+          docker tag ml-agents:${GITHUB_SHA} gcr.io/unity-ai-ml-agents-stg/ml-agents:latest
+      - name: Push docker image (stg)
+        run: |
+          docker push gcr.io/unity-ai-ml-agents-stg/ml-agents:${GITHUB_SHA}
+          docker push gcr.io/unity-ai-ml-agents-stg/ml-agents:latest
+      - uses: GoogleCloudPlatform/github-actions/setup-gcloud@master
```
chriselion, line 32: Do you want the "prod" runs to execute on test tags? If not you can do something like this:
https://github.com/Unity-Technologies/ml-agents/blob/2dcfebdb49219702d82618dd955c25e26942cef3/.github/workflows/publish_pypi.yaml#L43

(I don't know if there's a better way to do this without repeating the commands and conditions)

.github/workflows/build-docker.yaml
```diff
@@ -0,0 +1,47 @@
...
+        with:
+          version: '270.0.0'
+          service_account_email: ${{ secrets.GCP_STG_EMAIL }}
+          service_account_key: ${{ secrets.GCP_STG_PRIVATE_KEY }}
+          export_default_credentials: true
+      - name: Configure GCR Docker auth (stg)
+        run: gcloud auth configure-docker
+      - name: Build docker image
+        run: docker build --tag ml-agents:${GITHUB_SHA} --build-arg SHA=${GITHUB_SHA} .
```
chriselion, line 23: Do you want the actual SHA here, or the tag? You can extract the tag from the GITHUB_REF variable, but it will be prefixed by `refs/tags/`, e.g. `refs/tags/release_42`

chriselion: see also: https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables

chriselion (approved):
Not my area of expertise but looks OK.

.github/workflows/build-docker.yaml
harperj: I think it would be OK to just not trigger this job at all except on full release tags.

.github/workflows/build-docker.yaml
harperj: Given that the Dockerfile takes a SHA, I think tagging with SHA is reasonable.  Could also add the git ref tag later if needed.

PR #4552, opened 2020-10-09
Add useVectorFrozenFlag option in FoodCollector

dongruoping:
### Proposed change(s)

Bring back useVectorFrozenFlag option that was originally proposed in #4511 , since adding it as default will cause GridFoodCollector to crash.

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


chriselion:
Why does it break GridCollector? Just because the observation size changes?

chriselion (approved):

dongruoping:
> Why does it break GridCollector? Just because the observation size changes?

Yes right.

PR #4562, opened 2020-10-12
[CI] use downloader, update versions

chriselion:
### Proposed change(s)
- Stop using editors from bokken, use downloader instead. Clean up the corresponding code for finding the executable.
- Remove ml-agents/tests/yamato/editmode_tests.py - it hasn't been used since January when we started running package tests.
- Consolidate standalone-build-test and the tests that depend on it to have a common list of versions.


.yamato/test_versions.metafile
```diff
@@ -0,0 +1,6 @@
+# List of editor versions for standalone-build-test and its dependencies.
```
chriselion, line 1: This basically gets "included" at the start of each of the test files, so that they're all using the same versions. See https://internaldocs.hq.unity3d.com/yamato/yaml/templating/#include-metadata-files for more details

chriselion:
Looks like inference is hanging on 2020.2. Debugging that, so hold off on reviews.

.yamato/test_versions.metafile
```diff
@@ -0,0 +1,7 @@
...
+test_editors:
+  - version: 2018.4
+  - version: 2019.4
+  - version: 2020.1
+# Waiting on a barracuda fix, see https://jira.unity3d.com/browse/MLA-1464
+#  - version: 2020.2
```
chriselion, line 7: Inference is broken for the built player on 2020.2. Disabled until fixed in a new barracuda release.

vincentpierre (approved):

PR #4592, opened 2020-10-21
ActionBuffer with hybrid actions for RemotePolicy

dongruoping:
### Proposed change(s)

Allow policy to construct and use ActionBuffer with hybrid actions.

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/com.unity.ml-agents/Runtime/Actuators/IActionReceiver.cs
+++ b/com.unity.ml-agents/Runtime/Actuators/IActionReceiver.cs
@@ -69,6 +69,11 @@ namespace Unity.MLAgents.Actuators
         /// <returns>An <see cref="ActionBuffers"/> instance initialized with a <see cref="ActionSpec"/> and a float array.
         public static ActionBuffers FromActionSpec(ActionSpec actionSpec, float[] actions)
         {
+            if (actions == null)
+            {
+                return new ActionBuffers(ActionSegment<float>.Empty, ActionSegment<int>.Empty);
+            }
+
             Debug.Assert(actions.Length == actionSpec.NumContinuousActions + actionSpec.NumDiscreteActions,
                 $"The length of '{nameof(actions)}' does not match the total size of ActionSpec.\n" +
                 $"{nameof(actions)}.Length: {actions.Length}\n" +
--- a/com.unity.ml-agents/Runtime/Actuators/IActionReceiver.cs
+++ b/com.unity.ml-agents/Runtime/Actuators/IActionReceiver.cs
@@ -62,6 +62,31 @@ namespace Unity.MLAgents.Actuators
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


com.unity.ml-agents/Runtime/Actuators/IActionReceiver.cs
```diff
@@ -61,6 +61,36 @@ public ActionBuffers(ActionSegment<float> continuousActions, ActionSegment<int>
             DiscreteActions = discreteActions;
         }
 
+        /// <summary>
+        /// Create an <see cref="ActionBuffers"/> instance with ActionSpec and all actions stored as a float array.
+        /// </summary>
+        /// <param name="actionSpec"><see cref="ActionSpec"/> of the <see cref="ActionBuffers"/></param>
+        /// <param name="actions">The float array of all actions, including discrete and continuous actions.</param>
+        /// <returns>An <see cref="ActionBuffers"/> instance initialized with a <see cref="ActionSpec"/> and a float array.
+        public static ActionBuffers FromActionSpec(ActionSpec actionSpec, float[] actions)
```
chriselion, line 70: Let's leave this as `internal` for now, since the interface might change later - we're still not sure how the inference results will be passed back from training, but we might get the int[] values too.

com.unity.ml-agents/Runtime/Actuators/IActionReceiver.cs
```diff
@@ -61,6 +61,36 @@ public ActionBuffers(ActionSegment<float> continuousActions, ActionSegment<int>
...
+
+            Debug.Assert(actions.Length == actionSpec.NumContinuousActions + actionSpec.NumDiscreteActions,
+                $"The length of '{nameof(actions)}' does not match the total size of ActionSpec.\n" +
+                $"{nameof(actions)}.Length: {actions.Length}\n" +
+                $"{nameof(actionSpec)}: {actionSpec.NumContinuousActions + actionSpec.NumDiscreteActions}");
+
+            float[] continuousActions = new float[actionSpec.NumContinuousActions];
+            int[] discreteActions = new int[actionSpec.NumDiscreteActions];
+            int offset = 0;
+            Array.Copy(actions, continuousActions, actionSpec.NumContinuousActions);
```
chriselion, line 85: Instead of copying the float actions, you can make an ActionSegment with something like
```
var continuousActionSegment = new ActionSegment(actions, 0, actionSpec.NumContinuousActions)
```
That's one less allocation we need for each agent on each step.

It might also be worth avoiding creating the discreteActions array if NumDiscreteActions==0 (and use `Array.Empty<int>())` instead)

chriselion: This looks good now.

chriselion (approved):

PR #4607, opened 2020-10-26
use int64 steps, check for NaN actions

chriselion:
### Proposed change(s)

* Use int64 for tensorflow policy steps, to avoid int32 overflow after 2**31 steps.
* Also use int64 tensor for torch policy steps. The int32 overflow problem doesn't exist in torch since it use float tensors by default. But if setting the global step to a large number and then call get_current_step() where the step count is converted to int, the conversion will cause numerical error and the result will not be as expected.
* Check for NaN actions before they get sent back to Unity. Currently, we only check the observations, but if we send NaN actions, we'll very likely get NaN observations and "blame" the observations.
* After this change, old Tensorflow model will become incompatible and cannot be loaded by new code. Torch is not affected. 

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


ml-agents/mlagents/trainers/policy/tf_policy.py
```diff
@@ -270,8 +270,16 @@ def get_action(
         )
 
         self.save_memories(global_agent_ids, run_out.get("memory_out"))
+        action = run_out.get("action")
+        # Fast NaN check on the action
+        # See https://stackoverflow.com/questions/6736590/fast-check-for-nan-in-numpy for background.
+        d = np.sum(action)
```
chriselion, line 276: There's a few ways to do this; we were using `np.dot` for observations at one point:
https://github.com/Unity-Technologies/ml-agents/blob/81acfaf1baa93ffd477b1b25e9e668c705a1a56a/ml-agents-envs/mlagents/envs/brain.py#L245
but that morphed to `np.mean` during some refactoring:
https://github.com/Unity-Technologies/ml-agents/pull/3022/files#diff-614d11115c4d74319295ea34b44a3b7d8ab40b43865c1c3e351a8f6c9602c673R92

`np.sum` is one of the recommended approaches in the stackoverflow thread.

ml-agents/mlagents/trainers/policy/tf_policy.py
```diff
@@ -270,8 +270,16 @@ def get_action(
...
+        has_nan = np.isnan(d)
+        if has_nan:
+            raise RuntimeError("NaN action detected.")
```
chriselion, line 279: Just raising for now; there's not much the user can do about it. We could use `np.nan_to_num` but that would probably trigger every step and give bad results.

vincentpierre (approved):
Do not forget to cherry pick on release_10 if you want these changes in the next release

PR #4610, opened 2020-10-28
[bug-fix] Use float64 when converting np.ndarray to torch.tensor, cap Torch version to 1.7.x

ervteng:
### Proposed change(s)

PyTorch changed the behavior of tensor conversion - previously, if `dtype` was None it used the default tensor type. Now, if None it will use the type of the thing that is being converted. 

This PR forces everything to use `float32`.  In addition, we cap the Torch version to 1.7.X so that we won't break when Torch 1.8.0 comes out and breaks something. 

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/setup.py
+++ b/ml-agents/setup.py
@@ -64,7 +64,7 @@ setup(
         "protobuf>=3.6",
         "pyyaml>=3.1.0",
         # Windows ver. of PyTorch doesn't work from PyPi
-        'torch>=1.6.0;platform_system!="Windows"',
+        'torch>=1.6.0,<1.8.0;platform_system!="Windows"',
         "tensorboard>=1.15",
         "cattrs>=1.0.0",
         "attrs>=19.3.0",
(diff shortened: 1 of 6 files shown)


andrewcoh (approved):

ml-agents/setup.py
```diff
@@ -64,7 +64,7 @@ def run(self):
         "protobuf>=3.6",
         "pyyaml>=3.1.0",
         # Windows ver. of PyTorch doesn't work from PyPi
-        'torch>=1.6.0;platform_system!="Windows"',
+        'torch>=1.6.0,<1.8.0;platform_system!="Windows"',
```
chriselion, line 67: Not directly related to this change, but can we put a link to the windows instructions (either to our docs or external docs) as a comment here?

chriselion (approved):
Looks good. If the float32 restriction is a pain and doesn't provide any performance benefit, we can remove the checks for it.

Issue #4596, opened 2020-10-22
Please allow option to export model to .nn without the action_probs model


Phong13:
**Is your feature request related to a problem? Please describe.**
Executing the barracuda models is the biggest performance cost in my project. Using 3 layer by 512 node models, here are the barracuda ModelRunner execution times with different models exported to the .nn file:

-  4.00 ms    .nn file with "Action", "Action_probs", "Value_Estimate"
-  2.65 ms    .nn file with "Action", Action_probs"
-  2.00 ms    .nn file with "Action"

**Describe the solution you'd like**
When using Barracuda models in inference mode, it is a huge performance savings to be executing leaner .nn files with fewer models in them. Please provide an option to control what gets exported to the .nn files so that it is possible to export the leanest possible .nn file. 

Is the "action_probs" model needed in inference mode? Looking at the code in ModelRunner it does not appear that the output from the "action_probs" output is retrieved. Is this model needed by the "action" model somehow? In my testing, the .nn files appear to execute without this model. 

**Describe alternatives you've considered**

**Additional context**

Phong13:
Note that it is useful to be able to export the models to separate independent .nn files. In my project I sometimes use the "value_estimate" and sometimes use the "action" output. These are not used at the same time in my project. It is most performant for me to have two models that can be queried separately.

chriselion:
Hi @Phong13 - sorry for the delay on this.

You're right that it doesn't look like we're using "action_probs" or "value_estimate" at inference time. You could possibly hack the list of output nodes here
https://github.com/Unity-Technologies/ml-agents/blob/672b6081e6d36b85aa76b7014e12341b8ceb0597/ml-agents/mlagents/trainers/tf/model_serialization.py#L43-L45
(although it sounds like that might be what you're already doing). We can look into removing things from this list or making it more configurable in the future.

Would it be possible for you to provide a model file that demonstrates this, so that I can run it by the Barracuda team and make sure there's not a missed optimization opportunity for them? The actual weights don't matter so you can just run training for one step. The easiest format to work with would be the `frozen_graph_def.pb` file that's generated as part of training. Alternatively, if there's one of our example scenes that's similar to your setup and can be easily modified (either in the training configuration or Agent parameters) that would work too.

Thanks!

Phong13:
[...]

PR #4613, opened 2020-10-30
[MLA-1519] Don't mark action_probs as an output node.

chriselion:
### Proposed change(s)
`action_probs` is not used for inference but was marked as an output in TensorFlow models. Remove it from the list of nodes to flag as output.

https://github.com/Unity-Technologies/ml-agents/issues/4596
https://jira.unity3d.com/browse/MLA-1519

### Types of change(s)

- [ ] Bug fix

### Checklist


dongruoping (approved):
Tested with continuous and discrete control and they both works.

PR #4629, opened 2020-11-09
Refactor BarracudaModel loader checks

dongruoping:
### Proposed change(s)

* Move methods for getting model attributes (GetInputNames, GetOutputNames, GetNumVisualInputs) to BarracudaModelExtensions.
* Add `CheckExpectedTensors` to check the existence all expected tensors and call it before all other model checks, to make sure it's safe to call the extension methods and doesn't throw exception.
* Explicitly checks the model has expected action_output_shape/is_continuous for different model format (current/deprecated).
 

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/com.unity.ml-agents/Runtime/Inference/BarracudaModelExtensions.cs
+++ b/com.unity.ml-agents/Runtime/Inference/BarracudaModelExtensions.cs
@@ -1,3 +1,5 @@
+using System.Collections.Generic;
+using System.Linq;
 using Unity.Barracuda;
 
 namespace Unity.MLAgents.Inference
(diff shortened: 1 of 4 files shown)


com.unity.ml-agents/Runtime/Inference/BarracudaModelExtensions.cs
```diff
@@ -7,6 +9,140 @@ namespace Unity.MLAgents.Inference
...
+        /// </summary>
+        /// <param name="model">
+        /// The Barracuda engine model for loading static parameters.
+        /// </param>
+        /// <returns>Array of the output tensor names of the model</returns>
+        public static string[] GetOutputNames(this Model model)
+        {
+            var names = new List<string>();
+
+            if (model == null)
```
chriselion, line 118: I'm not quite sure how nulls work with extension methods - may need to figure out if this was being hit anywhere, and add a null check there instead.

chriselion (approved):

com.unity.ml-agents/Runtime/Inference/BarracudaModelExtensions.cs
dongruoping: From my experiments it works just the same as calling the method directly with a null reference passed in, and these are not triggered anywhere as far as I know.
No model given plus inference type == inference_only is the only case I found the model could be null, and it's already captured properly.

PR #4649, opened 2020-11-13
[bug-fix] Add clipping to PyTorch policy, fix initialization

ervteng:
### Proposed change(s)

This PR is motivated by the perpetually lower resulting rewards that Torch PPO gets on ragdoll environments, especially Walker. It addresses two differences between Torch and TF PPO:

1. Torch layers were initialized using Kaiming Normal, but PyTorch by default assumes that you're using Leaky ReLU and adds a default gain of `sqrt(2)`. This PR changes the nonlinearity assumed in the initializer to `linear`, removing that extra factor of `sqrt(2)`. This initialization was the main cause of wildly higher value losses in Torch. 
2. In TensorFlow, when not using `tanh` squashing on Gaussian outputs (PPO), we truncate the output by `clip(action, -3, 3) / 3`. This PR adds the same thing to Torch. 

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


ml-agents/mlagents/trainers/ppo/optimizer_torch.py
```diff
@@ -168,7 +168,7 @@ def update(self, batch: AgentBuffer, num_sequences: int) -> Dict[str, float]:
         value_loss = self.ppo_value_loss(
             values, old_values, returns, decay_eps, loss_masks
         )
-        policy_loss = self.ppo_policy_loss(
+        policy_loss = 2 * self.ppo_policy_loss(
```
ervteng, line 171: This is the reason why the PR is a WIP. The Policy Loss is 2x lower than in TF, and I can't figure out why. Adding the 2x factor here makes PPO train better.

vincentpierre (approved):
Don't forget to cherry pick to release_10_branch after merge.

PR #4651, opened 2020-11-13
Policy output actiontuple

andrewcoh:
### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


ml-agents-envs/mlagents_envs/base_env.py
```diff
@@ -250,20 +250,13 @@ class ActionTuple:
...
-    ):
-        if continuous is not None and continuous.dtype != np.float32:
-            continuous = continuous.astype(np.float32, copy=False)
-        self._continuous = continuous
-        if discrete is not None and discrete.dtype != np.int32:
-            discrete = discrete.astype(np.int32, copy=False)
-        self._discrete = discrete
+    def __init__(self):
+        self._continuous = None
+        self._discrete = None
```
chriselion, line 259: How about keeping the continuous and discrete arguments, and adding e.g.
```python
if continuous is not None:
  self.add_continuous(continuous)
# TODO same for discrete
```
?

ml-agents/mlagents/trainers/torch/action_log_probs.py
```diff
@@ -5,6 +5,38 @@
 from mlagents.trainers.torch.utils import ModelUtils
 
 
+class LogProbsTuple:
```
chriselion, line 8: If this is identical to ActionTuple in functionality, but you want to make sure that they aren't accidentally used interchangeably, you could  make them both derive from  e.g. `_ActionTupleBase`

chriselion (commented):
I like this a lot better than the dictionary, although the class still feels clunky. I _think_ you can keep the initializer args which should shrink the code a bit.

ml-agents-envs/mlagents_envs/base_env.py
ervteng: +1

ml-agents-envs/mlagents_envs/base_env.py
```diff
@@ -323,20 +330,26 @@ def empty_action(self, n_agents: int) -> ActionTuple:
...
         Generates ActionTuple corresponding to a random action (either discrete
         or continuous) for a number of agents.
         :param n_agents: The number of agents that will have actions generated
         """
+        action_tuple = ActionTuple()
         continuous = np.random.uniform(
             low=-1.0, high=1.0, size=(n_agents, self.continuous_size)
         )
-        discrete = np.zeros((n_agents, self.discrete_size), dtype=np.int32)
+        action_tuple.add_continuous(continuous)
```
ervteng, line 352: Can we just do something like 

```suggestion
        _continuous = continuous
        _discrete = np.column_stack() if self.discrete_size > 0 else None
        action_tuple = ActionTuple(_continuous, _discrete)
```

ervteng (approved):

PR #4694, opened 2020-11-30
pin pip to 20.2 for now

chriselion:
### Proposed change(s)
pip 20.3 throws an exception trying to install our dependencies. pin to 20.2 until they get a fix

We might need this later for tests that run on yamato, but as of the moment, the pypi mirror we use internally hasn't gotten 20.3

https://github.com/pypa/pip/issues/9180

### Types of change(s)
- [x] Bug workaround

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -8,6 +8,7 @@ on:
     - 'gym-unity/**'
     - 'test_constraints*.txt'
     - 'test_requirements.txt'
+    - '.github/workflows/pytest.yml'
   push:
     branches: [master]
 
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -44,7 +44,8 @@ jobs:
       run: python -c "import sys; print(sys.version)"
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


.github/workflows/pytest.yml
```diff
@@ -8,6 +8,7 @@ on:
     - 'gym-unity/**'
     - 'test_constraints*.txt'
     - 'test_requirements.txt'
+    - '.github/workflows/pytest.yml'
```
chriselion, line 11: Make sure tests rerun if this file changes.

andrewcoh (approved):

.github/workflows/pytest.yml
```diff
@@ -44,7 +45,8 @@ jobs:
       run: python -c "import sys; print(sys.version)"
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip
+        # pin pip to workaround https://github.com/pypa/pip/issues/9180
+        python -m pip install pip==20.2
```
pradyunsg, line 49:
```suggestion
        python -m pip install pip==20.2.4
```

chriselion: Thanks for the suggestion, but I think this is enough for now. Hopefully you'll have a patch release soon :)

PR #4699, opened 2020-12-02
Fix SubprocessEnvManager hanging on unexpected exceptions.

hvpeteet:
### Proposed change(s)

Fixes 2 issues:

1. **Issue:** SubprocessEnvManager could hang if a subprocess sent an unexpected exception (e.g. NAN error). **fix:** This is fixed by forwarding unknown exceptions (in addition to the previous whitelist of specific exceptions) to the manager from the workers.
1. **Issue:** SubprocessEnvManager could miss handling exception messages from workers if the workers shut down before flushing their messages to the queue. **fix:** This is fixed by removing the cancel_join_thread() call, and instead sending a sentinel value to indicate worker shutdown. The manager will now wait for workers to shutdown for 10 seconds before killing them rather than possibly shutting down before them.

https://jira.unity3d.com/browse/MLA-1199

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### Other comments
N/A

### DIFF
--- a/ml-agents/mlagents/trainers/subprocess_env_manager.py
+++ b/ml-agents/mlagents/trainers/subprocess_env_manager.py
@@ -1,5 +1,6 @@
 from typing import Dict, NamedTuple, List, Any, Optional, Callable, Set
 import cloudpickle
+from datetime import datetime, timedelta
 import enum
 
 from mlagents_envs.environment import UnityEnvironment
@@ -48,6 +49,7 @@ class EnvironmentCommand(enum.Enum):
     RESET = 4
     CLOSE = 5
     ENV_EXITED = 6
+    CLOSED = 7
 
 
 class EnvironmentRequest(NamedTuple):
@@ -93,7 +95,7 @@ class UnityEnvWorker:
         except (BrokenPipeError, EOFError):
             raise UnityCommunicationException("UnityEnvironment worker: recv failed.")
 
-    def close(self):
+    def request_close(self):
         try:
             self.conn.send(EnvironmentRequest(EnvironmentCommand.CLOSE))
         except (BrokenPipeError, EOFError):
@@ -101,8 +103,6 @@ class UnityEnvWorker:
                 f"UnityEnvWorker {self.worker_id} got exception trying to close."
             )
             pass
-        logger.debug(f"UnityEnvWorker {self.worker_id} joining process.")
-        self.process.join()
 
 
 def worker(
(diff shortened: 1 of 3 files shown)


ml-agents/mlagents/trainers/subprocess_env_manager.py
```diff
@@ -306,10 +315,31 @@ def training_behaviors(self) -> Dict[BehaviorName, BehaviorSpec]:
...
         self.step_queue.close()
+        # Sanity check to kill zombie workers and report an issue if they occur.
+        if self.workers_alive > 0:
+            logger.error("SubprocessEnvManager had workers that didn't signal shutdown")
+            for env_worker in self.env_workers:
+                if not env_worker.closed and env_worker.process.is_alive():
+                    env_worker.process.terminate()
+                    raise AssertionError(
+                        "A SubprocessEnvManager worker did not shut down correctly"
+                    )
```
hvpeteet, line 341: I am not sure if we need this, it may be good enough to log the warning.

chriselion: I think logging is better here; raising will mean that we don't try to shut down other workers that might be stuck, and we'd have to handle the error higher up too.

ml-agents/mlagents/trainers/subprocess_env_manager.py
```diff
@@ -186,17 +188,22 @@ def _generate_all_results() -> AllStepResult:
             EnvironmentResponse(EnvironmentCommand.ENV_EXITED, worker_id, ex)
         )
         _send_response(EnvironmentCommand.ENV_EXITED, ex)
+    except Exception as ex:
+        logger.info(
```
chriselion, line 192:
```suggestion
        logger.error(
```

ml-agents/mlagents/trainers/subprocess_env_manager.py
```diff
@@ -306,10 +315,31 @@ def training_behaviors(self) -> Dict[BehaviorName, BehaviorSpec]:
 
     def close(self) -> None:
         logger.debug("SubprocessEnvManager closing.")
+        for env_worker in self.env_workers:
+            env_worker.request_close()
+        # Pull messages out of the queue until every worker has CLOSED or we time out.
+        deadline = datetime.now() + WORKER_SHUTDOWN_TIMEOUT
```
chriselion, line 321: Generally I prefer `time.time()` (which returns epoch time in seconds) but that's a matter of taste.

chriselion (approved):

ml-agents/mlagents/trainers/subprocess_env_manager.py
hvpeteet: Switched

PR #4732, opened 2020-12-09
Adds SensorTypes and GoalSensors

awjuliani:
### Proposed change(s)

Replicates changes from https://github.com/Unity-Technologies/ml-agents/tree/develop-taggedobservations with `TaskSensor` changes to `GoalSensor`. This was easier than cherry picking from the original branch, as it was quite outdated after the hybrid action changes.

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


com.unity.ml-agents/Runtime/Sensors/ISensor.cs
```diff
@@ -16,6 +16,28 @@ public enum SensorCompressionType
         PNG
     }
 
+    /// <summary>
```
andrewcoh, line 19: Whether or not this enum makes it into master will ultimately be up to the toolkit pod but for our purposes whatever works.

ml-agents-envs/mlagents_envs/base_env.py
```diff
@@ -367,10 +374,13 @@ class BehaviorSpec(NamedTuple):
     - observation_shapes is a List of Tuples of int : Each Tuple corresponds
     to an observation's dimensions. The shape tuples have the same ordering as
     the ordering of the DecisionSteps and TerminalSteps.
+    - sensor_types is a List of SensorTypes, each corresponding to the type of
+    sensor (i.e. observation, goal, etc).
     - action_spec is an ActionSpec NamedTuple
     """
 
     observation_shapes: List[Tuple]
+    sensor_types: List[SensorType]
```
andrewcoh, line 383: Looks like we need to add on ObsSpec tuple to contain sensor types and obs shapes

andrewcoh (approved):

com.unity.ml-agents/Runtime/Sensors/ISensor.cs
awjuliani: Agreed. This is just for testing purposes.

ml-agents-envs/mlagents_envs/base_env.py
awjuliani: Oh yeah. I was using the old code, but it makes sense to do this when we want to make it official.

PR #4733, opened 2020-12-09
Adds goal signal to GridWorld environment

awjuliani:
### Proposed change(s)

Adds goal signal to GridWorld environment. Will need https://github.com/Unity-Technologies/ml-agents/pull/4732 before GoalSensor can be added to environment. 

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/Project/Assets/ML-Agents/Examples/GridWorld/Scripts/GridAgent.cs
+++ b/Project/Assets/ML-Agents/Examples/GridWorld/Scripts/GridAgent.cs
@@ -2,6 +2,7 @@ using System;
 using UnityEngine;
 using System.Linq;
 using Unity.MLAgents;
+using Unity.MLAgents.Sensors;
 using Unity.MLAgents.Actuators;
 using UnityEngine.Serialization;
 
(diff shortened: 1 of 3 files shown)


Project/Assets/ML-Agents/Examples/GridWorld/Scripts/GridAgent.cs
```diff
@@ -67,6 +84,18 @@ public override void WriteDiscreteActionMask(IDiscreteActionMask actionMask)
         }
     }
 
+    private void ProvideReward(GridGoal hitObject)
+    {
+        if (gridGoal == hitObject)
```
andrewcoh, line 89: I wonder if we should ignore collision with the wrong goal, rather than give it a negative reward.  This is what im doing in WallJump/Pushblock.

andrewcoh (approved):

Project/Assets/ML-Agents/Examples/GridWorld/Scripts/GridAgent.cs
awjuliani: Well, I kept the original reward function. I worry that doing what you say will make the task too easy? I'd be curious to see the Push/Jump task with the +1/-1 reward.

PR #4742, opened 2020-12-13
set multiple obstacles from area

brccabral:
added a parameter to GridArea to control how many obstacles will be created in the scene


awjuliani:
Hi @brccabral 

Thanks for making this PR. It looks like you also change the default number of obstacles to 2 in the inspector here. Could you change that back to 1? Otherwise, I think our users will appreciate this additional flexibility.

brccabral:
Done @awjuliani .

awjuliani (approved):
Looks good to me.

awjuliani:
Thank you for the contribution, @brccabral !

PR #4746, opened 2020-12-14
Hybrid FoodCollector

dongruoping:
### Proposed change(s)

FoddCollector with continuous moving and discrete shooting.

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


Project/Assets/ML-Agents/Examples/FoodCollector/Scripts/FoodCollectorAgent.cs
```diff
@@ -139,16 +129,7 @@ public void MoveAgent(ActionSegment<int> act)
             if (shootCommand)
             {
                 m_Shoot = true;
-                dirToGo *= 0.5f;
```
chriselion, line 142: This got lost in the conversion, right?

Project/Assets/ML-Agents/Examples/FoodCollector/Scripts/FoodCollectorAgent.cs
```diff
@@ -216,32 +197,34 @@ void Unsatiate()
...
-        discreteActionsOut[1] = 0;
-        discreteActionsOut[2] = 0;
+        var continuousActionsOut = actionsOut.ContinuousActions;
+        continuousActionsOut[0] = 0;
+        continuousActionsOut[1] = 0;
+        continuousActionsOut[2] = 0;
         if (Input.GetKey(KeyCode.D))
         {
-            discreteActionsOut[2] = 2;
+            continuousActionsOut[2] = 2;
```
chriselion, line 212: These should be 1 and -1 (or the other way around), not 1 and 2

Project/Assets/ML-Agents/Examples/FoodCollector/Scripts/FoodCollectorAgent.cs
chriselion: Maybe easier to keep everything in one function and just pass the ActionBuffers, instead of doing it individually.

docs/Learning-Environment-Examples.md
```diff
@@ -315,11 +315,15 @@ you would like to contribute environments, please see our
...
-  - Vector Action space: (Discrete) 4 Branches:
-    - Forward Motion (3 possible actions: Forward, Backwards, No Action)
-    - Side Motion (3 possible actions: Left, Right, No Action)
-    - Rotation (3 possible actions: Rotate Left, Rotate Right, No Action)
-    - Laser (2 possible actions: Laser, No Action)
+  - Vector Action space:
+    - `FoodCollector` scene: 3 Continuous Actions + 1 Discrete Action with 2 Branches
+      - 3 continuous actions correspond to Forward Motion, Side Motion and Rotation
+      - Laser (2 possible actions: Laser, No Action)
+    - `GridFoodCollector` and `VisualFoodCollector` scenes: (Discrete) 4 Branches:
```
chriselion, line 322: I think we either need to change the ActionSpecs for the other scenes (and retrain), or have two versions of the code to handle discrete or continuous movement.

chriselion: I'd prefer the former (less code to maintain) but it's your call.

andrewcoh, line 322: Shouldn't these be the same as above since the share the FoodCollectorAgent script?

ml-agents/mlagents/trainers/demo_loader.py
```diff
@@ -70,15 +70,28 @@ def make_demo_buffer(
...
-            demo_raw_buffer["continuous_action"].append(
-                current_pair_info.action_info.vector_actions_deprecated
-            )
-        if behavior_spec.action_spec.discrete_size > 0:
-            demo_raw_buffer["discrete_action"].append(
-                current_pair_info.action_info.vector_actions_deprecated
-            )
+        if (
+            current_pair_info.action_info.continuous_actions is None
+            and current_pair_info.action_info.continuous_actions is None
```
chriselion, line 75:
```suggestion
            and current_pair_info.action_info.discrete_actions is None
```
(I think, not totally sure)

com.unity.ml-agents/Editor/DemonstrationDrawer.cs
```diff
@@ -23,9 +24,9 @@ internal class DemonstrationEditor : UnityEditor.Editor
         const string k_NumberStepsName = "numberSteps";
         const string k_NumberEpisodesName = "numberEpisodes";
         const string k_MeanRewardName = "meanReward";
-        const string k_ActionSpecName = "ActionSpec";
+        const string k_ActionSpecName = "m_ActionSpec";
```
chriselion, line 27: What was the problem here? Should we have a unit test that would have caught it?

(OK to do outside of this PR)

chriselion (approved):

com.unity.ml-agents/Editor/DemonstrationDrawer.cs
dongruoping: I wrote this when we were adding an ActionSpec field, and later on we changed it to private field m_ActionSpec and public property ActionSpec. 

Since properties won't be serialized, `property.FindPropertyRelative("ActionSpec")` will get nothing and the demo import will fail. The exception is ignored though, resulted in demo files not imported properly and no information displayed in inspector, but no direct error message.

chriselion: Ah sorry, missed the fact that it was in the drawer.

chriselion (approved):

PR #4767, opened 2020-12-17
[CI] move as much as possible to linux

chriselion:
### Proposed change(s)
* Change as many tests as possible to run on Linux instead of Mac.
* Extend the StandaloneBuild command to allow building for linux. Note that we have to keep the (now obsolete) "OSX" in the name, since the backcompat tests will still need to call that. Switch all tests that use the standalone executable (e.g. LL-API) to linux.
* Training tests now run in 2 parts. For current C# and current python, run on linux (depending on the standalone linux build job). Backwards compat tests still need to run on Mac because we need the custom TF build (and because we're running old versions of the C# code that don't know how to build for linux). 
* Performance tests and backwards compatibility tests moved to nightly triggers (still on Mac)

### Types of change(s)
- [x] CI


.yamato/compressed-sensor-test.yml
```diff
@@ -5,7 +5,7 @@ test_compressed_obs_{{ editor.version }}:
   name: Test Compressed Sensor Observation {{ editor.version }}
   agent:
     type: Unity::VM::osx
-    image: ml-agents/ml-agents-bokken-mac:0.1.4-492264
```
chriselion, line 8: Not sure this will help much, but might be faster job startup times if images are cached.

.yamato/com.unity.ml-agents-test.yml
```diff
@@ -92,7 +92,7 @@ test_{{ package.name }}_{{ platform.name }}_{{ editor.version }}:
     - .yamato/com.unity.ml-agents-pack.yml#pack
   triggers:
     cancel_old_ci: true
-    {% if platform.name == "mac" %}
+    {% if platform.name == "linux" %}
```
chriselion, line 95: Run package tests on linux

ml-agents/tests/yamato/yamato_utils.py
```diff
@@ -115,6 +116,8 @@ def init_venv(
     pip_commands = ["--upgrade pip", "--upgrade setuptools"]
     if mlagents_python_version:
         # install from pypi
+        if platform != "darwin":
```
chriselion, line 119: `pip install` below would fail, this is just a friendlier error.

.yamato/compressed-sensor-test.yml
chriselion: NM, running on linux now.

Project/Assets/ML-Agents/Editor/Tests/StandaloneBuildTest.cs
```diff
@@ -28,13 +30,17 @@ public static void BuildStandalonePlayerOSX()
                 {
                     scenePath = args[i + 1];
                 }
+                else if (args[i] == k_BuildTargetFlag)
+                {
+                    buildTarget = (BuildTarget)Enum.Parse(typeof(BuildTarget), args[i + 1], ignoreCase: true);
```
chriselion, line 35: TIL!

surfnerd (approved):
sweet.  Thank you.

.yamato/test_versions.metafile
```diff
@@ -2,13 +2,13 @@
 # csharp_backcompat_version is used in training-int-tests to determine the
 # older package version to run the backwards compat tests against.
 test_editors:
-  - version: 2018.4
```
chriselion, line 5: Will re-enable these before merging to master

.yamato/training-int-tests.yml
```diff
@@ -43,6 +40,59 @@ test_mac_training_int_{{ editor.version }}:
         - "artifacts/inference.onnx.txt"
     standalonebuild:
       paths:
-        - "artifacts/testplayer*/**"
+        - "artifacts/testPlayer*/**"
         - "artifacts/models/**"
 {% endfor %}
+
+
+test_mac_backcompat_2020.1:
```
chriselion, line 48: Will move this to its own file with a nightly trigger instead of running on every PR.

surfnerd (approved):

PR #4780, opened 2020-12-19
[MLA-1540] Training Analytics

chriselion:
### Proposed change(s)
Adds analytics events for training. We will send  3 events:
* 1 for info collected on the Unity side for each Behavior being trained, with information like ActionSpec and ObservationSpecs.
* 1 for info collected on the python side for each Behavior being trained, with information like PPO/SAC, curriculum, etc.
* 1 for info collected on the python side for "global" information like python version, torch version, etc.

The reason for splitting the two per-behavior events is so that we can backport the first to the verified branch and get some data from there too (without needing to rely on trainer upgrades).

Data is sent from python to the Unity executable, from there we'll send it through EditorAnalytics so the [Unity Privacy Policy](https://unity3d.com/legal/privacy-policy) will apply.

### Example events
These are from one of the behaviors from the WallJump scene with curriculum.
#### ml_agents_training_environment_initialized
```json
 {
    "TrainingSessionGuid": "6a709347-d170-4b6d-90b9-f74f76c9df27",
    "TrainerPythonVersion": "3.8.7",
    "MLAgentsVersion": "0.24.0.dev0",
    "MLAgentsEnvsVersion": "0.24.0.dev0",
    "TorchVersion": "1.7.1",
    "TorchDeviceType": "cpu",
    "NumEnvironments": 1,
    "NumEnvironmentParameters": 2
}
```

#### ml_agents_remote_policy_initialized
```json
{
    "TrainingSessionGuid": "6a709347-d170-4b6d-90b9-f74f76c9df27",
    "BehaviorName": "df93fe80efe2beadbbb54f3a4d7eee25",
    "ObservationSpecs": [
        {
            "SensorName": "StackingSensor_size6_OffsetRayPerceptionSensor",
            "CompressionType": "None",
            "DimensionInfos": [
                {
                    "Size": 210,
                    "Flags": 0
                }
            ]
        },
        {
            "SensorName": "StackingSensor_size6_RayPerceptionSensor",
            "CompressionType": "None",
            "DimensionInfos": [
                {
                    "Size": 210,
                    "Flags": 0
                }
            ]
        },
        {
            "SensorName": "StackingSensor_size6_VectorSensor_size4",
            "CompressionType": "None",
            "DimensionInfos": [
                {
                    "Size": 24,
                    "Flags": 0
                }
            ]
        }
    ],
    "ActionSpec": {
        "NumContinuousActions": 0,
        "NumDiscreteActions": 4,
        "BranchSizes": [
            3,
            3,
[...]

PR #4816, opened 2021-01-03
Added SUM as aggregation type for custom statistics

brccabral:
Added SUM as aggregation type for custom statistics

### DIFF
--- a/Project/Assets/ML-Agents/Examples/Hallway/Scripts/HallwayAgent.cs
+++ b/Project/Assets/ML-Agents/Examples/Hallway/Scripts/HallwayAgent.cs
@@ -160,5 +160,7 @@ public class HallwayAgent : Agent
             symbolXGoal.transform.position = new Vector3(7f, 0.5f, 22.29f) + area.transform.position;
             symbolOGoal.transform.position = new Vector3(-7f, 0.5f, 22.29f) + area.transform.position;
         }
+        statsRecorder.Add("Goal/Correct", 0, StatAggregationMethod.Sum);
+        statsRecorder.Add("Goal/Wrong", 0, StatAggregationMethod.Sum);
     }
 }
--- a/Project/Assets/ML-Agents/Examples/Hallway/Scripts/HallwayAgent.cs
+++ b/Project/Assets/ML-Agents/Examples/Hallway/Scripts/HallwayAgent.cs
@@ -18,7 +18,7 @@ public class HallwayAgent : Agent
     Renderer m_GroundRenderer;
     HallwaySettings m_HallwaySettings;
     int m_Selection;
-    StatsRecorder statsRecorder;
+    StatsRecorder m_statsRecorder;
 
     public override void Initialize()
     {
@@ -26,7 +26,7 @@ public class HallwayAgent : Agent
         m_AgentRb = GetComponent<Rigidbody>();
         m_GroundRenderer = ground.GetComponent<Renderer>();
         m_GroundMaterial = m_GroundRenderer.material;
-        statsRecorder = Academy.Instance.StatsRecorder;
+        m_statsRecorder = Academy.Instance.StatsRecorder;
     }
 
     public override void CollectObservations(VectorSensor sensor)
@@ -85,13 +85,13 @@ public class HallwayAgent : Agent
             {
                 SetReward(1f);
                 StartCoroutine(GoalScoredSwapGroundMaterial(m_HallwaySettings.goalScoredMaterial, 0.5f));
-                statsRecorder.Add("Goal/Correct", 1, StatAggregationMethod.Sum);
+                m_statsRecorder.Add("Goal/Correct", 1, StatAggregationMethod.Sum);
             }
             else
             {
                 SetReward(-0.1f);
                 StartCoroutine(GoalScoredSwapGroundMaterial(m_HallwaySettings.failMaterial, 0.5f));
-                statsRecorder.Add("Goal/Wrong", 1, StatAggregationMethod.Sum);
+                m_statsRecorder.Add("Goal/Wrong", 1, StatAggregationMethod.Sum);
             }
             EndEpisode();
         }
@@ -160,7 +160,7 @@ public class HallwayAgent : Agent
             symbolXGoal.transform.position = new Vector3(7f, 0.5f, 22.29f) + area.transform.position;
             symbolOGoal.transform.position = new Vector3(-7f, 0.5f, 22.29f) + area.transform.position;
         }
-        statsRecorder.Add("Goal/Correct", 0, StatAggregationMethod.Sum);
-        statsRecorder.Add("Goal/Wrong", 0, StatAggregationMethod.Sum);
+        m_statsRecorder.Add("Goal/Correct", 0, StatAggregationMethod.Sum);
+        m_statsRecorder.Add("Goal/Wrong", 0, StatAggregationMethod.Sum);
     }
 }
--- a/Project/Assets/ML-Agents/Examples/Hallway/Scripts/HallwayAgent.cs
+++ b/Project/Assets/ML-Agents/Examples/Hallway/Scripts/HallwayAgent.cs
@@ -18,6 +18,7 @@ public class HallwayAgent : Agent
     Renderer m_GroundRenderer;
     HallwaySettings m_HallwaySettings;
     int m_Selection;
+    StatsRecorder statsRecorder;
 
     public override void Initialize()
     {
@@ -25,6 +26,7 @@ public class HallwayAgent : Agent
         m_AgentRb = GetComponent<Rigidbody>();
         m_GroundRenderer = ground.GetComponent<Renderer>();
         m_GroundMaterial = m_GroundRenderer.material;
+        statsRecorder = Academy.Instance.StatsRecorder;
     }
 
     public override void CollectObservations(VectorSensor sensor)
@@ -83,11 +85,13 @@ public class HallwayAgent : Agent
             {
                 SetReward(1f);
                 StartCoroutine(GoalScoredSwapGroundMaterial(m_HallwaySettings.goalScoredMaterial, 0.5f));
+                statsRecorder.Add("Goal/Correct", 1, StatAggregationMethod.Sum);
             }
             else
             {
                 SetReward(-0.1f);
                 StartCoroutine(GoalScoredSwapGroundMaterial(m_HallwaySettings.failMaterial, 0.5f));
+                statsRecorder.Add("Goal/Wrong", 1, StatAggregationMethod.Sum);
             }
             EndEpisode();
         }
--- a/ml-agents-envs/mlagents_envs/side_channel/stats_side_channel.py
+++ b/ml-agents-envs/mlagents_envs/side_channel/stats_side_channel.py
@@ -14,6 +14,9 @@ class StatsAggregationMethod(Enum):
     # Only the most recent value is reported.
     MOST_RECENT = 1
 
+    # Values within the summary period are summed up before reporting.
+    SUM = 2
+
 
 StatList = List[Tuple[float, StatsAggregationMethod]]
 EnvironmentStats = Mapping[str, StatList]
@@ -35,6 +38,7 @@ class StatsSideChannel(SideChannel):
     def on_message_received(self, msg: IncomingMessage) -> None:
         """
         Receive the message from the environment, and save it for later retrieval.
+
         :param msg:
         :return:
         """
@@ -47,6 +51,7 @@ class StatsSideChannel(SideChannel):
     def get_and_reset_stats(self) -> EnvironmentStats:
         """
         Returns the current stats, and resets the internal storage of the stats.
+
         :return:
         """
         s = self.stats
--- a/ml-agents/mlagents/trainers/agent_processor.py
+++ b/ml-agents/mlagents/trainers/agent_processor.py
@@ -332,11 +332,11 @@ class AgentManager(AgentProcessor):
         for stat_name, value_list in env_stats.items():
             for val, agg_type in value_list:
                 if agg_type == StatsAggregationMethod.AVERAGE:
-                    self.stats_reporter.add_stat(stat_name, val)
+                    self.stats_reporter.add_stat(stat_name, val, agg_type)
+                elif agg_type == StatsAggregationMethod.SUM:
+                    self.stats_reporter.add_stat(stat_name, val, agg_type)
                 elif agg_type == StatsAggregationMethod.MOST_RECENT:
                     # In order to prevent conflicts between multiple environments,
                     # only stats from the first environment are recorded.
                     if worker_id == 0:
                         self.stats_reporter.set_stat(stat_name, val)
-                elif agg_type == StatsAggregationMethod.SUM:
-                    self.stats_reporter.add_stat(stat_name, val, agg_type)
--- a/ml-agents/mlagents/trainers/agent_processor.py
+++ b/ml-agents/mlagents/trainers/agent_processor.py
@@ -71,7 +71,6 @@ class AgentProcessor:
     ) -> None:
         """
         Adds experiences to each agent's experience history.
-
         :param decision_steps: current DecisionSteps.
         :param terminal_steps: current TerminalSteps.
         :param previous_action: The outputs of the Policy's get_action method.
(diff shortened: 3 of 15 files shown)


chriselion (changes requested):
Overall, I think this is a good change and I can see how it would be useful. I have some feedback on the particular implementation though. There are two (mostly mutually exclusive) ways to go from here:

## Option 1

Instead of renaming `StatsSummary.mean`, you could add a `sum` field, and also add an `aggregation_method` field. Then in the StatsWriter implementation (for example https://github.com/Unity-Technologies/ml-agents/blob/f91212bebefa5190e7b1447aa358313c8494760c/ml-agents/mlagents/trainers/stats.py#L173) you can use aggregation_method to decide whether to use `mean` or `sum`.

You could also add an `aggregated_value` property to `StatsSummary` like
```python
@property
def aggregated_value(self):
  return self.sum if aggregated_value == StatsAggregationMethod.SUM else self.mean
```

## Option 2
Add a method to `StatsReporter` like `increment_stat`, and call that from `record_environment_stats()` when `agg_type == SUM`. That would look something like

``` python
    def increment_stat(self, key: str, value: float) -> None:
        with StatsReporter.lock:
            if StatsReporter.stats_dict[self.category][key]:
                # Add to the last stat. If we're always using increment_stat, this
                # should be the only element in the list. 
                StatsReporter.stats_dict[self.category][key][-1] += value
            else:
                # New list with the value as the only element
                StatsReporter.stats_dict[self.category][key] = [value]
```

(but make sure you test it 😃 )

---

I think I prefer the second approach because it's more consistent with the existing idea of "aggregating" in the StatsReporter and it doesn't require storing the AggregationMethod, but the first approach is more flexible (especially since we're planning to let users add their own StatsWriters soon, see the [plugins PR](https://github.com/Unity-Technologies/ml-agents/pull/4780)).

Do you have a preference for either one?

ml-agents-envs/mlagents_envs/side_channel/stats_side_channel.py
```diff
@@ -35,6 +38,7 @@ def __init__(self) -> None:
     def on_message_received(self, msg: IncomingMessage) -> None:
         """
         Receive the message from the environment, and save it for later retrieval.
+
```
chriselion, line 41: Please remove the extra whitespaces that were added in the docstrings.

ml-agents/mlagents/trainers/tests/test_agent_processor.py
```diff
@@ -272,8 +272,8 @@ def test_agent_manager_stats():
         manager.record_environment_stats(env_stats, worker_id=0)
 
     expected_stats = {
-        "averaged": StatsSummary(mean=2.0, std=mock.ANY, num=2),
-        "most_recent": StatsSummary(mean=4.0, std=0.0, num=1),
+        "averaged": StatsSummary(stats_value=2.0, std=mock.ANY, num=2),
```
chriselion, line 275: You should also extend this test to include `StatsAggregationMethod.SUM`

ml-agents/mlagents/trainers/stats.py
```diff
@@ -126,13 +130,14 @@ def write_stats(
             if self.rank is not None:
                 log_info.append(f"Rank: {self.rank}")
 
-            log_info.append(f"Mean Reward: {stats_summary.mean:0.3f}")
+            log_info.append(f"Mean Reward: {stats_summary.stats_value:0.3f}")
             log_info.append(f"Std of Reward: {stats_summary.std:0.3f}")
+            log_info.append(f"Num of Reward: {stats_summary.num:0.3f}")
```
chriselion, line 135: I don't think you should add this; it's not useful for training.

ml-agents/mlagents/trainers/tests/test_learn.py
```diff
@@ -75,17 +76,24 @@ def test_run_training(
                 learn.run_training(0, options)
                 mock_init.assert_called_once_with(
                     trainer_factory_mock.return_value,
-                    "results/ppo",
+                    os.path.join("results", "ppo"),
```
chriselion, line 79: I don't think this is a bad change, but does it need to happen in this PR? Does this test fail on a certain platform (Windows?) without it?

ml-agents-envs/mlagents_envs/side_channel/stats_side_channel.py
brccabral: the extra spaces are necessary for PyCharm

ml-agents/mlagents/trainers/stats.py
brccabral: For me it was. If you have a Decision Request script attached, the expected value may change.
In my case, MaxStep = 3000, DecisionRequest = 6, but I wasn't sure how this works at that time.
My custom stat was 0.3 every step.
0.3 * 3000 = 900
but instead I was getting 150, and I didn't know why.
Only after I added the "Num of Reward" I noticed that I was getting 500 Num.
That is because 3000 / 6 = 500.
So, 
0.3 * 3000 / 6 = 150.
Because I didn't know how DecisionRequest worked, it took me a while to figure this out.
With the "Num of Rewards" I think it would have saved me a few hours.
Well, that was my experience, I won't mind if you want to remove it

ml-agents/mlagents/trainers/tests/test_learn.py
brccabral: Yes, it fails in Windows (I am using Windows)

brccabral:
Hi @chriselion 
In my commit "b7328ba" I did just as you mentioned in Option 1.

```
class StatsSummary(NamedTuple):
    mean: float
    std: float
    sum: float
    num: int
    aggregation: StatsAggregationMethod

    @staticmethod
    def empty() -> "StatsSummary":
        return StatsSummary(0.0, 0.0, 0.0, 0, StatsAggregationMethod.AVERAGE)
```

but I would have to rewrite all the writers classes (```ConsoleWriter```, ```GaugeWriter```, ```TensorboardWriter```) as they all have their own ```write_stats()``` method.
By renaming it to "stats_value", all three classes get the implementation without major changes.
Also, if the aggregation is SUM, there is no need to store AVG data in ```stats_dict[]```.
And, in the future, I could use other aggregations, MAX/MIN/PERCENTILE.
Would the class StatsSummary have more and more fields?

Option 2 would not give much flexibility to add other aggregations, we are writing our own implementations, but numpy already have them implemented.

ml-agents/mlagents/trainers/tests/test_agent_processor.py
brccabral: New commit to extend it

ml-agents/mlagents/trainers/tests/test_learn.py
chriselion: OK. I'll make a note for us to run pytest on Windows too.

ml-agents-envs/mlagents_envs/side_channel/stats_side_channel.py
chriselion: Please try to find the settings to disable them; I use PyCharm as do several other team members, and this hasn't been a problem before. At a minimum, you need to revert files that only contain the whitespace changes:
* ml-agents/mlagents/trainers/ppo/trainer.py
* ml-agents/mlagents/trainers/torch/components/reward_providers/gail_reward_provider.py
* ml-agents/mlagents/trainers/trainer/rl_trainer.py

On the other hand, if you know of a way to automate this from the command line (and enforce that the `:param ...` names match the args), I'd love that in another PR.

ml-agents/mlagents/trainers/stats.py
```diff
@@ -234,26 +243,36 @@ def add_property(self, property_type: StatsPropertyType, value: Any) -> None:
...
         """
         Add a float value stat to the StatsReporter.
+
         :param key: The type of statistic, e.g. Environment/Reward.
         :param value: the value of the statistic.
+        :param aggregation: the aggregation method for the statistic, default StatsAggregationMethod.AVERAGE.
         """
         with StatsReporter.lock:
             StatsReporter.stats_dict[self.category][key].append(value)
+            StatsReporter.stats_aggregation[self.category][key] = aggregation
```
chriselion, line 269: `set_stat()` below should also needs this line to set stats_aggregation

chriselion:
>  I would have to rewrite all the writers classes (ConsoleWriter, GaugeWriter, TensorboardWriter) as they all have their own write_stats() method.

If you add the property I suggested, you wouldn't need to add any extra logic in the writers. You could even call the property `stats_value`.

> Also, if the aggregation is SUM, there is no need to store AVG data in stats_dict[].

Maybe, but I'd rather not make the logic any more complicated.

> And, in the future, I could use other aggregations, MAX/MIN/PERCENTILE. Would the class StatsSummary have more and more fields?

We can add them as we need them.

ml-agents/mlagents/trainers/stats.py
```diff
@@ -279,14 +299,25 @@ def write_stats(self, step: int) -> None:
...
-        :returns: A StatsSummary NamedTuple containing (mean, std, count).
+        :returns: A StatsSummary NamedTuple containing (mean, std, sum, count, aggregation).
         """
         if len(StatsReporter.stats_dict[self.category][key]) > 0:
-            return StatsSummary(
-                mean=np.mean(StatsReporter.stats_dict[self.category][key]),
-                std=np.std(StatsReporter.stats_dict[self.category][key]),
-                num=len(StatsReporter.stats_dict[self.category][key]),
-            )
+            if (
```
chriselion, line 308: It's not your fault, but this method is getting messy and should be cleaned up. Something like
```python
stat_values = StatsReporter.stats_dict[self.category][key]
if len(stat_values) == 0:
    return StatsSummary.empty()

return StatsSummary(
    mean=np.mean(stat_values),
    sum=np.sum(stat_values),
    std=np.std(stat_values),
    num=len(stat_values),
    aggregation_method = StatsReporter.stats_aggregation[self.category][key],
)
```
should be a bit cleaner.

brccabral:
Hi @chriselion ,
Did the changes you requested, except the docstring, someone already have a tracking issue in jetbrains
https://youtrack.jetbrains.com/issue/PY-26281

This is what happens in my case:
image

ml-agents/mlagents/trainers/stats.py
```diff
@@ -95,8 +108,8 @@ def write_stats(
     ) -> None:
         for val, stats_summary in values.items():
             set_gauge(
-                GaugeWriter.sanitize_string(f"{category}.{val}.mean"),
-                float(stats_summary.mean),
+                GaugeWriter.sanitize_string(f"{category}.{val}.aggregated_value"),
```
chriselion, line 111: This is going to break some of our internal processes (that live outside of this repo).

Can you make it:
```
set_gauge(GaugeWriter.sanitize_string(f"{category}.{val}.mean"), float(stats_summary.mean))
set_gauge(GaugeWriter.sanitize_string(f"{category}.{val}.sum"), float(stats_summary.sum))
```
instead?

ml-agents/mlagents/trainers/stats.py
chriselion: Yeah, I'd like you to remove it. You'll be able to add your own StatsWriter that gets called soon.

ml-agents/mlagents/trainers/stats.py
```diff
@@ -128,11 +141,12 @@ def write_stats(
 
             log_info.append(f"Mean Reward: {stats_summary.mean:0.3f}")
             log_info.append(f"Std of Reward: {stats_summary.std:0.3f}")
+            log_info.append(f"Num of Reward: {stats_summary.num:0.3f}")
             log_info.append(is_training)
 
             if self.self_play and "Self-play/ELO" in values:
                 elo_stats = values["Self-play/ELO"]
-                log_info.append(f"ELO: {elo_stats.mean:0.3f}")
+                log_info.append(f"ELO: {elo_stats.aggregated_value:0.3f}")
```
chriselion, line 149: should stay as `mean`

ml-agents/mlagents/trainers/stats.py
```diff
@@ -279,14 +313,19 @@ def write_stats(self, step: int) -> None:
 
     def get_stats_summaries(self, key: str) -> StatsSummary:
         """
-        Get the mean, std, and count of a particular statistic, since last write.
+        Get the mean, std, count, sum and aggregation method of a particular statistic, since last write.
+
         :param key: The type of statistic, e.g. Environment/Reward.
-        :returns: A StatsSummary NamedTuple containing (mean, std, count).
+        :returns: A StatsSummary NamedTuple containing (mean, std, sum, count, aggregation).
```
chriselion, line 319:
```suggestion
        :returns: A StatsSummary containing summary statistics.
```
(just make this a bit more future-proof)

ml-agents/mlagents/trainers/trainer/rl_trainer.py
```diff
@@ -100,6 +100,7 @@ def _clear_update_buffer(self) -> None:
     def _is_ready_update(self):
         """
         Returns whether or not the trainer has enough elements to run update model
+
```
chriselion, line 103: Please revert this file.

ml-agents/mlagents/trainers/torch/components/reward_providers/gail_reward_provider.py
```diff
@@ -128,6 +128,7 @@ def compute_estimate(
         """
         Given a mini_batch, computes the estimate (How much the discriminator believes
         the data was sampled from the demonstration data).
+
```
chriselion, line 131: Please revert this file.

Project/Assets/ML-Agents/Examples/Hallway/Scripts/HallwayAgent.cs
```diff
@@ -18,13 +18,15 @@ public class HallwayAgent : Agent
...
     int m_Selection;
+    StatsRecorder statsRecorder;
 
     public override void Initialize()
     {
         m_HallwaySettings = FindObjectOfType<HallwaySettings>();
         m_AgentRb = GetComponent<Rigidbody>();
         m_GroundRenderer = ground.GetComponent<Renderer>();
         m_GroundMaterial = m_GroundRenderer.material;
+        statsRecorder = Academy.Instance.StatsRecorder;
```
chriselion, line 29:
```suggestion
        m_StatsRecorder = Academy.Instance.StatsRecorder;
```
The Unity coding standard for says this should be prefixed with `m_` and `PascalCase`'d. There's no official public mirror, but [here's a copy](https://github.com/needle-mirror/com.unity.polybrush/blob/755cb3b982de374af4f2e2c1bc826689810cf1d1/CodeStyleGuide.cs~#L156-L157) someone scraped from a package:

chriselion:
Thanks, I think it's almost there! I left a few final comments but otherwise it looks pretty good.

Sorry to keep harping on the newlines, but I don't think a bug in pycharm's display is a worthwhile reason to change. I'm OK with them in the files that you have other changes in, but files where those are the only changes should be reverted (I can do this for you in git, as long as you have the setting on the PR to allow repo owners to push changes).

brccabral:
All done, @chriselion

ml-agents/mlagents/trainers/stats.py
chriselion: I removed this (and updated the tests)

ml-agents/mlagents/trainers/agent_processor.py
```diff
@@ -112,7 +113,12 @@ def add_experiences(
                     )
 
     def _process_step(
-        self, step: Union[TerminalStep, DecisionStep], global_id: str, index: int
+        self,
+        step: Union[
+            TerminalStep, DecisionStep
+        ],  # pylint: disable=unsubscriptable-object
```
chriselion, line 119: See https://github.com/PyCQA/pylint/issues/2377 (*sigh*)

ml-agents/mlagents/trainers/stats.py
```diff
@@ -37,14 +40,23 @@ def _dict_to_str(param_dict: Dict[str, Any], num_tabs: int) -> str:
         )
 
 
-class StatsSummary(NamedTuple):
+class StatsSummary(NamedTuple):  # pylint: disable=inherit-non-class
```
chriselion, line 43: See https://github.com/PyCQA/pylint/issues/3876 (*sigh*)

chriselion (approved):

PR #4817, opened 2021-01-04
ignore yamato dir for pytest checks

chriselion:
### Proposed change(s)
Small testing improvement - the scripts under `tests/yamato` are only run by yamato (as the name implies) so we don't need to run pytest when they change. This would save some pytest runs on e.g. https://github.com/Unity-Technologies/ml-agents/pull/4813

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -9,6 +9,8 @@ on:
     - 'test_constraints*.txt'
     - 'test_requirements.txt'
     - '.github/workflows/pytest.yml'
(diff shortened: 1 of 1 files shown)


chriselion:
"you may only define one of `paths` and `paths-ignore` for a single event" from https://github.com/Unity-Technologies/ml-agents/actions/runs/461597735

So guess we won't be doing this...

PR #4821, opened 2021-01-05
Update cattrs dependencies to support python3.9

chriselion:
### Proposed change(s)
AFAICT there's no version of cattrs that works with both python 3.6 and 3.9. So we'll have to require <1.1.0 for python 3.6, and >=1.1.0 for other versions. We'll continue to test all versions so we should spot any incompatibilities quickly.

Also, some tests were failing when setting up `AgentInfoProto`s for demonstrations, because the arrays for agent masks were `np.float64`s instead of `np.bool`s. I'm not sure why this was only manifesting on python3.9, though.

Note that this doesn't actually guarantee python 3.9 support, but it at least unblocks it more that it currently is.

https://forum.unity.com/threads/installation-requires-wrong-version-of-cattrs-macos.1033111/
https://jira.unity3d.com/browse/MLA-1661

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


ml-agents-envs/mlagents_envs/tests/test_envs.py
```diff
@@ -45,6 +45,7 @@ def test_port_defaults(
     )
     env = UnityEnvironment(file_name=file_name, worker_id=0, base_port=base_port)
     assert expected == env._port
+    env.close()
```
chriselion, line 48: These weren't an actual problem, but cause noisy errors about closed handles when the tests shut down.

dongruoping (approved):

PR #4831, opened 2021-01-07
Switching to linux for proto gen, fix output paths

chriselion:
### Proposed change(s)
* Change to linux image
* Fix artifact paths - previously, we'd enter `protobuf-definitions` but not leave it before creating the `artifacts/` directory, so the yamato job would be looking in the wrong place for it.


surfnerd (approved):
LGTM, thanks.

awjuliani (approved):

PR #4832, opened 2021-01-08
Add generated files to test trigger

chriselion:
### Proposed change(s)
Old issue that I forgot about until making modifications to the proto test yesterday.

The original issue was caused by accidentally modifying the generated files (via new pre-commit hook) - this should have triggered a run of the proto test (thus catching the unintended changes) but didn't. The solution is to also trigger when the generated .cs or .py* files have changed.

https://jira.unity3d.com/browse/MLA-1215

### Types of change(s)

- [x] Bug fix

### Other comments
Will manually test in another branch (need to remove the condition on this yaml file)

### DIFF
--- a/.yamato/protobuf-generation-test.yml
+++ b/.yamato/protobuf-generation-test.yml
@@ -32,6 +32,8 @@ test_linux_protobuf_generation:
       pull_request.target match "release.+") AND
       NOT pull_request.draft AND
       (pull_request.changes.any match "protobuf-definitions/**" OR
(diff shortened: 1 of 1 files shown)


chriselion:
On https://github.com/Unity-Technologies/ml-agents/pull/4833 and https://github.com/Unity-Technologies/ml-agents/pull/4834, you can see the
```
remove yaml condition (shouldn't run test)    cd399bd
```
doesn't triggers the test, but the subsequent commit that modifies the generated file does trigger them.

surfnerd (approved):

PR #4448, opened 2020-09-02
[feature] Enable default settings for TrainerSettings

ervteng:
### Proposed change(s)

Based on feedback from CarryCastle. Enables specifying a default TrainerSettings for all behaviors. The default is specified using the same structure as the others under a new header `default_settings`. These are now used as the default values for all behaviors, including those not found in the settings YAML. 

```
default_settings:
    trainer_type: ppo
    ... etc.
behaviors:
    CrawlerStatic:
```

For behavior names that are defined in the YAML, any missing fields will be filled first from the `default_settings`, then those defined in the TrainerSettings class in`settings.py`. 

WIP: Tests, documentation

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist
- [x] Added tests that prove my fix is effective or that my feature works
- [x] Updated the [changelog](https://github.com/Unity-Technologies/ml-agents/blob/master/com.unity.ml-agents/CHANGELOG.md) (if applicable)
- [x] Updated the [documentation](https://github.com/Unity-Technologies/ml-agents/tree/master/docs) (if applicable)
- [ ] Updated the [migration guide](https://github.com/Unity-Technologies/ml-agents/blob/master/docs/Migrating.md) (if applicable)

PR #4842, opened 2021-01-11
make sure DefaultTrainerDict is pickle-able

chriselion:
### Proposed change(s)
In the current form, the DefaultTrainerDict can't be deserialized from pickle (and thus can't be passed to a worker process from the subprocess module, which I'm trying to do in https://github.com/Unity-Technologies/ml-agents/pull/4780). It appears that the initializer needs to be able to take both forms of arguments; I'm not sure if there's a cleaner way to do this.

Added in https://github.com/Unity-Technologies/ml-agents/pull/4448

### Types of change(s)
- [x] Bug fix

### Checklist

### DIFF
--- a/ml-agents/mlagents/trainers/settings.py
+++ b/ml-agents/mlagents/trainers/settings.py
@@ -682,7 +682,8 @@ class TrainerSettings(ExportableSettings):
     class DefaultTrainerDict(collections.defaultdict):
         def __init__(self, *args):
             # Depending on how this is called, args may have the defaultdict
-            # callable at the start of the list or not.
+            # callable at the start of the list or not. In particular, unpickling
+            # will pass [TrainerSettings].
             if args and args[0] == TrainerSettings:
                 super().__init__(*args)
             else:
--- a/ml-agents/mlagents/trainers/settings.py
+++ b/ml-agents/mlagents/trainers/settings.py
@@ -681,6 +681,8 @@ class TrainerSettings(ExportableSettings):
 
     class DefaultTrainerDict(collections.defaultdict):
         def __init__(self, *args):
+            # Depending on how this is called, args may have the defaultdict
+            # callable at the start of the list or not.
             if args and args[0] == TrainerSettings:
                 super().__init__(*args)
             else:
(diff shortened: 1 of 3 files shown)


chriselion:
Without the fix, the added test fails:
```python
    def test_pickle():
        # Make sure RunOptions is pickle-able.
        run_options = RunOptions()
        p = pickle.dumps(run_options)
>       pickle.loads(p)

ml-agents/mlagents/trainers/tests/test_settings.py:526: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = DefaultTrainerDict(<class 'mlagents.trainers.settings.TrainerSettings'>, {})
args = (<class 'mlagents.trainers.settings.TrainerSettings'>,)

    def __init__(self, *args):
>       super().__init__(TrainerSettings, *args)
E       TypeError: 'type' object is not iterable

ml-agents/mlagents/trainers/settings.py:684: TypeError
```

ervteng (approved):

ml-agents/mlagents/trainers/settings.py
```diff
@@ -681,7 +681,12 @@ def structure(d: Mapping, t: type) -> Any:
 
     class DefaultTrainerDict(collections.defaultdict):
         def __init__(self, *args):
-            super().__init__(TrainerSettings, *args)
+            # Depending on how this is called, args may have the defaultdict
+            # callable at the start of the list or not.
```
ervteng, line 685: Maybe add a note about how it addresses an issue with pickle load for posterity

chriselion:
only change between the last commit and the previous one was adding a comment, so going to merge without tests finishing.

ml-agents/mlagents/trainers/settings.py
chriselion: Done

PR #4846, opened 2021-01-12
Add default init/gain to LinearEncoder

andrewcoh:
### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/mlagents/trainers/torch/attention.py
+++ b/ml-agents/mlagents/trainers/torch/attention.py
@@ -95,6 +95,7 @@ class EntityEmbeddings(torch.nn.Module):
         # If not concatenating self, input to encoder is just entity size
         if not concat_self:
             self.self_size = 0
+        # Initialization scheme from http://www.cs.toronto.edu/~mvolkovs/ICML2020_tfixup.pdf
         self.ent_encoders = torch.nn.ModuleList(
             [
                 LinearEncoder(
@@ -165,6 +166,7 @@ class ResidualSelfAttention(torch.nn.Module):
             num_heads=num_heads, embedding_size=embedding_size
         )
 
+        # Initialization scheme from http://www.cs.toronto.edu/~mvolkovs/ICML2020_tfixup.pdf
         self.fc_q = linear_layer(
             embedding_size,
             embedding_size,
--- a/ml-agents/mlagents/trainers/torch/attention.py
+++ b/ml-agents/mlagents/trainers/torch/attention.py
@@ -97,7 +97,13 @@ class EntityEmbeddings(torch.nn.Module):
             self.self_size = 0
         self.ent_encoders = torch.nn.ModuleList(
             [
-                LinearEncoder(self.self_size + ent_size, 1, embedding_size)
+                LinearEncoder(
+                    self.self_size + ent_size,
+                    1,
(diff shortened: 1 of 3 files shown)


ml-agents/mlagents/trainers/torch/attention.py
```diff
@@ -97,7 +97,13 @@ def __init__(
             self.self_size = 0
         self.ent_encoders = torch.nn.ModuleList(
             [
-                LinearEncoder(self.self_size + ent_size, 1, embedding_size)
+                LinearEncoder(
+                    self.self_size + ent_size,
+                    1,
+                    embedding_size,
+                    kernel_init=Initialization.Normal,
+                    kernel_gain=(0.125 / embedding_size) ** 0.5,
```
chriselion, line 105: Is there a reference how you picked this gain? Could you add a comment and/or link?

andrewcoh: I'm following the initialization scheme from http://www.cs.toronto.edu/~mvolkovs/ICML2020_tfixup.pdf and gain from https://github.com/openai/multi-agent-emergence-environments/blob/master/ma_policy/layers.py#L89

ml-agents/mlagents/trainers/torch/layers.py
```diff
@@ -120,14 +120,21 @@ class LinearEncoder(torch.nn.Module):
...
     """
 
-    def __init__(self, input_size: int, num_layers: int, hidden_size: int):
+    def __init__(
+        self,
+        input_size: int,
+        num_layers: int,
+        hidden_size: int,
+        kernel_init: Initialization = Initialization.KaimingHeNormal,
+        kernel_gain: float = 1.0,
```
chriselion, line 129: Would it make sense to default this to None and than have something like
```
if kernel_gain is None:
  if kernel_init == Initialization.KaimingHeNormal:
    kernel_gain = 1.0
  elif kernel_init == Initialization.Normal:
    kernel_gain = (0.125 / hidden_size) ** 0.5
  else:
    kernel_gain = 1.0
```
?

chriselion: Either that or have static methods to create a LinearEncoder that have different gain defaults?

vincentpierre (approved):

ml-agents/mlagents/trainers/torch/attention.py
andrewcoh: Added the reference

ml-agents/mlagents/trainers/torch/layers.py
andrewcoh: it might make sense to put that in `linear_layer`. Personally, my take is to just provide the args since anyone playing with them is probably an advanced user but I am open to this.

Thoughts @vincentpierre @ervteng

chriselion: I don't have strong feelings on it. Maybe better to wait until we have >2 data points to extrapolate from.

chriselion (approved):

PR #4849, opened 2021-01-12
Initial implementation using IHeuristicProvider.

surfnerd:
### Proposed change(s)

An initial implementation using IHeuristicProvider which allows both IActuators and Agent to implement the Heuristic function in order to generate actions for behaviors.  Agents will have their heuristic method called if their BrainParameters are set up that way.  Otherwise, any IActuator that implements IHeuristic will have its Heuristic method called on it with the appropriate ActionBuffers for that IActuator.

### Types of change(s)

- [x] Bug fix
- [x] New feature
- [x] Code refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Other (please describe)

### Checklist


com.unity.ml-agents/Runtime/Actuators/ActionSpec.cs
```diff
@@ -29,18 +29,18 @@ public struct ActionSpec
         /// <summary>
         /// The number of continuous actions that an Agent can take.
         /// </summary>
-        public int NumContinuousActions { get { return m_NumContinuousActions; } set { m_NumContinuousActions = value; } }
+        public int NumContinuousActions { get => m_NumContinuousActions; set => m_NumContinuousActions = value; }
 
         /// <summary>
```
surfnerd, line 34: i'll revert these changes

com.unity.ml-agents/Runtime/Agent.cs
```diff
@@ -995,7 +1003,12 @@ void InitializeActuators()
             var param = m_PolicyFactory.BrainParameters;
             m_VectorActuator = new VectorActuator(this, param.ActionSpec);
             m_ActuatorManager = new ActuatorManager(attachedActuators.Length + 1);
-            m_LegacyActionCache = new float[m_VectorActuator.TotalNumberOfActions()];
+#pragma warning disable 618
+            m_LegacyActionCache = new float[param.VectorActionSpaceType == SpaceType.Continuous
```
chriselion, line 1007: I think the old logic (`m_VectorActuator.TotalNumberOfActions()]`) was OK, since the vectorActuator should be sized to support only the BrainParameters (not the other actuators on the agent).

I'd also be OK allocating these lazily where they're needed.

com.unity.ml-agents/Runtime/Agent.cs
```diff
@@ -1241,11 +1254,18 @@ public virtual void OnActionReceived(ActionBuffers actions)
...
             {
-                m_LegacyActionCache = actions.ContinuousActions.Array;
+                Array.Copy(actions.ContinuousActions.Array,
+                    m_LegacyActionCache,
+                    actionSpec.NumContinuousActions);
             }
             else
             {
-                m_LegacyActionCache = Array.ConvertAll(actions.DiscreteActions.Array, x => (float)x);
+                var discreteArray = Array.ConvertAll(actions.DiscreteActions.Array, x => (float)x);
```
chriselion, line 1263: This will still allocated a new array each step, right?

com.unity.ml-agents/Runtime/Agent.cs
```diff
@@ -889,27 +894,30 @@ public virtual void Heuristic(in ActionBuffers actionsOut)
...
-                        discreteActionSegment[i] = (int)convertedOut[i];
+                        discreteActionSegment[i] = (int)m_LegacyHeuristicCache[i];
                     }
                     actionsOut.ContinuousActions.Clear();
                     break;
             }
 #pragma warning restore CS0618
 
+            // Send heuristic buffers to actuators if they implement IHeuristic
+            m_ActuatorManager.ExecuteHeuristic();
```
chriselion, line 920: Discussed offline, but I think the right caller for m_ActuatorManager.ExecuteHeuristic is the HeuristicPolicy itself. 

Actually, as it's written, this might recurse infinitely for Agents that don't override `Heuristic(ActionBuffers)` - HeuristicPolicy calls Agent.Heuristic, Agent.Heuristic calls ActuatorManager.ExecuteHeuristic(), and ActuatorManager.ExecuteHeuristic() calls Agent.Heuristic.

com.unity.ml-agents/Runtime/Agent.cs
surfnerd: yeah

com.unity.ml-agents/Runtime/Agent.cs
surfnerd: Makes sense, I'll update the flow.

com.unity.ml-agents/Runtime/Agent.cs
surfnerd: I'll convert this back

com.unity.ml-agents/Runtime/Agent.cs
chriselion: So why copy? I would say either
* Keep the old behavior
* or convert when writing to m_LegacyActionCache, e.g. `m_LegacyActionCache[i] = (float) actions.DiscreteActions.Array[i]`

Personally, I'd lean towards not optimizing deprecated code, and just go with what's simplest/easiest to maintain.

com.unity.ml-agents/Runtime/Actuators/IHeuristicProvider.cs
```diff
@@ -0,0 +1,18 @@
+namespace Unity.MLAgents.Actuators
+{
+    /// <summary>
+    /// Interface that allows objects to fill out an <see cref="ActionBuffers"/> data structure for controlling
+    /// behavior of Agents.
```
chriselion, line 5:
```suggestion
    /// behavior of Agents or Actuators.
```

com.unity.ml-agents/Runtime/Actuators/VectorActuator.cs
```diff
@@ -55,6 +73,11 @@ public void OnActionReceived(ActionBuffers actionBuffers)
             m_ActionReceiver.OnActionReceived(ActionBuffers);
         }
 
+        public void Heuristic(in ActionBuffers actionBuffersOut)
+        {
+            m_HeuristicProvider.Heuristic(actionBuffersOut);
```
chriselion, line 78:
```suggestion
            m_HeuristicProvider?.Heuristic(actionBuffersOut);
```
since `actionReceiver as IHeuristicProvider` could result in null?

surfnerd: huh, had that in there before. not sure why I removed it.

com.unity.ml-agents/Runtime/Agent.cs
```diff
@@ -889,27 +894,27 @@ public virtual void Heuristic(in ActionBuffers actionsOut)
             // Disable deprecation warnings so we can call the legacy overload.
 #pragma warning disable CS0618
 
+            Array.Clear(m_LegacyHeuristicCache, 0, m_LegacyHeuristicCache.Length);
```
chriselion, line 897: I think it's reasonable that the values are persistent across steps. Discussed changing this in https://jira.unity3d.com/browse/MLA-1129

surfnerd: ok, i can take it out

com.unity.ml-agents.extensions/Runtime/Match3/Match3ActuatorComponent.cs
```diff
@@ -5,15 +5,15 @@
 namespace Unity.MLAgents.Extensions.Match3
 {
     /// <summary>
-    /// Actuator component for a Match 3 game. Generates a Match3Actuator at runtime.
+    /// Actuator component for a Board game. Generates a Match3Actuator at runtime.
```
chriselion, line 8: revert these?

com.unity.ml-agents/Runtime/Actuators/ActuatorManager.cs
```diff
@@ -204,6 +204,49 @@ public void WriteActionMask()
             }
         }
 
+        /// <summary>
+        /// Iterates through all of the IActuators in this list and calls their
+        /// <see cref="IHeuristicProvider.Heuristic"/> method on them, if implemented, with the appropriate
+        /// <see cref="ActionSegment{T}"/>s depending on their <see cref="ActionSpec"/>.
+        /// </summary>
+        public void Heuristic(in ActionBuffers actionBuffersOut)
```
chriselion, line 212: nit: ApplyHeuristics()?

com.unity.ml-agents/Runtime/Agent.cs
```diff
@@ -826,11 +831,11 @@ void ResetData()
         public virtual void Initialize() { }
 
         /// <summary>
-        /// Implement `Heuristic()` to choose an action for this agent using a custom heuristic.
+        /// Implement <see cref="IHeuristicProvider"/> to choose an action for this agent using a custom heuristic.
```
chriselion, line 834: I think this is OK referring to Agent.Heuristic since it's talking specifically about the Agent. Same below in the comments on OnActionReceived()

com.unity.ml-agents.extensions/Runtime/Match3/Match3Actuator.cs
```diff
@@ -5,16 +5,31 @@
 
 namespace Unity.MLAgents.Extensions.Match3
 {
+    public enum HeuristicQuality
```
chriselion, line 8: I don't think we need this in the base class, since by default it will have no effect - it switches between hardcoded 1.0 and virtual method that returns 1.0.

chriselion: (and just use `var movePoints = EvalMovePoints(move);` below)

com.unity.ml-agents.extensions/Runtime/Match3/AbstractBoard.cs
```diff
@@ -27,6 +27,11 @@ public abstract class AbstractBoard : MonoBehaviour
         /// </summary>
         public int NumSpecialTypes;
 
+        /// <summary>
+        /// A random seed used to generate a board, if needed.
+        /// </summary>
+        public int RandomSeed = -1;
```
chriselion, line 33: Better on the ActuatorComponent instead? That's the only thing that uses it, right?

chriselion (approved):
A few nits but looks good. Just want to make sure that if we're not going to rename the Match3 classes that the related changes get reverted too.

com.unity.ml-agents.extensions/Runtime/Match3/Match3Actuator.cs
surfnerd: good call

chriselion (approved):

PR #4850, opened 2021-01-13
Team manager prototype 

dongruoping:
### Proposed change(s)

Merge for prototype testing.
Interfaces may change in future

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


com.unity.ml-agents/Runtime/Agent.cs
```diff
@@ -560,6 +580,13 @@ void NotifyAgentDone(DoneReason doneReason)
             m_Info.storedActions.Clear();
         }
 
+        public void SendDoneToTrainer()
```
chriselion, line 583: internal

com.unity.ml-agents/Runtime/Agent.cs
```diff
@@ -50,6 +50,11 @@ internal struct AgentInfo
         /// </summary>
         public int episodeId;
 
+        /// <summary>
+        /// Team Manager identifier.
+        /// </summary>
+        public string teamManagerId;
```
chriselion, line 56: This shouldn't be public. I think a better approrach is to add an accessor for m_TeamManager, and add a method to the interface for an ID.

ml-agents-envs/mlagents_envs/base_env.py
```diff
@@ -120,11 +122,15 @@ def __getitem__(self, agent_id: AgentId) -> DecisionStep:
             agent_mask = []
             for mask in self.action_mask:
                 agent_mask.append(mask[agent_index])
+        team_manager_id = None
+        if self.team_manager_id is not None and self.team_manager_id != "":
```
chriselion, line 126: nit: `if not self.team_manager_id` should cover both cases.

com.unity.ml-agents.extensions/Runtime/Teams/BaseTeamManager.cs
```diff
@@ -0,0 +1,35 @@
+using System.Collections.Generic;
+using Unity.MLAgents;
+using Unity.MLAgents.Sensors;
+
+namespace Unity.MLAgents.Extensions.Teams
+{
+    public class BaseTeamManager : ITeamManager
+    {
+        readonly string m_Id = System.Guid.NewGuid().ToString();
```
chriselion, line 9: You could also make this an integer - have a global static ID and increment with `[Interlocked.Increment](https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.increment?view=net-5.0)` (which is threadsafe)

PR #4865, opened 2021-01-15
Run pytest on GPU

dongruoping:
### Proposed change(s)

* Make python unit tests device-friendly

* Add daily trigger to run pytest on GPU

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


ml-agents/mlagents/trainers/tests/torch/test_simple_rl.py
```diff
@@ -39,13 +39,15 @@
 SAC_TORCH_CONFIG = sac_dummy_config()
 
 
+@pytest.mark.check_environment_trains
```
dongruoping, line 42: Mark and disable the simple_rl tests since they broken easily and not the main purpose of the GPU tests.
The rest of the tests should be able to capture the non-contiguous errors and device errors.

.yamato/pytest-gpu.yml
```diff
@@ -0,0 +1,21 @@
...
+    image: package-ci/ubuntu:stable
+    flavor: b1.large
+  commands:
+    - |
+      sudo apt-get update && sudo apt-get install -y python3-venv
+      python3 -m venv venv && source venv/bin/activate
+      python3 -m pip install pyyaml --index-url https://artifactory.prd.it.unity3d.com/artifactory/api/pypi/pypi/simple
+      python3 -u -m ml-agents.tests.yamato.setup_venv
+      python3 -m pip install --progress-bar=off -r test_requirements.txt --index-url https://artifactory.prd.it.unity3d.com/artifactory/api/pypi/pypi/simple
+      python3 -m pip install torch==1.7.1+cu101 torchvision==0.8.2+cu101 torchaudio==0.7.2 -f https://download.pytorch.org/whl/torch_stable.html --index-url https://artifactory.prd.it.unity3d.com/artifactory/api/pypi/pypi/simple
```
chriselion, line 14:
```suggestion
      python3 -m pip install torch==1.7.1+cu101 torchvision==0.8.2+cu101 torchaudio==0.7.2 -f https://download.pytorch.org/whl/torch_stable.html
```
Don't think this is necessary since you're already specifying the download.pytorch.org URL

.yamato/pytest-gpu.yml
```diff
@@ -0,0 +1,21 @@
...
+      python3 -m pytest -m "not check_environment_trains" --cov=ml-agents --cov=ml-agents-envs --cov=gym-unity --cov-report html --junitxml=junit/test-results-${{ matrix.python-version }}.xml -p no:warnings
```
chriselion, line 15:
```suggestion
      python3 -m pytest -m "not check_environment_trains" --junitxml=junit/test-results.xml -p no:warnings
```
Don't need the coverage options (since we're only running a subset and don't want to fail for low coverage) and `matrix.python-version` is part of the github action definition.

chriselion: Might want to make `junit` an artifact directory so the file is saved (not sure it'll be useful though).

I think something like this is all you need https://github.com/brccabral/ml-agents/blob/efa55731c689e6abad062a64509c51be6f936351/.yamato/training-int-tests.yml#L33-L36

chriselion (approved):
Looks good! Can you add a `pytest.ini` file with this in it?
```
[pytest]
addopts = --strict-markers
markers =
    check_environment_trains: Slow training tests, don't run on yamato
```
(see https://docs.pytest.org/en/stable/mark.html#raising-errors-on-unknown-marks) just to defend against future typos?

.yamato/pytest-gpu.yml
dongruoping: Without this it shows a bunch of Retry warnings with connection errors, so it actually makes difference. .
It still install successfully at the end but I think we better keep it.

PR #4867, opened 2021-01-16
Don't clamp timescale to 100 outside of editor

chriselion:
### Proposed change(s)
Remove the upper limit of 100 for the timescale when using the player. We can't remove the limit entirely, since the Editor still requires it.

Note that having a lower bound of 1 is still worthwhile, so we clamp to [1, Infinity] in the player for simplicity.

Also made a SideChannel tests directory, since I have another one to add in another PR.

https://jira.unity3d.com/browse/MLA-748
First noted in https://github.com/Unity-Technologies/ml-agents/pull/3596
[Editor code](https://github.cds.internal.unity3d.com/unity/unity/blob/f37d13a76f487037a69975b580497b251bdcabb6/Runtime/Input/TimeManager.cpp#L561-L584)

### Types of change(s)
- [x] QoL improvement

### Checklist


com.unity.ml-agents/Runtime/SideChannels/EngineConfigurationChannel.cs
```diff
@@ -46,7 +46,15 @@ protected override void OnMessageReceived(IncomingMessage msg)
...
-                    timeScale = Mathf.Clamp(timeScale, 1, 100);
+
+                    // There's an upper limit for the timeScale in the editor (but not in the player)
+                    // Always ensure that timeScale >= 1 also,
+#if UNITY_EDITOR
+                    const float maxTimeScale = 100f;
+#else
+                    const cloat maxTimeScale = float.PositiveInfinity;
+#endif
+                    timeScale = Mathf.Clamp(timeScale, 1, maxTimeScale);
```
surfnerd, line 57: Could there be cases where someone wants to train at timescales between (0,1]?

surfnerd: and do we want to support that?

awjuliani: I can't imagine for training, but there could hypothetically be an inference use-case where someone might want it to go slower to better inspect the agent behavior.

awjuliani (approved):
Looks good to me. I wasn't aware of the 100 limit in the editor, but I imagine it was kept in place to prevent locking up the editor when in play mode.

awjuliani:
I wonder if we have an environment which would benefit from the higher timescale. It would likely have to have neither physics nor rendering, so perhaps Match3?

com.unity.ml-agents/Runtime/SideChannels/EngineConfigurationChannel.cs
chriselion: This block of code only affects timescale information coming from python (generally `mlagents-learn` commandline, but could be the yaml config or low-level API too), so inference isn't affected by this.

So far there haven't been any requests for timescale < 1, but we can change the lower bound to, say, 0.01.

chriselion:
> I wonder if we have an environment which would benefit from the higher timescale. It would likely have to have neither physics nor rendering, so perhaps Match3?

AFAIK we don't override the defaults for training in any scene.

I'm not sure Match3 would benefit since it doesn't have any physics or animation; moves are requested on every step.

surfnerd (approved):

PR #4878, opened 2021-01-21
Track histogram of environment reward

awjuliani:
### Proposed change(s)

Tracks the episodic reward as a histogram in addition to a scalar in tensorboard. This allows the multi-modality of rewards to be more apparent to end-users. 

Below is an example of such a graph generated from a PushBlock training run. 

Screen Shot 2021-01-21 at 3 44 38 PM

Here is the learning curve from FoodCollector:

Screen Shot 2021-01-21 at 4 11 42 PM

Not sure that my current solution is the most optimal given how the StatsSummary and StatsWriters are current designed. Open to feedback. 

### Types of change(s)

- [ ] Bug fix
- [X] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


ml-agents/mlagents/trainers/stats.py
```diff
@@ -325,10 +332,15 @@ def get_stats_summaries(self, key: str) -> StatsSummary:
         if len(stat_values) == 0:
             return StatsSummary.empty()
 
+        if key == "Environment/Cumulative Reward":
```
chriselion, line 335: I'd say just always set `full = np.array(stat_values)`. This will play nicely with https://github.com/Unity-Technologies/ml-agents/pull/4788 (if I ever get time to finish it).

ml-agents/mlagents/trainers/stats.py
```diff
@@ -330,5 +335,6 @@ def get_stats_summaries(self, key: str) -> StatsSummary:
             std=np.std(stat_values),
             num=len(stat_values),
             sum=np.sum(stat_values),
+            full_dist=stat_values,
```
ervteng, line 338: This might be out of the scope of this PR, but I wonder if since we're feeding the entire list of `stat_values`, we can just compute `std`, `num`, `sum`, and `mean` as `@property` values inside the `StatsSummary`? e.g.
```
class StatssSummary(NamedTuple):
    full_dist: List[float]
    @property
    def mean(self):
        return mean(self.full_dist)
```

awjuliani: This makes a lot of sense to me. The way the tests are all currently written involves a lot of hand-coded mean/std values though. I think it might necessitate a slightly bigger refactor to rework the tests to deal with the underlying values rather than their summary stats.

awjuliani: I went ahead and made this change. I think it cleans up things nicely.

ml-agents/mlagents/trainers/stats.py
```diff
@@ -58,6 +55,22 @@ def aggregated_value(self):
...
+    def mean(self):
+        return np.mean(self.full_dist)
+
+    @property
+    def std(self):
+        return np.std(self.full_dist)
+
+    @property
+    def num(self):
+        return len(self.full_dist)
```
chriselion, line 68: `StatsSummary.empty().num` will return 1 now.

awjuliani: My first thought is to use an empty list for `full_dist` when calling `empty()`. That means the other stats will return `NaN` which is technically more correct than returning `0`, thought may break something else. Do you have a different preference?

chriselion:
Can you make it so that user stats can use this too? I think it's pretty small, you'd just need to
* Add a new StatsAggregationMethod enum values (say, HISTOGRAM) on python and C#
* Make sure `Environment/Cumulative Reward` uses the new value
* Check StatsSummary.aggregation_method in TensorBoard writer instead of hard-coding the key

awjuliani:
> Can you make it so that user stats can use this too? I think it's pretty small, you'd just need to
> 
> * Add a new StatsAggregationMethod enum values (say, HISTOGRAM) on python and C#
> * Make sure `Environment/Cumulative Reward` uses the new value
> * Check StatsSummary.aggregation_method in TensorBoard writer instead of hard-coding the key

Made these changes.

ml-agents/mlagents/trainers/stats.py
chriselion: Only other thought is to do e.g. `np.mean(self.full_dist) if self.full_dist) else 0.0` which is basically the old behavior, and same for `np.std` (`np.sum` is safe)

chriselion: But I think this is fine too.

chriselion (approved):

ervteng (approved):
🚢 🇮🇹

PR #4882, opened 2021-01-24
Optimizing test running time

loopylangur:
Hi,

I was able to reduce the running time of `test_2d_sac` test from 70  seconds to about 10 seconds on my local machine by changing some hyper-parameters.

I also ran the test several times to ensure that the test is not flaky.

Environment:
```
python 3.8
torch=1.7.1
numpy==1.19.5
```
Is this something that you guys will be interested in? If yes, then I can help you optimize some other tests in this repo as well. If you have any other suggestions/edits I will be happy to integrate them as well. Please let me know.

Thanks!

### DIFF
--- a/ml-agents/mlagents/trainers/tests/torch/test_simple_rl.py
+++ b/ml-agents/mlagents/trainers/tests/torch/test_simple_rl.py
@@ -142,7 +142,7 @@ def test_2d_sac(action_sizes):
         SAC_TORCH_CONFIG.hyperparameters,
         buffer_init_steps=2000,
         learning_rate=0.023,
-        batch_size=1
+        batch_size=1,
     )
     config = attr.evolve(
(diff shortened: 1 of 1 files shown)


chriselion:
Thanks for trying to improve the test times. I talked it over with some folks for the team, and we're worried that this might end up "overfitting" to the particular seed and network configuration, and make it more brittle to changes in the future. So I think we'll leave the tests as they are for now.

loopylangur:
Hi @chriselion,

Thanks for your response. What do u mean by network configurations here? I can try optimizing the test using different seeds, but I am not sure if there is a mechanism to try out different network configurations?

Would something like this be a more reasonable way to optimize the tests and avoid overfitting?

chriselion:
I mean that the network configuration might change as we refactor things in the code in the future, and make the tests more difficult to pass then.

miguelalonsojr:
Closing PR as stale.

PR #4885, opened 2021-01-27
replace old bot with gh action

chriselion:
### Proposed change(s)
Switch to the https://github.com/dessant/lock-threads github action.

The action has an internal limit of 50 issues per run: 

> To avoid triggering abuse prevention mechanisms on GitHub, only 50 issues and pull requests will be handled at once. If your repository has more than that, it will just take a few hours or days to process them all.

I'll run once manually on master to make sure it works.


.github/workflows/lock.yml
```diff
@@ -0,0 +1,23 @@
...
+    - cron: '0 0 * * *'
+
+jobs:
+  lock:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dessant/lock-threads@v2
+        with:
+          github-token: ${{ github.token }}
+          issue-lock-inactive-days: '365'
```
chriselion, line 14: Will reduce to 30 once the backlog is clear

mmattar (approved):

Issue #4883, opened 2021-01-25
Low level memory allocation every frame


szantner:
**Describe the bug**
I detect seemingly low level memory allocations in the EnvironmentStep call every frame as seen in the screenshot using Deep Profile in Profiler

**To Reproduce**
Using multiple agents in inference mode.

**Console logs / stack traces**

**Screenshots**
![image](https://user-images.githubusercontent.com/170305/105715905-71edce80-5f1e-11eb-9edd-8f2c1a3df9c7.png)

**Environment (please complete the following information):**
- Unity Version: Unity 2020.2.2f1
- OS + version: Windows 10
- _ML-Agents version_: 1.0.6

chriselion:
Hi,
I think you're right that we're allocating a BinaryWriter and MemoryStream on each step, even if there are no messages to process:
https://github.com/Unity-Technologies/ml-agents/blob/30bcb018bcac6dd73ed64d34016131d03cd6514d/com.unity.ml-agents/Runtime/SideChannels/SideChannelManager.cs#L121-L126

We can definitely early out in that case (or maybe reuse the objects). Is that the main allocation you're seeing?

chriselion:
We've got this logged in our internal tracker as MLA-1720. I'll probably get to it this week, but it should be pretty straightforward if you want to submit a PR for it instead.

szantner:
Hi Chris,

Thank you for the quick response. It is probably best if I leave the modifications for you! :) 

Btw - and maybe that should be a separate issue - but I saw a strange memory allocation too in another part of ml_agents, but that I managed to fix on my side. So if I did the following, that caused memory allocations:

`sensor.AddObservation(_observationLadarInfoOffRoad);`

where  `private List<float> _observationLadarInfoOffRoad;` 

When I added the observations in a for loop myself, the memory allocation disappeared:

`
for (int i = 0, length = _observationLadarInfoOffRoad.Count; i < length; i++) {
    sensor.AddObservation(_observationLadarInfoOffRoad[i]);
}
`

chriselion:
I think I see what you're saying. This is happening in your Agent's`CollectObservations(VectorSensor sensor)` method, right? If you use `sensor.AddObservation(_observationLadarInfoOffRoad);` it would treat the List as an IEnumerable:
[...]

PR #4887, opened 2021-01-27
[MLA-1724] Reduce use of IEnumerable during inference

chriselion:
### Proposed change(s)
This makes several changes to decrease the amount of garbage generated during inference. Most of it is from changing IEnumerables to ILists and iterating over them by index, but there are a few others changes:
* Reuse a Dictionary and List in ModelRunner
* Rearrange the ProfileMarkers in ModelRunner to not do any string manipulation
* Add an overload to VectorSensor that uses an IList (we don't actually use either the old or new form in any examples)

For a test scene (3DBall with decisions being requested every step), this reduces the GC Allocation for a typical frame for the ModelRunner.DecideAction from 102.2KB to 74.4KB. The vast majority of the remainder is coming from Barracuda (and will be addressed by them separately).

#### Before
image

#### After
image

Note that tensor generation and application (for continuous actions) now have no garbage generated.

https://jira.unity3d.com/browse/MLA-1724
https://github.com/Unity-Technologies/ml-agents/issues/4883

### Types of change(s)

- [x] Bug fix

### Checklist

### DIFF
--- a/com.unity.ml-agents/Runtime/Inference/ModelRunner.cs
+++ b/com.unity.ml-agents/Runtime/Inference/ModelRunner.cs
@@ -24,12 +24,14 @@ namespace Unity.MLAgents.Inference
         TensorApplier m_TensorApplier;
 
         NNModel m_Model;
+        string m_ModelName;
         InferenceDevice m_InferenceDevice;
         IWorker m_Engine;
         bool m_Verbose = false;
         string[] m_OutputNames;
         IReadOnlyList<TensorProxy> m_InferenceInputs;
-        IReadOnlyList<TensorProxy> m_InferenceOutputs;
+        List<TensorProxy> m_InferenceOutputs;
+        Dictionary<string, Tensor> m_InputsByName;
         Dictionary<int, List<float>> m_Memories = new Dictionary<int, List<float>>();
 
         SensorShapeValidator m_SensorShapeValidator = new SensorShapeValidator();
@@ -56,6 +58,7 @@ namespace Unity.MLAgents.Inference
         {
             Model barracudaModel;
             m_Model = model;
+            m_ModelName = model.name;
             m_InferenceDevice = inferenceDevice;
             m_TensorAllocator = new TensorCachingAllocator();
             if (model != null)
@@ -84,6 +87,8 @@ namespace Unity.MLAgents.Inference
                 seed, m_TensorAllocator, m_Memories, barracudaModel);
             m_TensorApplier = new TensorApplier(
                 actionSpec, seed, m_TensorAllocator, m_Memories, barracudaModel);
+            m_InputsByName = new Dictionary<string, Tensor>();
+            m_InferenceOutputs = new List<TensorProxy>();
         }
 
         public InferenceDevice InferenceDevice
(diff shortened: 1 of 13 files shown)


com.unity.ml-agents/Runtime/Sensors/VectorSensor.cs
```diff
@@ -172,6 +174,18 @@ public void AddObservation(IEnumerable<float> observation)
             }
         }
 
+        /// <summary>
+        /// Adds a list or array of float observations to the vector observations of the agent.
+        /// </summary>
+        /// <param name="observation">Observation.</param>
+        public void AddObservation(IList<float> observation)
```
chriselion, line 181: What are the overload resolution rules in C# for this? If the caller passes an List<float> or float[], how do I know whether the IList<float> or IEnumerable<float> will be chosen? I can give this a different name to make it less ambiguous...

com.unity.ml-agents/Runtime/Sensors/ObservationWriter.cs
```diff
@@ -144,6 +145,27 @@ public void AddRange(IEnumerable<float> data, int writeOffset = 0)
             }
         }
 
+        public void AddList(IList<float> data, int writeOffset = 0)
```
chriselion, line 148: Same comments for overloading resolution, but I made this unambiguous (and marked the old version obsolete).

com.unity.ml-agents/Runtime/Sensors/ObservationWriter.cs
```diff
@@ -122,6 +122,7 @@ internal void SetTarget(TensorProxy tensorProxy, int batchIndex, int channelOffs
         /// </summary>
         /// <param name="data"></param>
         /// <param name="writeOffset">Optional write offset.</param>
+        [Obsolete("Use AddList() for better performance")]
```
chriselion, line 125: This is in the public API so we can't remove it, but we can strongly steer users to AddList instead.

com.unity.ml-agents/Runtime/Inference/ModelRunner.cs
```diff
@@ -169,31 +171,33 @@ public void DecideBatch()
             }
 
             Profiler.BeginSample("ModelRunner.DecideAction");
+            Profiler.BeginSample(m_ModelName);
 
-            Profiler.BeginSample($"MLAgents.{m_Model.name}.GenerateTensors");
```
chriselion, line 173: These were allocating for the string concatenation, and `m_Model.name` also allocates, so cache it at the start.

com.unity.ml-agents/Runtime/Inference/ModelRunner.cs
```diff
@@ -96,15 +101,14 @@ public NNModel Model
             get { return m_Model; }
         }
 
-        static Dictionary<string, Tensor> PrepareBarracudaInputs(IEnumerable<TensorProxy> infInputs)
+        void PrepareBarracudaInputs(IReadOnlyList<TensorProxy> infInputs)
         {
-            var inputs = new Dictionary<string, Tensor>();
-            foreach (var inp in infInputs)
+            m_InputsByName.Clear();
```
chriselion, line 106: Changed this and FetchBarracudaOutputs to reuse their results between frames.

surfnerd (approved):

surfnerd (approved):

Issue #4876, opened 2021-01-21
[help-wanted ]Problems with using GPU


torsilver:
When running the sample program "Pyramids"
I set Agent's Inference Device as GPU
But MLAgents - Learn seems to call the CPU as well.
I have installed PyTorch1.7.1 and Cuda11.2
Is there a problem with my version of software?

pip list
Package                Version
---------------------- -----------
absl-py                0.11.0
attrs                  20.3.0
cached-property        1.5.2
cachetools             4.2.0
cattrs                 1.0.0
certifi                2020.12.5
chardet                4.0.0
cloudpickle            1.6.0
google-auth            1.24.0
google-auth-oauthlib   0.4.2
grpcio                 1.34.1
h5py                   3.1.0
idna                   2.10
importlib-metadata     3.4.0
lxml                   4.6.2
Markdown               3.3.3
mlagents               0.23.0
mlagents-envs          0.23.0
numpy                  1.18.5
oauthlib               3.1.0
opcua                  0.98.12
Pillow                 8.1.0
pip                    20.3.3
protobuf               3.14.0
pyasn1                 0.4.8
pyasn1-modules         0.2.8
pypiwin32              223
python-dateutil        2.8.1
pytz                   2020.5
pywin32                300
PyYAML                 5.4
requests               2.25.1
requests-oauthlib      1.3.0
rsa                    4.7
setuptools             51.3.3
six                    1.15.0
tensorboard            2.4.1
tensorboard-plugin-wit 1.8.0
torch                  1.7.1+cu110
torchaudio             0.7.2
torchvision            0.8.2+cu110
typing-extensions      3.7.4.3
urllib3                1.26.2
Werkzeug               1.0.1
wheel                  0.36.2
zipp                   3.4.0

chriselion:
Hi @torsilver,
The "Inference Device" controls how the Agent performs inference, not training.

Can you run this from the commandline?
```
python -c 'import torch; print(torch.cuda.is_available())'
```
If that prints `True`, ML-Agents should use the GPU for training. If it prints `False`, there's something wrong with your `torch` setup, but it's not specific to ML-Agents, so you'll need to look elsewhere to get help debugging it.

torsilver:
> Hi @torsilver,
> The "Inference Device" controls how the Agent performs inference, not training.
> 
> Can you run this from the commandline?
> 
> ```
> python -c 'import torch; print(torch.cuda.is_available())'
> ```
> 
[...]

PR #4888, opened 2021-01-27
Set torch device from commandline

chriselion:
### Proposed change(s)
The torch device string can now be set from the commandline. The default behavior is the same as before: it will use cuda if it's detected, or cpu otherwise. This allows overriding the default logic (e.g. force cpu even if you support cuda), or specifying a device index if e.g. you have two different GPU types.

Also logs the device type so that it's obvious which is being used.

https://forum.unity.com/threads/ml-agent-server-build-headless-with-visual-observations-and-multiple-gpus.1042456/
https://github.com/Unity-Technologies/ml-agents/issues/4876

### Types of change(s)
- [x] New feature

### Checklist

### DIFF
--- a/com.unity.ml-agents/CHANGELOG.md
+++ b/com.unity.ml-agents/CHANGELOG.md
@@ -27,6 +27,7 @@ removed when training with a player. The Editor still requires it to be clamped 
 #### ml-agents / ml-agents-envs / gym-unity (Python)
 - Added a `--torch-device` commandline option to `mlagent-learn`, which sets the default
   [`torch.device`](https://pytorch.org/docs/stable/tensor_attributes.html#torch.torch.device) used for training. (#4888)
+- The `--cpu` commandline option had no effect and was removed. Use `--torch-device=cpu` to force CPU training. (#4888)
 
 ### Bug Fixes
 #### com.unity.ml-agents (C#)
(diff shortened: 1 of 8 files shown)


ml-agents/mlagents/trainers/cli_utils.py
```diff
@@ -252,6 +252,15 @@ def _create_parser() -> argparse.ArgumentParser:
         help="Whether to run the Unity executable in no-graphics mode (i.e. without initializing "
         "the graphics driver. Use this only if your agents don't use visual observations.",
     )
+
+    torch_conf = argparser.add_argument_group(title="Torch Configuration")
```
dongruoping, line 256: `--cpu` option should be moved into this group or just integrate with the device argument

dongruoping (commented):
It would be good to also mention this option somewhere in to doc

com.unity.ml-agents/CHANGELOG.md
```diff
@@ -25,6 +25,9 @@ removed when training with a player. The Editor still requires it to be clamped
   Changed the namespace and file names of classes in com.unity.ml-agents.extensions. (#4849)
 
 #### ml-agents / ml-agents-envs / gym-unity (Python)
+- Added a `--torch-device` commandline option to `mlagent-learn`, which sets the default
```
ervteng, line 28:
```suggestion
- Added a `--torch-device` commandline option to `mlagents-learn`, which sets the default
```

chriselion: Fixed

ml-agents/mlagents/trainers/cli_utils.py
chriselion: Removed `--cpu` since it's obsolete (and added a note in the changelog)

ervteng (approved):
:shipit:

PR #4889, opened 2021-01-28
Add verified package info to table

chriselion:
### Proposed change(s)
(I totally over-engineered this one)

Cleanup the table generation logic, add a column for the package docs, and add some verified packages to the table.
At least one the most recent verified package will be present and highlighted, more will be shown if they're in the time range

https://jira.unity3d.com/browse/MLA-1696

### Types of change(s)
- [x] Documentation update


chriselion:
Here's the new version
image

and the rich diff (bumped some old releases off)

mmattar:
Don't have a strong opinion here, but I don't think it's immediately obvious to a newcomer what the difference between a Verified Package and a normal release is. It _may_ be more intuitive if we have a separate, dedicated table for verified packages.

surfnerd (approved):

PR #4890, opened 2021-01-28
Add url to package.json

chriselion:
### Proposed change(s)
This doesn't seem to do anything currently in the Package Manager UI :( but at least makes it a little easier to find the source that a package came from.

### DIFF
--- a/com.unity.ml-agents/package.json
+++ b/com.unity.ml-agents/package.json
@@ -14,7 +14,6 @@
   },
   "author": {
     "name": "Unity Technologies Inc.",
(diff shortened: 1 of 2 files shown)


chriselion:
:sadtrombone:

>     Error: A Unity package must not have an author field. Please remove the field. Read more about this error and potential solutions at https://docs.unity3d.com/Packages/com.unity.package-validation-suite@latest/index.html?preview=1&subfolder=/manual/manifest_validation_error.html%23a_unity_package_must_not_have_an_author_field

PR #4913, opened 2021-02-05
Convert 3DBallHard to use Observables

chriselion:
### Proposed change(s)
Convert 3DBallHard to use Observables - this removes the CollectObservations method and sets the vector observation size to 0 in the BrainParameters.

Needed to retrain the model too, since the order of the observations change; the old nn file is replaced with an onnx model.


vincentpierre (approved):
Looks good to me.
 Should we link this example into the documentation about observables ?

chriselion:
There was already an example in the docs of how 3DBall could be modified to use [Observable]. I updated it with the actual code from 3DBallHard

PR #4914, opened 2021-02-05
R2v yamato linux

chriselion:
### Proposed change(s)
Try porting the yamato changes from master. This is mostly the changes from https://github.com/Unity-Technologies/ml-agents/pull/4767

### Types of change(s)
- [x] CI


.pre-commit-config.yaml
```diff
@@ -77,18 +77,6 @@ repos:
     hooks:
     -   id: python-check-mock-methods
 
--   repo: https://github.com/pre-commit/mirrors-pylint
```
chriselion, line 80: Removed on master, was causing false positives

PR #4925, opened 2021-02-08
add option for Burst inference

chriselion:
### Proposed change(s)
Followup from slack with Florent: Add InferenceDevice.Burst and make it the default. I don't think changing the existing behavior of InferenceDevice.CPU in a minor version is allowed though.

Also need to keep CPU as the default for Agent.SetModel: https://github.com/Unity-Technologies/ml-agents/blob/729b4cfe860c522a5dc245aa9f621781c263d73b/com.unity.ml-agents/Runtime/Agent.cs#L586-L589
for now.

https://jira.unity3d.com/browse/MLA-1763
https://jira.unity3d.com/browse/MLA-1765 (followup to break the behavior in the next major version)

### Types of change(s)
- [ ] New feature

### Checklist


chriselion:
Prefab changes coming soon.

surfnerd (approved):

PR #4927, opened 2021-02-09
[CI] check timers for values, run in dev mode

chriselion:
### Proposed change(s)
Follow-up from some previous problems with inference. Some torch export settings were producing models that were loading (and thus passing the ModelOverride checks) but raising assertions during inference. To help spot this in the future:
* Run development builds for training tests
* Ensure that we get _some_ reward data in the timer output. If inference completely barfs, there shouldn't be anything ther (will confirm in another test).

### Useful links (Github issues, JIRA tickets, ML-Agents forum threads etc.)
https://jira.unity3d.com/browse/MLA-1657

### Types of change(s)
- [ ] CI

PR #4928, opened 2021-02-09
[DO NOT MERGE] test failing assert

chriselion:
Make sure https://github.com/Unity-Technologies/ml-agents/pull/4927 actually fails if we assert during inference. Will manually trigger a yamato job.

### DIFF
--- a/com.unity.ml-agents/Runtime/Inference/ModelRunner.cs
+++ b/com.unity.ml-agents/Runtime/Inference/ModelRunner.cs
@@ -157,7 +157,7 @@ namespace Unity.MLAgents.Inference
 
         public void DecideBatch()
         {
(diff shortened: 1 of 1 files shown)


com.unity.ml-agents/Runtime/Inference/ModelRunner.cs
```diff
@@ -156,6 +157,7 @@ public void PutObservations(AgentInfo info, List<ISensor> sensors)
 
         public void DecideBatch()
         {
+            Debug.Assert(2 + 2 == 5);
```
surfnerd, line 160:
```suggestion
            UnityEngine.Assertions.Assert.IsTrue(2 + 2 == 5);
```

PR #4930, opened 2021-02-09
surface specific GRPC errors more visibly

chriselion:
### Proposed change(s)
We want to ignore some GRPC errors but raise others. Currently just opting in to one error (ResourceExhausted, when the payload is too large), but I'm willing to reconsider.

Also not sure how to test this; I did it manually with some breakpoints (to confirm the `StatusCode.Unavailable` case) and modifying GridWorld with a large uncompressed sensor (to confirm `StatusCode.ResourceExhausted`),

https://jira.unity3d.com/browse/MLA-1719

### Types of change(s)
- [x] Bug fix

### Checklist

### DIFF
--- a/com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
+++ b/com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
@@ -440,6 +440,7 @@ namespace Unity.MLAgents
             {
                 return null;
             }
+
             try
             {
                 var message = m_Client.Exchange(WrapMessage(unityOutput, 200));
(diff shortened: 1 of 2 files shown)


com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
```diff
@@ -455,8 +456,31 @@ UnityInputProto Exchange(UnityOutputProto unityOutput)
                 QuitCommandReceived?.Invoke();
                 return message.UnityInput;
             }
+            catch (RpcException rpcException)
+            {
+                // Log more verbose errors if they're something the user can possibly do something about.
+                switch (rpcException.Status.StatusCode)
+                {
+                    case StatusCode.Unavailable:
+                        // This can happen when python disconnects. Ignore it to avoid noisy logs.
```
chriselion, line 465: This happens roughly 20%-25% of the time when hitting Ctrl-C from python. Not exactly sure why it happens instead of the non-200 status (handled above).

com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
```diff
@@ -455,8 +456,31 @@ UnityInputProto Exchange(UnityOutputProto unityOutput)
...
+                        break;
+                    case StatusCode.ResourceExhausted:
+                        // This happens is the message body is too large. There's no way to
+                        // gracefully handle this, but at least we can show the message and the
+                        // user can try to reduce the number of agents or observation sizes.
+                        Debug.LogError($"GRPC Exception: {rpcException.Message}. Exiting communicator.");
+                        break;
+                    default:
+                        // Other unknown errors. Keep these quiet for now.
+                        break;
```
chriselion, line 475: I could be convinced to log other exceptions here (maybe as Info instead of Error). That would at least help us for future unknown unknowns.

com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
```diff
@@ -455,8 +456,31 @@ UnityInputProto Exchange(UnityOutputProto unityOutput)
...
+                }
+                m_IsOpen = false;
+                QuitCommandReceived?.Invoke();
+                return null;
+            }
             catch
             {
+                // Fall-through for other error types
```
chriselion, line 483: Same comment about logging here - I don't think we ever hit it, so it would be nice to know when we do, but it's also potentially noisy to users.

surfnerd (approved):
so happy

PR #4935, opened 2021-02-10
[MLA-1767] Refactor communicator connection exceptions

chriselion:
### Proposed change(s)
Previously we were
* Using exceptions for control flow, since RpcCommunicator.Initialize returns a struct (so it can't return null to indicate normal lack of trainer).
* Hiding all exceptions, so that actual bugs aren't surfaced. One example of this was a SideChannel throwing an exception on the initial message; this would silently stop training and go to inference.

The main change here is to make ICommunicator.Initialize() return a bool. Unexpected exceptions (basically anything but a RpcException with Unavailable status, which is what happens when there's no communicator listening) are logged.

Also, exceptions during SideChannel processing are wrapped in a try/catch, so that a single SideChannel can't bring down the rest of the system.

### Types of change(s)
- [x] Code refactor

### Checklist

### DIFF
--- a/com.unity.ml-agents/Runtime/Academy.cs
+++ b/com.unity.ml-agents/Runtime/Academy.cs
@@ -454,18 +454,19 @@ namespace Unity.MLAgents
                         TrainerCapabilities = unityRlInitParameters.TrainerCapabilities;
                         TrainerCapabilities.WarnOnPythonMissingBaseRLCapabilities();
                     }
+                    else
+                    {
+                        Debug.Log($"Couldn't connect to trainer on port {port} using API version {k_ApiVersion}. Will perform inference instead.");
+                        Communicator = null;
+                    }
                 }
                 catch (Exception ex)
                 {
-                    Debug.Log($"Unexpected exception when trying to initialize communication: {ex}");
-                }
-
-                if (!initSuccessful)
-                {
-                    Debug.Log($"Couldn't connect to trainer on port {port} using API version {k_ApiVersion}. Will perform inference instead.");
+                    Debug.Log($"Unexpected exception when trying to initialize communication: {ex}\nWill perform inference instead.");
                     Communicator = null;
                 }
             }
+
             if (Communicator != null)
             {
                 Communicator.QuitCommandReceived += OnQuitCommandReceived;
--- a/com.unity.ml-agents/Runtime/Academy.cs
+++ b/com.unity.ml-agents/Runtime/Academy.cs
@@ -430,29 +430,39 @@ namespace Unity.MLAgents
             {
                 // We try to exchange the first message with Python. If this fails, it means
                 // no Python Process is ready to train the environment. In this case, the
-                //environment must use Inference.
+                // environment must use Inference.
+                bool initSuccessful = false;
+                var communicatorInitParams = new CommunicatorInitParameters
+                {
+                    unityCommunicationVersion = k_ApiVersion,
+                    unityPackageVersion = k_PackageVersion,
+                    name = "AcademySingleton",
+                    CSharpCapabilities = new UnityRLCapabilities()
+                };
+
                 try
                 {
-                    var unityRlInitParameters = Communicator.Initialize(
-                        new CommunicatorInitParameters
-                        {
-                            unityCommunicationVersion = k_ApiVersion,
-                            unityPackageVersion = k_PackageVersion,
-                            name = "AcademySingleton",
-                            CSharpCapabilities = new UnityRLCapabilities()
-                        });
-                    UnityEngine.Random.InitState(unityRlInitParameters.seed);
-                    // We might have inference-only Agents, so set the seed for them too.
-                    m_InferenceSeed = unityRlInitParameters.seed;
-                    TrainerCapabilities = unityRlInitParameters.TrainerCapabilities;
-                    TrainerCapabilities.WarnOnPythonMissingBaseRLCapabilities();
-                }
-                catch
-                {
-                    Debug.Log($"" +
-                        $"Couldn't connect to trainer on port {port} using API version {k_ApiVersion}. " +
-                        "Will perform inference instead."
+                    initSuccessful = Communicator.Initialize(
+                        communicatorInitParams,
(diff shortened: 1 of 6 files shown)


com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
```diff
@@ -64,8 +64,8 @@ public RpcCommunicator(CommunicatorInitParameters communicatorInitParameters)
 
         internal static bool CheckCommunicationVersionsAreCompatible(
             string unityCommunicationVersion,
-            string pythonApiVersion,
-            string pythonLibraryVersion)
```
chriselion, line 68: Removed this since it was unused.

com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
```diff
@@ -113,62 +113,75 @@ public UnityRLInitParameters Initialize(CommunicatorInitParameters initParameter
                     {
                         RlInitializationOutput = academyParameters
                     },
-                    out input);
-
-                var pythonPackageVersion = initializationInput.RlInitializationInput.PackageVersion;
```
chriselion, line 118: Moved this logic outside of the `try`. It no longer throws (`throw new UnityAgentsException("ICommunicator.Initialize() failed.");`) and instead returns false.

com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
```diff
@@ -113,62 +113,75 @@ public UnityRLInitParameters Initialize(CommunicatorInitParameters initParameter
...
+                else
+                {
+                    Debug.Log($"Unexpected exception when trying to initialize communication: {ex}");
+                }
+                initParametersOut = new UnityRLInitParameters();
+                return false;
             }
-            catch
-            {
-                var exceptionMessage = "The Communicator was unable to connect. Please make sure the External " +
```
chriselion, line 156: Removed this messaging - I'm not sure it was ever useful, since the exception was swallowed by the calling code.

com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
```diff
@@ -456,33 +467,35 @@ UnityInputProto Exchange(UnityOutputProto unityOutput)
                 QuitCommandReceived?.Invoke();
                 return message.UnityInput;
             }
-            catch (RpcException rpcException)
```
chriselion, line 459: Reorganized this to avoid duplicating some of the logic (e.g. `m_IsOpen = false`). Not sure if we have a preference; it's not mentioned in the style guide.

com.unity.ml-agents/Runtime/SideChannels/SideChannel.cs
```diff
@@ -34,9 +35,18 @@ public Guid ChannelId
 
         internal void ProcessMessage(byte[] msg)
         {
-            using (var incomingMsg = new IncomingMessage(msg))
+            try
```
chriselion, line 38: This threw me off when I was working on the Training Analytics side channel. If the SideChannel throws here, we disconnect the trainer. In general, I think we should wrap user code in bubble wrap.

chriselion:
You may want to hide whitespace changes when reviewing this 
image

com.unity.ml-agents/Runtime/Academy.cs
```diff
@@ -430,29 +430,39 @@ void InitializeEnvironment()
...
+                catch (Exception ex)
                 {
-                    Debug.Log($"" +
-                        $"Couldn't connect to trainer on port {port} using API version {k_ApiVersion}. " +
-                        "Will perform inference instead."
-                    );
+                    Debug.Log($"Unexpected exception when trying to initialize communication: {ex}");
+                }
+
+                if (!initSuccessful)
```
surfnerd, line 463: why not an else after the first ```if (initSuccessful) {}```

com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
```diff
@@ -113,62 +114,75 @@ public UnityRLInitParameters Initialize(CommunicatorInitParameters initParameter
...
-                    pythonCommunicationVersion,
-                    pythonPackageVersion);
-
-                // Initialization succeeded part-way. The most likely cause is a mismatch between the communicator
-                // API strings, so log an explicit warning if that's the case.
-                if (initializationInput != null && input == null)
+                    out input
+                );
+            }
+            catch (Exception ex)
```
surfnerd, line 120: could you do cascading catches here instead of type checking? i.e.

```
catch (RpcException ex) {}
catch (AnotherException ex) {}
catch (Exception ex) { // unknown exceptions handled here }
```

com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
```diff
@@ -456,33 +468,35 @@ UnityInputProto Exchange(UnityOutputProto unityOutput)
                 QuitCommandReceived?.Invoke();
                 return message.UnityInput;
             }
-            catch (RpcException rpcException)
+            catch (Exception ex)
```
surfnerd, line 471: same comment about cascading exceptions above.

com.unity.ml-agents/Runtime/Academy.cs
chriselion: It's meant to handle both `Communicator.Initialize` returning false and throwing. But I think you're right, it's cleaner to handle each separately.

surfnerd (approved):

com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
chriselion: Discussed offline, this is to avoid duplicating the common code between RpcException and Exception. Updated this to use pattern matching. https://docs.microsoft.com/en-us/dotnet/csharp/pattern-matching

PR #4936, opened 2021-02-10
[yamato] drop 2020.1, reduce test variants, tighten triggers

chriselion:
### Proposed change(s)
* Remove 2020.1 from the list of editors to test against, now that 2020.2 is out
* For gym, LL-API, and compressed sensor tests, only test against one version of the editor.
* Tighten the triggers for these tests since they don't depend on general ml-agents code.

### Types of change(s)
- [x] CI


chriselion:
Manually triggered the backcompat test.

surfnerd (approved):

PR #4946, opened 2021-02-14
Release 13 versions.

surfnerd:
Update versions for release 13 on the release branch.


gym-unity/gym_unity/__init__.py
```diff
@@ -1,5 +1,5 @@
 # Version of the library that will be used to upload to pypi
-__version__ = "0.24.0.dev0"
+__version__ = "0.24.0"
 
 # Git tag that will be checked to determine whether to trigger upload to pypi
 __release_tag__ = None
```
chriselion, line 5: Make sure to use the script that the release checklist says - that should set the release tag here (and you'll have to update a bunch of doc links once that happens).

utils/validate_release_links.py
```diff
@@ -72,28 +74,47 @@ def get_release_tag() -> Optional[str]:
...
-                continue
-
-            if filename in ALLOW_LIST:
-                if ALLOW_LIST[filename] is None or ALLOW_LIST[filename].search(line):
-                    continue
+    new_file_name = f"{filename}.orig"
+    with open(new_file_name, "a") as new_file:
+        with open(filename) as f:
+            for line in f:
+                keepLine = True
```
chriselion, line 88: nit: keep_line (python uses snake_case for variables)

utils/validate_release_links.py
```diff
@@ -72,28 +74,47 @@ def get_release_tag() -> Optional[str]:
...
-            if not RELEASE_PATTERN.search(line):
-                continue
-
-            if global_allow_pattern.search(line):
-                continue
-
-            if filename in ALLOW_LIST:
-                if ALLOW_LIST[filename] is None or ALLOW_LIST[filename].search(line):
-                    continue
+    new_file_name = f"{filename}.orig"
```
chriselion, line 84: I think we can simplify this a bit with a `tempfile` (don't need to clean it up after, and name should be unique). But don't worry about it for this PR.

utils/validate_release_links.py
```diff
@@ -72,28 +74,47 @@ def get_release_tag() -> Optional[str]:
...
+    with open(new_file_name, "a") as new_file:
+        with open(filename) as f:
+            for line in f:
+                keepLine = True
+                keepLine = not RELEASE_PATTERN.search(line)
+                keepLine |= global_allow_pattern.search(line) is not None
+                if filename in ALLOW_LIST:
```
chriselion, line 91: nit: could hoist this out of the `for line` loop. And you could simplify a bit with something like
```python
allow_list_pattern = ALLOW_LIST.get(filename, re.compile(""))
...
keep_line |= allow_list_pattern.search(line) is not None
```
(I think - I'm assuming the empty string `re` matches everything)

chriselion (approved):

utils/validate_release_links.py
surfnerd: good idea

PR #4957, opened 2021-02-17
update defines, compile out Initialize body on non-desktop

chriselion:
### Proposed change(s)
https://github.com/Unity-Technologies/ml-agents/pull/4935/files introduced compile errors on non-desktop platforms.
image

This fixes them by omitting the body of Initialize(), and just returning false.

### Types of change(s)
- [x] Bug fix


com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
```diff
@@ -183,6 +188,10 @@ out input
             UpdateEnvironmentWithInput(input.RlInput);
             initParametersOut = initializationInput.RlInitializationInput.ToUnityRLInitParameters();
             return true;
+#else
+            initParametersOut = new UnityRLInitParameters();
```
surfnerd, line 192: does this still reference the grpc stuff?

surfnerd: does that matter?

surfnerd (approved):

com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
chriselion: No, this is the C#-only class.

The breaking change in the other PR was referencing RpcException and the status codes, the classes in the signature didn't change (although the return value was moved to an `out` parameter)

PR #4965, opened 2021-02-18
Communicator factory

chriselion:
### Proposed change(s)
Follow up for the "RpcCommunicator on non-desktop" compilation problems.
* Makes all of RpcCommunicator.cs empty on non-desktop.
* Adds a CommunicatorFactory that returns either an RpcCommunicator or null
* Cleans up RpcCommunicator - it was being passed `CommunicatorInitParameters` in both it's constructor and Initialize() method. It stored the former but only used the port from it.
* (possibly controversial) Setting `CommunicatorFactory.Enabled = false` will skip creating (and attempting to connect) with RpcCommunicator. This is a bit clunky but it's the only way we have to avoid trying to connect to a trainer, which could be a problem for shipping games.

### Types of change(s)
- [x] Bug fix
- [x] Code refactor

### Checklist

### DIFF
--- a/com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
+++ b/com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
@@ -1,6 +1,5 @@
-# if UNITY_EDITOR || UNITY_STANDALONE_WIN || UNITY_STANDALONE_OSX || UNITY_STANDALONE_LINUX
+#if UNITY_EDITOR || UNITY_STANDALONE_WIN || UNITY_STANDALONE_OSX || UNITY_STANDALONE_LINUX
 using Grpc.Core;
-#endif
 #if UNITY_EDITOR
 using UnityEditor;
 #endif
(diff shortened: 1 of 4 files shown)


com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
```diff
@@ -44,23 +43,17 @@ internal class RpcCommunicator : ICommunicator
...
         UnityToExternalProto.UnityToExternalProtoClient m_Client;
-#endif
-        /// The communicator parameters sent at construction
-        CommunicatorInitParameters m_CommunicatorInitParameters;
 
         /// <summary>
         /// Initializes a new instance of the RPCCommunicator class.
         /// </summary>
-        /// <param name="communicatorInitParameters">Communicator parameters.</param>
-        public RpcCommunicator(CommunicatorInitParameters communicatorInitParameters)
```
chriselion, line 58: Don't pass these and don't store them anymore. They were only used for the port; that's now passed to Initialize()

surfnerd (approved):
🙏

vincentpierre (approved):
How did you test this was working for mobile builds ?

chriselion:
Manual testing that I can do a WebGL build locally.

I have a PR to start doing a WebGL build in yamato to guard against similar compiler errors in the future: https://github.com/Unity-Technologies/ml-agents/pull/4966 (currently failing because it lacks this fix)

PR #4988, opened 2021-02-22
[WIP] Demonstration provider

chriselion:
### Proposed change(s)
Very rough WIP to convert GAIL and BC to use a new DemonstrationProvider interface.

The eventual goal (beyond the scope of this PR) is to let users define their own DemonstrationProvider interface in a plugin and use that instead of the given LocalDemonstrationProvider.

In this PR:
* Add DemonstrationProvider interface
* Break up the previous demo_loader() code into demonstration_proto_utils and LocalDemonstrationProvider impl
* Convert GAIL and BC to use DemonstrationProvider.

TODO
* Fix tets
* Decide what fields should actually be on DemonstrationExample/DemonstratoinTrajectory
* Remove old demo_loader code

https://jira.unity3d.com/browse/MLA-1734

### Types of change(s)
- [x] New feature
- [x] Code refactor

### Checklist


ml-agents/mlagents/trainers/demonstrations/demonstration_proto_utils.py
```diff
@@ -0,0 +1,96 @@
...
+from google.protobuf.internal.decoder import _DecodeVarint32  # type: ignore
+from google.protobuf.internal.encoder import _EncodeVarint  # type: ignore
+
+
+INITIAL_POS = 33
+SUPPORTED_DEMONSTRATION_VERSIONS = frozenset([0, 1])
+
+
+@timed
+def load_demonstration(
```
chriselion, line 25: was demo_loader.load_demonstration

ml-agents/mlagents/trainers/demonstrations/local_demonstration_provider.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def pop_trajectories(self) -> List[DemonstrationTrajectory]:
+        trajectories = LocalDemonstrationProvider._info_action_pairs_to_trajectories(
+            self._behavior_spec, self._info_action_pairs
+        )
+        self._info_action_pairs = []
+        return trajectories
+
+    @staticmethod
+    def _get_demo_files(path: str) -> List[str]:
```
chriselion, line 43: from demo_loader.get_demo_files

ml-agents/mlagents/trainers/demonstrations/local_demonstration_provider.py
```diff
@@ -0,0 +1,146 @@
...
+
+        demo_paths = self._get_demo_files(file_path)
+        behavior_spec, info_action_pairs, = load_demonstration(demo_paths)
+        self._behavior_spec = behavior_spec
+        self._info_action_pairs = info_action_pairs
+
+    def get_behavior_spec(self) -> BehaviorSpec:
+        return self._behavior_spec
+
+    def pop_trajectories(self) -> List[DemonstrationTrajectory]:
```
chriselion, line 35: Need to add docstrings here. But the idea is that GAIL, etc could be converted to use pop_trajectories() directly. Then if we want DemonstrationProviders to be able to load new demonstrations on the fly, the logic can be kept in the DemonstrationProvider and the consumer doesn't need to know about it, it just gets a fresh batch of trajectories.

ml-agents/mlagents/trainers/demonstrations/demonstration_provider.py
```diff
@@ -0,0 +1,72 @@
...
+
+from typing import List, NamedTuple
+
+from mlagents_envs.base_env import ActionTuple, BehaviorSpec
+
+from mlagents.trainers.buffer import AgentBuffer, BufferKey
+from mlagents.trainers.trajectory import ObsUtil
+
+
+class DemonstrationExperience(NamedTuple):
```
chriselion, line 13: These are trimmed down versions of AgentExperience and Trajectory classes, based on what's currently in demo_loader.

ervteng: Hmm, I feel like we shouldn't duplicate the conversion code here between AgentExperience and Trajectory (esp. with the teammate observations coming in, it becomes quite a fat function - and at some point I imagine we'll have teammate demonstrations as well). 

Wonder if we can have a BaseAgentExperience be the base class that is used here and in the AgentProcessor, and have the AgentExperience (PolicyAgentExperience?) inherit from it? Or some other way of composing these two.

chriselion:
Shelving this for now, since research won't be able to make use of it for a while, and we're still not sure on the trajectory/experience handling.

PR #4999, opened 2021-02-24
Make TrainingAnalyticsSideChannel internal

chriselion:
Noticed TrainingAnalyticsSideChannel was missing docs on master (@surfnerd already fixed on the release branch), and realized that the class doesn't need to be public


surfnerd (approved):
:shipit:

PR #5007, opened 2021-02-24
Release 13 branch to master

surfnerd:
Merge changes from release 13 to master


com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
```diff
@@ -43,6 +47,7 @@ internal class RpcCommunicator : ICommunicator
         Dictionary<string, ActionSpec> m_UnsentBrainKeys = new Dictionary<string, ActionSpec>();
 
 
+#if MLA_SUPPORTED_TRAINING_PLATFORM
```
chriselion, line 50: I think this was the merge problem I was worried about. I think you can either remove all the `#if MLA_SUPPORTED_TRAINING_PLATFORM` except the top-level one, or just keep master the master version.

chriselion (approved):
Looks good (modulo the RpCommunicator merge)

com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
surfnerd: should be fixed now

PR #5014, opened 2021-02-26
Make GridSensor a non allocating object after initialization.

surfnerd:
### Proposed change(s)

Make the GridSensor as non-allocating as possible.

#### Before:
Screen Shot 2021-02-26 at 10 06 26 AM

#### Now: (only allocations are from Barracuda)
Screen Shot 2021-02-26 at 9 58 51 AM

#### New UI for estimating the max number of colliders you will hit per cell:
Screen Shot 2021-02-26 at 10 29 19 AM

For food collector, I let it run for a bit and saw that it never went above 4, so that's what I set it at.

Tooltips for that UI:
```
        [Header("Collider Buffer Properties")]
        [Tooltip("The absolute max size of the Collider buffer used in the non-allocating Physics calls.  In other words" +
            " the Collider buffer will never grow beyond this number even if there are more Colliders in the Grid Cell.")]
        public int AbsoluteMaxCollidersPerCell = 500;
        [Tooltip(
            "The Estimated Max Number of Colliders to expect per cell.  This number is used to " +
            "pre-allocate an array of Colliders in order to take advantage of the OverlapBoxNonAlloc " +
            "Physics API.  If the number of colliders found is >= EstimatedMaxCollidersPerCell the array " +
            "will be resized to double its current size.  The hard coded absolute size is 500.")]
        public int EstimatedMaxCollidersPerCell = 4;
```

### Types of change(s)

- [x] Bug fix. Removal of unnecessary memory allocations.

### DIFF
--- a/com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
+++ b/com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
@@ -639,7 +639,7 @@ namespace Unity.MLAgents.Extensions.Sensors
                 Profiler.BeginSample("ClosestPointOnBounds");
                 var closestColliderPoint = foundColliders[i].ClosestPointOnBounds(cellCenter);
                 Profiler.EndSample();
-                var currentDistance = Vector3.Distance(closestColliderPoint, rootReference.transform.position);
+                var currentDistance = (closestColliderPoint - rootReference.transform.position).sqrMagnitude;
 
                 Profiler.BeginSample("IndexCheck");
                 // Checks if our colliders contain a detectable object
@@ -662,7 +662,7 @@ namespace Unity.MLAgents.Extensions.Sensors
             }
 
             if (!ReferenceEquals(closestColliderGo, null))
-                LoadObjectData(closestColliderGo, cellIndex, distance / SphereRadius);
+                LoadObjectData(closestColliderGo, cellIndex, (float)Math.Sqrt(distance) / SphereRadius));
             Profiler.EndSample();
         }
 
--- a/com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
+++ b/com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
@@ -172,7 +172,7 @@ namespace Unity.MLAgents.Extensions.Sensors
         /// <summary>
         /// Radius of grid, used for normalizing the distance.
         /// </summary>
-        protected float SphereRadius;
+        protected float InverseSphereRadius;
 
         /// <summary>
         /// Total Number of cells (width*height)
@@ -310,7 +310,7 @@ namespace Unity.MLAgents.Extensions.Sensors
             NumCells = GridNumSideX * GridNumSideZ;
             float sphereRadiusX = (CellScaleX * GridNumSideX) / Mathf.Sqrt(2);
             float sphereRadiusZ = (CellScaleZ * GridNumSideZ) / Mathf.Sqrt(2);
-            SphereRadius = Mathf.Max(sphereRadiusX, sphereRadiusZ);
+            InverseSphereRadius = 1.0f / Mathf.Max(sphereRadiusX, sphereRadiusZ);
             ChannelOffsets = new int[ChannelDepth.Length];
             DiffNumSideZX = (GridNumSideZ - GridNumSideX);
             OffsetGridNumSide = (GridNumSideZ - 1f) / 2f;
@@ -431,11 +431,14 @@ namespace Unity.MLAgents.Extensions.Sensors
             m_perceptionTexture2D = new Texture2D(GridNumSideX, GridNumSideZ, TextureFormat.RGB24, false);
         }
 
+        /// <inheritdoc cref="ISensor.Reset"/>
+        void ISensor.Reset() { }
+
         /// <summary>
         /// Clears the perception buffer before loading in new data. If the gridDepthType is ChannelHot, then it initializes the
         /// Reset() also reinits the cell activity array (for debug)
         /// </summary>
-        public void Reset()
+        public void ClearPerceptionBuffer()
         {
             if (m_PerceptionBuffer != null)
             {
@@ -557,7 +560,7 @@ namespace Unity.MLAgents.Extensions.Sensors
             {
                 return Array.Empty<float>();
             }
-            Reset();
+            ClearPerceptionBuffer();
             using (TimerStack.Instance.Scoped("GridSensor.Perceive"))
             {
                 var halfCellScale = new Vector3(CellScaleX / 2f, CellScaleY, CellScaleZ / 2f);
@@ -628,7 +631,7 @@ namespace Unity.MLAgents.Extensions.Sensors
         {
             Profiler.BeginSample("GridSensor.ParseColliders");
             GameObject closestColliderGo = null;
-            var distance = float.MaxValue;
+            var minDistanceSquared = float.MaxValue;
 
             for (var i = 0; i < numFound; i++)
             {
@@ -639,7 +642,7 @@ namespace Unity.MLAgents.Extensions.Sensors
                     continue;
 
                 var closestColliderPoint = foundColliders[i].ClosestPointOnBounds(cellCenter);
-                var currentDistance = (closestColliderPoint - rootReference.transform.position).sqrMagnitude;
+                var currentDistanceSquared = (closestColliderPoint - rootReference.transform.position).sqrMagnitude;
 
                 // Checks if our colliders contain a detectable object
                 var index = -1;
@@ -651,15 +654,15 @@ namespace Unity.MLAgents.Extensions.Sensors
                         break;
                     }
                 }
-                if (index > -1 && currentDistance < distance)
+                if (index > -1 && currentDistanceSquared < minDistanceSquared)
                 {
-                    distance = currentDistance;
+                    minDistanceSquared = currentDistanceSquared;
                     closestColliderGo = currentColliderGo;
                 }
             }
 
             if (!ReferenceEquals(closestColliderGo, null))
-                LoadObjectData(closestColliderGo, cellIndex, (float)Math.Sqrt(distance) / SphereRadius);
+                LoadObjectData(closestColliderGo, cellIndex, (float)Math.Sqrt(minDistanceSquared) * InverseSphereRadius);
             Profiler.EndSample();
         }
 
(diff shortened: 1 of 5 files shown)


com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
```diff
@@ -669,83 +734,100 @@ protected virtual void ValidateValues(float[] channelValues, GameObject currentC
...
         ///            the distance currentColliderGo is compared to the edge of the gridsensor</param>
-        protected virtual void LoadObjectData(GameObject currentColliderGo, int cellIndex, float normalized_distance)
+        protected virtual void LoadObjectData(GameObject currentColliderGo, int cellIndex, float normalizedDistance)
         {
-            for (int i = 0; i < DetectableObjects.Length; i++)
+            Profiler.BeginSample("GridSensor.LoadObjectData");
+            unsafe
             {
-                if (currentColliderGo != null && currentColliderGo.CompareTag(DetectableObjects[i]))
+                var channelHotVals = stackalloc float[ObservationPerCell];
```
surfnerd, line 744: Here's where I am allocating a stack buffer for use, where previously a GC allocation was made within the loop.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
```diff
@@ -536,73 +551,119 @@ protected void ChannelsToTexture(int channelIndex, int numChannelsToAdd)
...
-                if (currentColliderGo == rootReference)
+                if (ReferenceEquals(currentColliderGo, rootReference))
                     continue;
 
-                closestColliderPoint = foundColliders[i].ClosestPointOnBounds(cellCenter);
-                currentDistance = Vector3.Distance(closestColliderPoint, rootReference.transform.position);
+                Profiler.BeginSample("ClosestPointOnBounds");
+                var closestColliderPoint = foundColliders[i].ClosestPointOnBounds(cellCenter);
+                Profiler.EndSample();
+                var currentDistance = Vector3.Distance(closestColliderPoint, rootReference.transform.position);
```
chriselion, line 642: Small optimization - you can used distance squared here
`(closestColliderPoint - rootReference.transform.position).SqrMagnitude()`
and take a sqrt() before calling LoadObjectData instead.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
```diff
@@ -98,6 +99,18 @@ public enum GridDepthType { Channel, ChannelHot };
         [Tooltip("The reference of the root of the agent. This is used to disambiguate objects with the same tag as the agent. Defaults to current GameObject")]
         public GameObject rootReference;
 
+        const int k_AbsoluteMaxCollidersPerCell = 500;
```
surfnerd, line 102: Hard limit on Collider Buffer Size

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
```diff
@@ -536,73 +551,119 @@ protected void ChannelsToTexture(int channelIndex, int numChannelsToAdd)
...
+        /// if the number of Colliders in the buffer after the call is equal to the length of the buffer.
+        /// </summary>
+        /// <param name="cellCenter"></param>
+        /// <param name="halfCellScale"></param>
+        /// <param name="rotation"></param>
+        /// <returns></returns>
+        int BufferResizingOverlapBoxNonAlloc(Vector3 cellCenter, Vector3 halfCellScale, Quaternion rotation)
+        {
+            int numFound;
+            while (true)
```
surfnerd, line 601: Elion's proposal from havok daze.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
chriselion, line 102: I'd prefer to have this a parameter too - if someone wants to blow up their memory in order to get maximum fidelity, we should at allow them the option of doing it.

surfnerd: ok, was hoping you'd comment on it.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
chriselion: only skimmed the code, but instead of doing this or the old approach, can we have the GridDepthType.ChannelHot code write to m_PerceptionBuffer directly? Or better, make an ArraySegment for the part of the m_PerceptionBuffer that this cellIndex is allowed to write to? This would avoid the stack allocation and memcpy calls (in the GridDepthType.Channel too).

Hunter-Unity (approved):
code changes look good. opened in editor and confirmed there's a drastic reduction in GC

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
```diff
@@ -98,6 +99,20 @@ public enum GridDepthType { Channel, ChannelHot };
...
+        [Header("Collider Buffer Properties")]
+        [Tooltip("The absolute max size of the Collider buffer used in the non-allocating Physics calls.  In other words" +
+            " the Collider buffer will never grow beyond this number even if there are more Colliders in the Grid Cell.")]
+        public int AbsoluteMaxCollidersPerCell = 500;
+        [Tooltip(
+            "The Estimated Max Number of Colliders to expect per cell.  This number is used to " +
+            "pre-allocate an array of Colliders in order to take advantage of the OverlapBoxNonAlloc " +
+            "Physics API.  If the number of colliders found is >= EstimatedMaxCollidersPerCell the array " +
+            "will be resized to double its current size.  The hard coded absolute size is 500.")]
+        public int EstimatedMaxCollidersPerCell = 4;
```
chriselion, line 111: Maybe InitialColliderBufferSize and MaxColliderBufferSize? I think the `PerCell` is redundant and maybe even a bit misleading, because it could sound like we'd make the buffer (4 * number of cells) big.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
```diff
@@ -536,73 +553,112 @@ protected void ChannelsToTexture(int channelIndex, int numChannelsToAdd)
         /// <returns>A float[] containing all of the information collected from the gridsensor</returns>
         public float[] Perceive()
         {
+            if (m_CollidersPerCell == null)
+            {
+                return Array.Empty<float>();
+            }
             Reset();
```
chriselion, line 560: I missed this before. ISensor.Reset() is meant to be called by the Agent when it resets, so that any stateful information doesn't get carried over when it dies (e.g. StackingSensor). We should probably rename the current GridSensor.Reset() method to something else (`ClearPerceptionBuffer()?`) and make Reset() empty.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
```diff
@@ -536,73 +553,112 @@ protected void ChannelsToTexture(int channelIndex, int numChannelsToAdd)
...
+        /// if the number of Colliders in the buffer after the call is equal to the length of the buffer.
+        /// </summary>
+        /// <param name="cellCenter"></param>
+        /// <param name="halfCellScale"></param>
+        /// <param name="rotation"></param>
+        /// <returns></returns>
+        int BufferResizingOverlapBoxNonAlloc(Vector3 cellCenter, Vector3 halfCellScale, Quaternion rotation)
+        {
+            int numFound;
+            while (true)
```
chriselion, line 601:
```suggestion
            // Since we can only get a fixed number of results, requery
            // until we're sure we can hold them all (or until we hit the max size).
            while (true)
```

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
```diff
@@ -536,73 +553,112 @@ protected void ChannelsToTexture(int channelIndex, int numChannelsToAdd)
...
-        protected virtual void ParseColliders(Collider[] foundColliders, int cellIndex, Vector3 cellCenter)
+        protected virtual void ParseColliders(Collider[] foundColliders, int numFound, int cellIndex, Vector3 cellCenter)
         {
-            GameObject currentColliderGo = null;
+            Profiler.BeginSample("GridSensor.ParseColliders");
             GameObject closestColliderGo = null;
-            Vector3 closestColliderPoint = Vector3.zero;
-            float distance = float.MaxValue;
-            float currentDistance = 0f;
+            var distance = float.MaxValue;
```
chriselion, line 629: Rename to `minDistanceSquared`

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
```diff
@@ -536,73 +553,112 @@ protected void ChannelsToTexture(int channelIndex, int numChannelsToAdd)
...
 
                 // Continue if the current collider go is the root reference
-                if (currentColliderGo == rootReference)
+                if (ReferenceEquals(currentColliderGo, rootReference))
                     continue;
 
-                closestColliderPoint = foundColliders[i].ClosestPointOnBounds(cellCenter);
-                currentDistance = Vector3.Distance(closestColliderPoint, rootReference.transform.position);
+                var closestColliderPoint = foundColliders[i].ClosestPointOnBounds(cellCenter);
+                var currentDistance = (closestColliderPoint - rootReference.transform.position).sqrMagnitude;
```
chriselion, line 640: Rename to `currentDistanceSquared`

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
```diff
@@ -536,73 +553,112 @@ protected void ChannelsToTexture(int channelIndex, int numChannelsToAdd)
...
                 {
                     distance = currentDistance;
                     closestColliderGo = currentColliderGo;
                 }
             }
 
-            if (closestColliderGo != null)
-                LoadObjectData(closestColliderGo, cellIndex, distance / SphereRadius);
+            if (!ReferenceEquals(closestColliderGo, null))
+                LoadObjectData(closestColliderGo, cellIndex, (float)Math.Sqrt(distance) / SphereRadius);
```
chriselion, line 660: microoptimization: store `1.0f / SphereRadius`  as `InverseSphereRadius`, and multiply by that everywhere instead.

com.unity.ml-agents.extensions/Runtime/Unity.ML-Agents.Extensions.asmdef
```diff
@@ -5,6 +5,7 @@
         "Unity.ML-Agents",
         "Unity.ML-Agents.Extensions.Input"
     ],
-    "includePlatforms": [],
-    "excludePlatforms": []
+    "optionalUnityReferences": [],
+    "excludePlatforms": [],
+    "allowUnsafeCode": false
```
chriselion, line 10: can undo this now that we're not `stackalloc`'ing

chriselion (approved):

PR #5021, opened 2021-03-02
[WIP] Jobify grid sensor

surfnerd:
Good: FPS Boost
Bad: Only one collider detected per cell per frame.  Works well on food collector.

Perhaps this should be made into its own grid sensor type.  Need to think about the structure code now, and how we want it to change.

Would like to try this with tanks.


Project/Assets/ML-Agents/Examples/Sorter/Scripts/SorterAgent.cs
```diff
@@ -1,273 +1,273 @@
-using System.Collections.Generic;
```
surfnerd, line 1: i forget why i commented this out, but I will definitely not keep it this way.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
```diff
@@ -937,8 +1096,35 @@ public int Write(ObservationWriter writer)
...
+                {
+                    var c = m_Hits[cellIndex].collider;
+                    if (ReferenceEquals(c, null))
+                    {
+                        continue;
+                    }
+
+                    for (var i = 0; i < DetectableObjects.Length; i++)
+                    {
+                        if (c.CompareTag(DetectableObjects[i]))
```
surfnerd, line 1119: can't do this in a job :(

surfnerd:
Moved the jobified code to the tanks demo branch

PR #5026, opened 2021-03-03
run yamato tests on PRs that target v2-staging

chriselion:
### Proposed change(s)
Run yamato tests on PRs that target v2-staging (like this one!)

### Types of change(s)
- [x] CI


surfnerd (approved):

chriselion:
Yamato tests are running, so that proves this does what it needs to do. Will cancel them to save resources, and merge.

PR #5031, opened 2021-03-04
update master to main, remove pylint check

chriselion:
### Proposed change(s)
Update docs for for master->main rename.

Also remove the pylint check (this was done on master before)


.yamato/com.unity.ml-agents-test.yml
```diff
@@ -54,7 +54,7 @@ all_package_tests:
   triggers:
     cancel_old_ci: true
     recurring:
-      - branch: master
```
chriselion, line 57: Don't think these matter, but just for consistency...

dongruoping (approved):
👍

Issue #5011, opened 2021-02-26
Use a namedtuple (or something similar) for distinct sensors in the Python API


RedTachyon:
**Is your feature request related to a problem? Please describe.**
This is more of a quality-of-life request, but also to prevent possible errors because it's not clear to me what to expect.

When you call `.get_steps(...)`, you get the decision and terminal steps. Let's focus on decision steps (although it applies to both). Under that, you get `decisions.obs` which stores the observations collected by the agent(s). These observations are stored in a list, where each entry corresponds to one sensor, one of them being the "main sensor" filled with CollectObservations on the C# side.

The problem with this - I don't know what order they're in, or how to control that. From experimentation, I figured out that the main sensor is the last entry in the list. Other sensors are somewhere earlier in the list. If you just want to concatenate everything, it's not a problem, but e.g. to run a convolution on ray perception sensors and then connect it with the normal observations, you need to keep them separate and know which is which.

**Describe the solution you'd like**
I think a namedtuple would be the ideal structure for this, with fields named after the sensors used in the agent. Since it's a tuple, it will be compatible with the current list-based version, i.e. you can index into it or iterate over it. Since it's named, you can make sure you're grabbing the right sensor.

**Describe alternatives you've considered**
The alternative is to suck it up, trust that the order of sensors is invariant, and just hard-code it, at least until you want to change the sensors and then you have to rewrite the environment wrapper.

chriselion:
Sorry for the delay on this.

In the latest release, BehaviorSpec changed to have a list of ObservationSpec NamedTuple on it ([source](https://github.com/Unity-Technologies/ml-agents/blob/dd4fd79a059fbedc9149566acf1c99a71d50a9cf/ml-agents-envs/mlagents_envs/base_env.py#L477-L504)); these correspond 1:1 with the observations at runtime. It sounds like if we add the name to this, you'll be able to do what you want, right?

> From experimentation, I figured out that the main sensor is the last entry in the list.

Sensors are sorted alphabetically by name on the C# side. The agent's VectorSensor will get the name `$"VectorSensor_size{observationSize}"` (where `observationSize` is the Agent's observation space size). So that's *probably* the last one in the list, but not necessarily :)

RedTachyon:
Yes, I actually came across the BehaviorSpecs a bit later, but I figured trying to match the dimensions is a level of insanity I'm not ready for. Anyways, yes, adding the name would certainly work!

chriselion:
[...]

PR #5036, opened 2021-03-05
pass sensor name through to ObservationSpec

chriselion:
### Proposed change(s)
Add a `name` field to the Observation proto, set this in C# when creating the proto, and pass it to the ObservationSpec on python.

I'll also combine this into https://github.com/Unity-Technologies/ml-agents/pull/5030 (depending on which finishes first) to add the sensor name in the error message.

https://github.com/Unity-Technologies/ml-agents/issues/5011

### Types of change(s)
- [x] New feature

### Checklist


vincentpierre (approved):
I must admit that I am not a fan of exposing the sensor name to python because we do not have a use case for it. It is also an invitation for the user to try to communicate information with the name of the sensor, which could be a problem.
But these are not good reasons, if you are confident about this, let's do it. 
Is there maybe a way to test it end to end ?

ervteng:
I'm actually in support of this in the long run. In the future we can go as far as e.g. configuring different encoders for different sensors in the YAML. Also in demos we could use it to, for instance, record _all_ the sensor but only train using some subset of them, and save people from re-recording demos over and over.

chriselion:
@ervteng I don't think we should be picking encoders based on the string; we should add flags to explicitly control this (which is what we've been doing with e.g. dimension properties and ObservationType). This is intended to help with debugging, but I think it might make sense for the demonstration filtering too.

@vincentpierre Please see the linked issue - this was a request from a user of the LLAPI to help them debug their sensors.

> Is there maybe a way to test it end to end ?

 I'll a check in the LLAPI test that the name is set to something non-empty

com.unity.ml-agents/Runtime/Communicator/GrpcExtensions.cs
```diff
@@ -427,6 +427,11 @@ public static ObservationProto GetObservationProto(this ISensor sensor, Observat
                 }
             }
             observationProto.Shape.AddRange(shape);
+            var sensorName = sensor.GetName();
+            if (string.IsNullOrEmpty(sensorName))
```
chriselion, line 431: This is why we needed an end-to-end test.

PR #5060, opened 2021-03-09
non-IEnumerable interface for action masking

chriselion:
### Proposed change(s)
This fixes a few things that bugged me about `IDiscreteActionMask:`
* Removes the `GetMask` and `ResetMask` methods - these were part of the implementation in `ActuatorDiscreteActionMask`, but they weren't meant for public use (and there was no safe way to use them in an `OnActionReceived` callback).
* Adds a method that doesn't require an `IEnumerable`. The previous approach required creating an object to pass, and also needed to allocate memory to iterate over it.
* Made what you're doing more explicit, and also undoable - "`WriteMask`" always felt vague, and I always had to look up whether I was enabling or disabling the actions.

Things I haven't done yet but would like feedback on:
* Should we remove `IDiscreteActionMask.WriteMask()`? This would be a larger interface change for anyone using it, but I think it's for the best.
* Should we stop having this as an interface and use a concrete implementation instead? Making it an interface feels strange, as there's no way for a user to implement their own version and get it passed through the system (although this could change in the future if we want to decouple the implementations.

I'm leaning towards removing `WriteMask` but keeping the interface, but wanted to discuss it further.

I haven't updated existing tests or documentation yet; wanted to get feedback on the general approach first.

https://jira.unity3d.com/browse/MLA-1831
https://jira.unity3d.com/browse/MLA-1728

### Types of change(s)
- [x] Code refactor
- [x] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/com.unity.ml-agents.extensions/Runtime/Match3/Match3Actuator.cs
+++ b/com.unity.ml-agents.extensions/Runtime/Match3/Match3Actuator.cs
@@ -31,10 +31,10 @@ namespace Unity.MLAgents.Extensions.Match3
         /// <param name="agent"></param>
         /// <param name="name"></param>
         public Match3Actuator(AbstractBoard board,
-            bool forceHeuristic,
-            int seed,
-            Agent agent,
-            string name)
+                              bool forceHeuristic,
+                              int seed,
+                              Agent agent,
+                              string name)
         {
             m_Board = board;
             m_Rows = board.Rows;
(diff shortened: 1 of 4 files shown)


com.unity.ml-agents.extensions/Runtime/Match3/Match3Actuator.cs
```diff
@@ -122,23 +115,33 @@ IEnumerable<int> InvalidMoveIndices()
                             "an invalid move will be passed to AbstractBoard.MakeMove()."
                         );
                     }
-                    // This means the last move won't be returned as an invalid index.
-                    yield break;
+                    actionMask.SetActionEnabled(branch, numMoves - 1, false);
```
chriselion, line 118: This needs to be `true`

vincentpierre (approved):

PR #4939, opened 2021-02-11
Move the Critic into the Optimizer

andrewcoh:
### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist
- [ ] Added tests that prove my fix is effective or that my feature works
- [ ] Updated the [changelog](https://github.com/Unity-Technologies/ml-agents/blob/master/com.unity.ml-agents/CHANGELOG.md) (if applicable)
- [ ] Updated the [documentation](https://github.com/Unity-Technologies/ml-agents/tree/master/docs) (if applicable)
- [ ] Updated the [migration guide](https://github.com/Unity-Technologies/ml-agents/blob/master/docs/Migrating.md) (if applicable)

PR #5062, opened 2021-03-09
[bug-fix] Fix save/restore critic, add test

ervteng:
### Proposed change(s)

Add critic to the list of modules for optimizer, so that it is saved/restored properly. This was introduced in #4939 and doesn't affect older releases of ML-Agents. 

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents/mlagents/trainers/tests/torch/saver/test_saver.py
+++ b/ml-agents/mlagents/trainers/tests/torch/saver/test_saver.py
@@ -1,3 +1,4 @@
+from mlagents.trainers.optimizer.torch_optimizer import TorchOptimizer
 import pytest
 from unittest import mock
 import os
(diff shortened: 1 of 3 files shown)


vincentpierre (approved):

ml-agents/mlagents/trainers/tests/torch/saver/test_saver.py
```diff
@@ -62,6 +64,42 @@ def test_load_save(tmp_path):
...
+    path1 = os.path.join(tmp_path, "runid1")
+    model_saver = TorchModelSaver(trainer_settings, path1)
+    model_saver.register(policy)
+    model_saver.register(optimizer)
+    model_saver.initialize_or_load()
+    policy.set_step(2000)
+    model_saver.save_checkpoint("MockBrain", 2000)
+
+    # create a new optimizer and policy
+    optimizer2 = OptimizerClass(policy, trainer_settings)
```
dongruoping, line 90: is it intentional that `optimizer2` is using `policy` not `policy2`?

dongruoping (commented):
Is this able to run on GPU? In _compare_two_policies we move the actors explicitly and I just saw that we also added `self.actor.to(default_device())` to TorchPolicy. I didn't see that in optimizers and I wonder if we need to do the same thing as well.

ml-agents/mlagents/trainers/tests/torch/saver/test_saver.py
ervteng: No this be policy2 - good catch

ervteng:
> Is this able to run on GPU? In _compare_two_policies we move the actors explicitly and I just saw that we also added `self.actor.to(default_device())` to TorchPolicy. I didn't see that in optimizers and I wonder if we need to do the same thing as well.

I think I need to add this too, but the different optimizers have different components. I'll try iterating through the `get_modules()` and moving each of those to the default device.

dongruoping (approved):

PR #5072, opened 2021-03-10
cleanup some mypy types

chriselion:
### Proposed change(s)
Small mypy cleanup (IMHO)

### DIFF
--- a/ml-agents/mlagents/trainers/coma/trainer.py
+++ b/ml-agents/mlagents/trainers/coma/trainer.py
@@ -58,7 +58,7 @@ class COMATrainer(RLTrainer):
             PPOSettings, self.trainer_settings.hyperparameters
         )
         self.seed = seed
-        self.policy: Policy = None  # type: ignore
+        self.policy: TorchPolicy = None  # type: ignore
         self.collected_group_rewards: Dict[str, int] = defaultdict(lambda: 0)
 
     def _process_trajectory(self, trajectory: Trajectory) -> None:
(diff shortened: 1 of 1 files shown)


ml-agents/mlagents/trainers/coma/trainer.py
```diff
@@ -253,9 +253,7 @@ def create_torch_policy(
         return policy
 
     def create_coma_optimizer(self) -> TorchCOMAOptimizer:
-        return TorchCOMAOptimizer(  # type: ignore
```
chriselion, line 256: These made me sad :(

ml-agents/mlagents/trainers/coma/trainer.py
```diff
@@ -58,7 +58,7 @@ def __init__(
             PPOSettings, self.trainer_settings.hyperparameters
         )
         self.seed = seed
-        self.policy: Policy = None  # type: ignore
+        self.policy: TorchPolicy = None  # type: ignore
```
andrewcoh, line 61: I guess this still throws a mypy error?

andrewcoh (approved):

ml-agents/mlagents/trainers/coma/trainer.py
chriselion: The `# type: ignore` is to handle it being assigned None - otherwise you'd have to use `Optional[TorchPolicy]` and check that it was non-None everywhere you used it.

PR #5074, opened 2021-03-10
[MLA-1731] update pip instructions

chriselion:
### Proposed change(s)
Update all `pip install` instructions for mlagents and mlagents_envs to specify the released pypi version. This standardizes on the command to be "python -m pip" instead of "pip" or "pip3", which is the recommended approach (see link).

This will run at the same time that we update the `release_xyz_docs` links.

https://www.youtube.com/watch?list=PLWBKAf81pmOaP9naRiNAqug6EBnkPakvY&v=gnYdk_U6UQ0&feature=youtu.be


utils/validate_release_links.py
```diff
@@ -90,21 +150,39 @@ def check_file(
...
                         and allow_list_pattern.search(line) is not None
                     )
 
-                    if keep_line:
+                    pip_install_ok = (
+                        has_allow_list_pattern
+                        or PIP_INSTALL_PATTERN.search(line) is None
+                        or pip_allow_pattern.search(line) is not None
+                    )
+                    # if not pip_install_ok:
```
ervteng, line 170: should this comment be here?

chriselion: Oops, that was debugging, will remove.

PR #5091, opened 2021-03-11
Update install docs to mention samples and link to the forum for feedback

surfnerd:
### Proposed change(s)

Update docs to mention sample and point users to the forum for feedback.

### Types of change(s)

- [x] Documentation update

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -3,7 +3,10 @@
 The ML-Agents Toolkit contains several components:
 
 - Unity package ([`com.unity.ml-agents`](../com.unity.ml-agents/)) contains the
-  Unity C# SDK that will be integrated into your Unity scene.
+  Unity C# SDK that will be integrated into your Unity scene.  This package contains
+  the [3DBall example](../Project/Assets/ML-Agents/Examples/3DBall/Scenes/3DBall.unity)
+  as a sample.  Please let us know what samples you'd like to see shipped with the package in
+  [this forum thread](https://forum.unity.com/threads/feedback-wanted-shipping-sample-s-with-the-ml-agents-package.1073468/).
 - Unity package
   ([`com.unity.ml-agents.extensions`](../com.unity.ml-agents.extensions/))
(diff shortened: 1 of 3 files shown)


docs/Installation.md
```diff
@@ -3,7 +3,10 @@
 The ML-Agents Toolkit contains several components:
 
 - Unity package ([`com.unity.ml-agents`](../com.unity.ml-agents/)) contains the
-  Unity C# SDK that will be integrated into your Unity scene.
+  Unity C# SDK that will be integrated into your Unity scene.  This package contains
+  the [3DBall example](../Project/Assets/ML-Agents/Examples/3DBall/Scenes/3DBall.unity)
```
chriselion, line 7: Can we just say something generic like "The package contains samples to show you how to get started with ML-Agents"? Otherwise we need to maintain the list here, and I think it distracts from the instructions.

If you want to call out the forum thread, can you link to it here instead? https://github.com/Unity-Technologies/ml-agents#community-and-feedback

docs/Installation.md
```diff
@@ -3,7 +3,10 @@
 The ML-Agents Toolkit contains several components:
 
 - Unity package ([`com.unity.ml-agents`](../com.unity.ml-agents/)) contains the
-  Unity C# SDK that will be integrated into your Unity scene.
+  Unity C# SDK that will be integrated into your Unity scene.  This package contains
```
chriselion, line 6:
```suggestion
  Unity C# SDK that will be integrated into your Unity project.  This package contains
```
(not from this PR, but I think that's more correct)

docs/Installation.md
surfnerd: sure

mmattar (commented):
We should also make a comment after Line 57 that you don't need to clone the repo to run through our onboarding guide.

mmattar (commented):
Also need to update this section: https://github.com/Unity-Technologies/ml-agents/blob/main/docs/Getting-Started.md#installation

chriselion (approved):

PR #5109, opened 2021-03-12
Make analytics module an optional dependency.

surfnerd:
### Proposed change(s)

Make analytics module an optional dependency and add tests to make sure that works.

### Types of change(s)

- [x] Bug fix

### Checklist


surfnerd:
Can probably remove all of the asset files in `ProjectSettings`

com.unity.ml-agents/Tests/Editor/Unity.ML-Agents.Editor.Tests.asmdef
```diff
@@ -23,5 +23,12 @@
     "autoReferenced": false,
     "defineConstraints": [
         "UNITY_INCLUDE_TESTS"
+    ],
+    "versionDefines": [
+        {
+            "name": "com.unity.modules.unityanalytics",
+            "expression": "1.1.0",
```
mbargiel, line 30: Was version `1.1.0` by design here? Shouldn't it be `1.0.0`?

surfnerd: No, must have missed this one

com.unity.ml-agents/Runtime/MLAgentsDefines.cs
```diff
@@ -0,0 +1,3 @@
+#if MLA_UNITY_ANALYTICS_MODULE || !UNITY_2019_4_OR_NEWER
```
chriselion, line 1: Just to clarify, this will stay enabled in 2018.4? (Not relevant for long, but I think this is the right move).

com.unity.ml-agents/Runtime/Academy.cs
```diff
@@ -414,7 +414,9 @@ void InitializeEnvironment()
             EnableAutomaticStepping();
 
             SideChannelManager.RegisterSideChannel(new EngineConfigurationChannel());
+#if MLA_UNITY_ANALYTICS_MODULE_ENABLED
```
chriselion, line 417: I think as a rule of thumb, the defines should only be used in the `Unity.MLAgents.Analytics` namespace. Using the defines to make the public `Analytics` methods no-ops means the user doesn't need to know about whether or not it's enabled; helps shrink the effective interface. 

In particular, there's a bug here because the capabilities still say that training analytics are enabled, but you're not registering the SideChannel, so unrecognized message UUIDs will be received.

com.unity.ml-agents/Runtime/MLAgentsDefines.cs
surfnerd: yes

com.unity.ml-agents/Runtime/Academy.cs
surfnerd: done

com.unity.ml-agents/Runtime/Analytics/InferenceAnalytics.cs
```diff
@@ -1,17 +1,31 @@
+#if MLA_UNITY_ANALYTICS_MODULE || !UNITY_2019_4_OR_NEWER
+#define MLA_UNITY_ANALYTICS_MODULE_ENABLED
+#endif
+
+#if MLA_UNITY_ANALYTICS_MODULE_ENABLED
+#endif
```
chriselion, line 6:
```suggestion
```
Redundant lines? Or did Rider auto-refactor something?

com.unity.ml-agents/Runtime/Analytics/TrainingAnalytics.cs
```diff
@@ -56,22 +65,22 @@ internal class TrainingAnalytics
...
-#else
-                AnalyticsResult result = AnalyticsResult.UnsupportedPlatform;
-#endif
                 if (result != AnalyticsResult.Ok)
                 {
                     return false;
                 }
+#else
+                return false;
+#endif // UNITY_EDITOR
```
chriselion, line 83:
```suggestion
#endif // MLA_UNITY_ANALYTICS_MODULE_ENABLED
```

chriselion (approved):
Small feedback, looks OK though.

com.unity.ml-agents/Runtime/Analytics/TrainingAnalytics.cs
surfnerd: that's down below, this is correctly commented with UNITY_EDITOR

com.unity.ml-agents/Runtime/Analytics/InferenceAnalytics.cs
surfnerd: remnants of old debugging code. thanks for catching it

com.unity.ml-agents/Runtime/Analytics/TrainingAnalytics.cs
chriselion: My mistake, I must have only seen the diff-highlighted code.

PR #5110, opened 2021-03-13
Make IActuator implement IHeuristicProvider.

surfnerd:
### Proposed change(s)

Have IActuator implement IHeuristicProvider.

[MLA-1799](https://jira.unity3d.com/browse/MLA-1799)
[MLA-1826](https://jira.unity3d.com/browse/MLA-1826)

### Types of change(s)

- [x] Code refactor
- [x] Breaking change

### Checklist

### DIFF
--- a/com.unity.ml-agents/Runtime/Actuators/IActuator.cs
+++ b/com.unity.ml-agents/Runtime/Actuators/IActuator.cs
@@ -3,7 +3,7 @@ namespace Unity.MLAgents.Actuators
     /// <summary>
     /// Abstraction that facilitates the execution of actions.
     /// </summary>
-    public interface IActuator : IActionReceiver
+    public interface IActuator : IActionReceiver, IHeuristicProvider
     {
         /// <summary>
         /// The specification of the actions for this IActuator.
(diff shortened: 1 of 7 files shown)


com.unity.ml-agents/Runtime/Actuators/IActuator.cs
```diff
@@ -3,7 +3,7 @@ namespace Unity.MLAgents.Actuators
     /// <summary>
     /// Abstraction that facilitates the execution of actions.
     /// </summary>
-    public interface IActuator : IActionReceiver
+    public interface IActuator : IActionReceiver, IHeuristicProvider
```
dongruoping, line 6: Do we still need this to be a separate interface with only one method? Can it be part of IActionReceiver?

dongruoping (approved):

com.unity.ml-agents/Runtime/Actuators/IActuator.cs
surfnerd: Hmm, good point. I made IActionReceiver in order to help with backward compatibility.  Let me think about this.

surfnerd: I think for now i'd like to leave it the way it is as it allows for indirection of actions.  There might be an actuator that you want to write that delegates actions to other action receivers.  maybe in a robotics use case or something.

PR #5160, opened 2021-03-19
update c# versions to 2.0.0-exp.1

surfnerd:
### Proposed change(s)

update c# versions to 2.0.0-exp.1

### DIFF
--- a/utils/validate_versions.py
+++ b/utils/validate_versions.py
@@ -87,7 +87,7 @@ def set_version(
     if csharp_version is not None:
         package_version = f"{csharp_version}-pre.1"
         if csharp_extensions_version is not None:
-            extension_version = csharp_extensions_version + r"-pre.1"
+            extension_version = f"{csharp_extensions_version}-pre.1"
         print(
             f"Setting package version to {package_version} in {MLAGENTS_PACKAGE_JSON_PATH}"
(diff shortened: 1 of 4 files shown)


utils/validate_versions.py
```diff
@@ -85,9 +85,9 @@ def set_version(
             f.write(new_contents)
 
     if csharp_version is not None:
-        package_version = csharp_version + "-preview"
+        package_version = csharp_version + r"-pre.1"
```
chriselion, line 88: Don't think you need the preceding `r` here. Can also convert to an f-string.
```suggestion
        package_version = f"{csharp_version}-pre.1"
```

utils/validate_versions.py
```diff
@@ -85,9 +85,9 @@ def set_version(
...
         if csharp_extensions_version is not None:
-            extension_version = csharp_extensions_version + "-preview"
+            extension_version = csharp_extensions_version + r"-pre.1"
```
chriselion, line 90:
```suggestion
            extension_version = f"{csharp_extensions_version}-pre.1"
```

utils/validate_versions.py
surfnerd: ah, whoops.  this was some half-baked idea about my invalid pickle format error earlier.

chriselion (approved):

PR #5162, opened 2021-03-19
Move physics/sensor tests that fail in edit mode depending on what's in the scene to playmode tests.

surfnerd:
### Proposed change(s)

Move Edit Mode tests that fail based on what's in the scene to Play Mode tests.
This change allows us to run tests without changing the scene we are in.

### Types of change(s)

- [x] Quality of Life

### Checklist


com.unity.ml-agents/Runtime/AssemblyInfo.cs
```diff
@@ -1,6 +1,11 @@
 using System.Runtime.CompilerServices;
 
 [assembly: InternalsVisibleTo("Unity.ML-Agents.Editor.Tests")]
+[assembly: InternalsVisibleTo("Unity.ML-Agents.Runtime.Sensor.Tests")]
```
chriselion, line 4: Just curious, can we wildcard this to something like `InternalsVisibleTo("Unity.ML-Agents.*.Tests")`

surfnerd: let me try

surfnerd: looks like ['no'](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?redirectedfrom=MSDN&view=net-5.0#remarks)

chriselion (approved):

PR #5164, opened 2021-03-19
[MLA-1634] Compression spec

chriselion:
### Proposed change(s)
Change ISensor interface, and remove ISparseChannelSensor interface.

### Types of change(s)
- [x] New feature
- [x] Code refactor
- [x] Breaking change
- [x] Documentation update

### Checklist


com.unity.ml-agents/Runtime/Communicator/GrpcExtensions.cs
```diff
@@ -488,34 +488,6 @@ public static UnityRLCapabilitiesProto ToProto(this UnityRLCapabilities rlCaps)
             };
         }
 
-        internal static bool IsTrivialMapping(ISensor sensor)
```
chriselion, line 491: Moved this logic to CompressionSpec

com.unity.ml-agents/Tests/Editor/Communicator/GrpcExtensionsTests.cs
```diff
@@ -168,23 +155,6 @@ public void TestGetObservationProtoCapabilities()
 
         }
 
-        [Test]
-        public void TestIsTrivialMapping()
```
chriselion, line 172: Moved to com.unity.ml-agents/Tests/Editor/Sensor/CompressionSpecTests.cs

surfnerd (approved):
feels so much cleaner. good stuff! :shipit:

chriselion:
@dongruoping I forgot to tag you before - does this look OK with the channel mapping?

com.unity.ml-agents/Runtime/Sensors/CompressionSpec.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <param name="sensorCompressionType">The compression type to use.</param>
+        /// <param name="compressedChannelMapping">Optional mapping mapping of the channels in compressed data to the actual channel after decompression.</param>
+        /// <returns></returns>
+        public static CompressionSpec Compressed(SensorCompressionType sensorCompressionType, int[] compressedChannelMapping = null)
+        {
+            return new CompressionSpec
+            {
+                m_SensorCompressionType = sensorCompressionType,
+                m_CompressedChannelMapping = compressedChannelMapping
+            };
```
dongruoping, line 81: technically you can pass in SensorCompressionType.None and non-null mapping?

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
```diff
@@ -233,7 +227,7 @@ internal byte[] CreateEmptyPNG()
         }
 
         /// <summary>
-        /// Constrct stacked CompressedChannelMapping.
+        /// Construct stacked CompressedChannelMapping.
         /// </summary>
         internal int[] ConstructStackedCompressedChannelMapping(ISensor wrappedSenesor)
```
dongruoping, line 232: Thought a bit about moving this to the CompressionSpec constructor, don't think it's necessary here though.
But if one day we add more compression types it might make sense to move this so we can do something like CompressionSpec(PNG, num_layer=4, stacking=2) and have it automatically setup for different compression.

dongruoping (approved):
Looks good to me.

Left a few comments but are more like thoughts on what we can do later.

com.unity.ml-agents/Runtime/Sensors/CompressionSpec.cs
chriselion: Yeah, you can. I think I'll just make this the constructor instead of using a static method.

com.unity.ml-agents/Runtime/Sensors/StackingSensor.cs
chriselion: Yeah, I think we can move more utility methods into CompressionSpec (even if they stay `internal static`). I think I'll leave this here for now, but we can move it later if it makes sense (and there's another use case)

Project/Assets/ML-Agents/Examples/SharedAssets/Scripts/SensorBase.cs
```diff
@@ -51,9 +51,9 @@ public virtual byte[] GetCompressedObservation()
         }
 
         /// <inheritdoc/>
-        public virtual SensorCompressionType GetCompressionType()
+        public virtual CompressionSpec GetCompressionSpec()
         {
-            return SensorCompressionType.None;
+            return CompressionSpec.Default();
         }
```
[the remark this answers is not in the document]
chriselion: We cached the shapes previously, because they required allocating memory. I kept that pattern for the `ObservationSpec`s - it's not really necessary for performance (struct with InplaceArray, so no allocations), but it's not a bad idea since they shouldn't change at runtime. I don't think it's necessary for CompressionSpecs.

PR #5169, opened 2021-03-22
exclude forum links

chriselion:
### Proposed change(s)
A recently added link to a forum thread is causing the nightly link check to fail. `curl`ing the link indicates it returns a 307 Temporary Redirect.

We already had an exclusion for links to the forum; this just expands the exclusion.

https://github.com/Unity-Technologies/ml-agents/runs/2163855728?check_suite_focus=true

### Types of change(s)
- [x] CI

### DIFF
--- a/markdown-link-check.full.json
+++ b/markdown-link-check.full.json
@@ -13,8 +13,8 @@
             "comment": "Allow future release tags"
         },
         {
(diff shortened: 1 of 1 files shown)


surfnerd (approved):

chriselion:
Tested locally and works fine.

PR #5179, opened 2021-03-23
Move pushblock with input system integration into the Project.

surfnerd:
### Proposed change(s)

- Remove input sample project and move it into 'Project'

### Types of change(s)

- [x] Code refactor
- [x] Other (please describe) - see above.


Project/Packages/manifest.json
```diff
@@ -1,18 +1,23 @@
 {
   "dependencies": {
+    "com.unity.ide.rider": "3.0.5",
+    "com.unity.inputsystem": "1.1.0-preview.3",
     "com.unity.ml-agents": "file:../../com.unity.ml-agents",
     "com.unity.ml-agents.extensions": "file:../../com.unity.ml-agents.extensions",
     "com.unity.nuget.newtonsoft-json": "2.0.0",
     "com.unity.test-framework": "1.1.22",
+    "com.unity.toolchain.macos-x86_64-linux-x86_64": "0.1.20-preview",
```
chriselion, line 9: Should this (and the rider package) be here?

chriselion:
Is there enough duplicated code that this could changes to be a new scene in PushBlock? Although maybe the original PushBlock is going away soon (to be combined with WallJump) - @vincentpierre any thoughts?

surfnerd:
The two agents are very different since I removed the code that controls the agent out of the original push block.  I did cringe at the duplication of GoalDetect.

Project/Packages/manifest.json
surfnerd: this was added automatically.  I added rider back because it makes quite a difference in development experience.

chriselion (approved):

PR #5181, opened 2021-03-23
[MLA-1824] make SensorComponent return ISensor[]

chriselion:
### Proposed change(s)
* Changes SensorComponent to return an array of sensors instead of just 1.
* Update Match3 sensors to only operate on one of cell values or special values at a time. Match3SensorComponent will return a sensor for cell values, and another for special values (if there are any). Note that this invalidates previously trained models.

Note that Match3 models were only retrained partially, since I expect them to change again soon.

https://jira.unity3d.com/browse/MLA-1824

### Types of change(s)
- [x] Code refactor
- [x] Breaking change

### Checklist


com.unity.ml-agents.extensions/Runtime/Match3/Match3Sensor.cs
```diff
@@ -103,14 +113,14 @@ public ObservationSpec GetObservationSpec()
...
         {
-            if (m_Board.Rows != m_Rows || m_Board.Columns != m_Columns || m_Board.NumCellTypes != m_NumCellTypes)
-            {
-                Debug.LogWarning(
-                    $"Board shape changes since sensor initialization. This may cause unexpected results. " +
-                    $"Old shape: Rows={m_Rows} Columns={m_Columns}, NumCellTypes={m_NumCellTypes} " +
-                    $"Current shape: Rows={m_Board.Rows} Columns={m_Board.Columns}, NumCellTypes={m_Board.NumCellTypes}"
-                );
-            }
+            // if (m_Board.Rows != m_Rows || m_Board.Columns != m_Columns || m_Board.NumCellTypes != m_NumCellTypes)
```
chriselion, line 116: Disabled for now, might return later (depending on how I handle multiple board sizes)

com.unity.ml-agents.extensions/Runtime/Match3/Match3Sensor.cs
```diff
@@ -119,22 +129,13 @@ public int Write(ObservationWriter writer)
                 {
                     for (var c = 0; c < m_Columns; c++)
                     {
-                        var val = m_Board.GetCellType(r, c);
-                        for (var i = 0; i < m_NumCellTypes; i++)
+                        var val = m_GridValues(r, c);
+
+                        for (var i = 0; i < m_OneHotSize; i++)
```
surfnerd, line 134: does this loop exist to 0 out the rest of one-hots?  Could it be simplified to something like:
```
writer[offset + val] = 1.0f;
```

chriselion: Yeah, as long as we know the memory is initialized to zero, we could do that (plus an extra check that `val < m_OneHotSize`)

surfnerd (approved):
LGTM

PR #5206, opened 2021-03-31
[bug-fix] Fix POCA LSTM, pad sequences in the back

ervteng:
### Proposed change(s)

This PR does two things:
- Fixes an issue with POCA's LSTM when `sequence_length` < `time_horizon`.

- Moves padding of sequences shorter than `sequence_length` to the back. This is because we store the initial memories for each sequence (both policy and critic), but since we pad from the front, short sequences will be initialized, then run through some zeros. For instance, given a trajectory `[s0, s1, s2]`, we will store the initial memory `m0`. Then, when we update the model, we might have padded the trajectory `[pad0, pad1, s0, s1, s2]`, but will put the memory `m0` in the LSTM with `pad0`. 

This PR will change the padding to `[s0, s1, s2, pad0, pad1]`, so that `m0` goes with `s0`. This will also let us implement burn-in very easily by masking the loss at the start of the sequence.

Pending: experiment results (currently running Hallway). 

JIRA [MLA-1873](https://jira.unity3d.com/browse/MLA-1873)

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


ml-agents/mlagents/trainers/optimizer/torch_optimizer.py
```diff
@@ -76,53 +77,52 @@ def _evaluate_by_sequence(
...
-            for signal_name in init_values.keys()
-        }
-
+        all_values: Dict[str, List[np.ndarray]] = defaultdict(list)
+        _mem = initial_memory
         # Evaluate other trajectories, carrying over _mem after each
         # trajectory
         for seq_num in range(
-            1, math.ceil((num_experiences) / (self.policy.sequence_length))
+            0, math.floor((num_experiences) / (self.policy.sequence_length))
```
vincentpierre, line 90:
```suggestion
            math.floor((num_experiences) / (self.policy.sequence_length))
```

vincentpierre:
```suggestion
            num_experiences // self.policy.sequence_length
```

ml-agents/mlagents/trainers/optimizer/torch_optimizer.py
```diff
@@ -76,53 +77,52 @@ def _evaluate_by_sequence(
...
+        last_seq_len = leftover
+        if last_seq_len > 0:
+            for _obs in tensor_obs:
+                last_seq_obs = _obs[-last_seq_len:]
+                seq_obs.append(last_seq_obs)
+
+            # For the last sequence, the initial memory should be the one at the
+            # end of this trajectory.
+            for _ in range(last_seq_len):
+                all_next_memories.append(ModelUtils.to_numpy(_mem.squeeze()))
```
vincentpierre, line 118: Why does it matter what the memory in the padding is? Can we make this all zeros ?

ml-agents/mlagents/trainers/poca/optimizer_torch.py
```diff
@@ -381,116 +382,110 @@ def _evaluate_by_sequence_team(
         num_experiences = self_obs[0].shape[0]
         all_next_value_mem = AgentBufferField()
         all_next_baseline_mem = AgentBufferField()
-        # In the buffer, the 1st sequence are the ones that are padded. So if seq_len = 3 and
-        # trajectory is of length 10, the 1st sequence is [pad,pad,obs].
+        # In the buffer, the last sequence are the ones that are padded. So if seq_len = 3 and
```
vincentpierre, line 385: Mention in this comment that this is about LSTM and memories.

ml-agents/mlagents/trainers/poca/optimizer_torch.py
```diff
@@ -381,116 +382,110 @@ def _evaluate_by_sequence_team(
...
+        all_values: Dict[str, List[np.ndarray]] = defaultdict(list)
+        all_baseline: Dict[str, List[np.ndarray]] = defaultdict(list)
+        _baseline_mem = init_baseline_mem
+        _value_mem = init_value_mem
 
         # Evaluate other trajectories, carrying over _mem after each
         # trajectory
         for seq_num in range(
-            1, math.ceil((num_experiences) / (self.policy.sequence_length))
+            0, math.floor((num_experiences) / (self.policy.sequence_length))
```
vincentpierre, line 398: This can be simplified now.

ml-agents/mlagents/trainers/poca/optimizer_torch.py
```diff
@@ -381,116 +382,110 @@ def _evaluate_by_sequence_team(
...
                     ModelUtils.to_numpy(_baseline_mem.squeeze())
                 )
 
-            start = seq_num * self.policy.sequence_length - (
-                self.policy.sequence_length - leftover
-            )
-            end = (seq_num + 1) * self.policy.sequence_length - (
-                self.policy.sequence_length - leftover
-            )
+            start = seq_num * self.policy.sequence_length
```
vincentpierre, line 406: This feels like there is a bunch of duplicate code with this part : https://github.com/Unity-Technologies/ml-agents/pull/5206/files#diff-01b42574a4de05de250e29039fc1bbc67b200f3a7c5ed0676b2eee549b943c11R95
Is it possible to combine some of it ?

ml-agents/mlagents/trainers/poca/optimizer_torch.py
```diff
@@ -381,116 +382,110 @@ def _evaluate_by_sequence_team(
...
+                groupmate_seq_act.append(_act)
+
+            # For the last sequence, the initial memory should be the one at the
+            # beginning of this trajectory.
+            seq_obs = []
+            last_seq_len = leftover
+            for _ in range(last_seq_len):
+                all_next_value_mem.append(ModelUtils.to_numpy(_value_mem.squeeze()))
+                all_next_baseline_mem.append(
+                    ModelUtils.to_numpy(_baseline_mem.squeeze())
```
vincentpierre, line 471: Why are we padding with this value and not NaN or zeros ?

ml-agents/mlagents/trainers/optimizer/torch_optimizer.py
```diff
@@ -76,53 +77,52 @@ def _evaluate_by_sequence(
...
                 seq_obs, _mem, sequence_length=self.policy.sequence_length
             )
             for signal_name, _val in values.items():
                 all_values[signal_name].append(_val)
+
+        # Compute values for the potentially truncated last sequence
+        seq_obs = []
+
+        last_seq_len = leftover
+        if last_seq_len > 0:
```
andrewcoh, line 110:
```suggestion
        if leftover > 0:
```

ml-agents/mlagents/trainers/optimizer/torch_optimizer.py
```diff
@@ -76,53 +77,52 @@ def _evaluate_by_sequence(
...
+            for _obs in tensor_obs:
+                last_seq_obs = _obs[-last_seq_len:]
```
andrewcoh, line 112: nit: why use `last_seq_len` instead of `leftover`. Or rename leftover when its first created.

ml-agents/mlagents/trainers/optimizer/torch_optimizer.py
```diff
@@ -76,53 +77,52 @@ def _evaluate_by_sequence(
...
+                seq_obs.append(last_seq_obs)
+
+            # For the last sequence, the initial memory should be the one at the
+            # end of this trajectory.
+            for _ in range(last_seq_len):
```
andrewcoh, line 117: Why do we add the last `_mem` for each extra leftover obs?

ml-agents/mlagents/trainers/optimizer/torch_optimizer.py
ervteng: This isn't in the padding, as the padding isn't added to the trajectory until later. This method only has unpadded data. Added a comment

ml-agents/mlagents/trainers/optimizer/torch_optimizer.py
ervteng: We need the mem array to be the same length as all the other ones in the buffer. But you're right this isn't needed. To address @vincentpierre's comment above, we could also add `_mem` once and add a bunch of zeros/NaNs, but this saves us from having to create/allocate those empty arrays.

ml-agents/mlagents/trainers/poca/optimizer_torch.py
ervteng: It's annoyingly similar but just slightly different enough to make it really hard to combine. The POCA version has to operate on group obs and actions as well 🤔

ml-agents/mlagents/trainers/optimizer/torch_optimizer.py
vincentpierre: I don't understand that comment :
```
            # For the last sequence, the initial memory should be the one at the
            # end of this trajectory.
```

 Can you give more info ?

ervteng:
Experimental update: 

image
 Hallway trains a little faster than on `main` (green line is `main`, 2 others are 2 runs of this PR). 
 
image
Value estimates are much cleaner. Value loss shows a similar difference. I suspect this is b/c of the lack of random-length zero-pad at the beginning of the sequence.

ml-agents/mlagents/trainers/poca/optimizer_torch.py
ervteng: The sequence at this point isn't padded - it's just the leftover bit that would go into a new sequence. E.g. if there were three sequences of length "2" `[[s0, s1],[s2, s3],[s4]]`, at this point this is just `s4`. When we pad in the buffer, it will become [s4, 0], but not at this point. 

With that said, only the 1st memory of each sequence is used. I'm repeating the memory rather than using zeros or ones so we don't need to allocate a new numpy array every time.

config/ppo/Hallway.yaml
```diff
@@ -5,7 +5,7 @@ behaviors:
       batch_size: 128
       buffer_size: 1024
       learning_rate: 0.0003
-      beta: 0.01
+      beta: 0.03
```
andrewcoh, line 8: Should we add in the documentation that it may be good practice to increase beta with LSTM? Or do we think that this is a peculiarity of Hallway?

ml-agents/mlagents/trainers/optimizer/torch_optimizer.py
```diff
@@ -76,53 +76,52 @@ def _evaluate_by_sequence(
...
-
-        init_values, _mem = self.critic.critic_pass(
-            seq_obs, initial_memory, sequence_length=first_seq_len
-        )
-        all_values = {
-            signal_name: [init_values[signal_name]]
-            for signal_name in init_values.keys()
-        }
-
+        # When using LSTM, we need to divide the trajectory into sequences of even length. Sometimes,
```
andrewcoh, line 79:
```suggestion
        # When using LSTM, we need to divide the trajectory into sequences of equal length. Sometimes,
```

ml-agents/mlagents/trainers/poca/optimizer_torch.py
```diff
@@ -381,116 +381,109 @@ def _evaluate_by_sequence_team(
...
-
-        all_seq_obs = self_seq_obs + groupmate_seq_obs
-        init_values, _value_mem = self.critic.critic_pass(
-            all_seq_obs, init_value_mem, sequence_length=first_seq_len
-        )
-        all_values = {
-            signal_name: [init_values[signal_name]]
-            for signal_name in init_values.keys()
-        }
+        # When using LSTM, we need to divide the trajectory into sequences of even length. Sometimes,
```
andrewcoh, line 385:
```suggestion
        # When using LSTM, we need to divide the trajectory into sequences of equal length. Sometimes,
```

andrewcoh (approved):

config/ppo/Hallway.yaml
ervteng: I think it's b/c of Hallway's sparse reward and reward structure that has a bunch of local maxima (at least 2, given the reward curve). Hard to say if we'd need to increase beta for other envs.
Our betas might just be too low across-the-board.

ml-agents/mlagents/trainers/optimizer/torch_optimizer.py
```diff
@@ -76,53 +76,52 @@ def _evaluate_by_sequence(
...
+        leftover_seq_len = num_experiences % self.policy.sequence_length
+
+        all_values: Dict[str, List[np.ndarray]] = defaultdict(list)
+        _mem = initial_memory
         # Evaluate other trajectories, carrying over _mem after each
         # trajectory
-        for seq_num in range(
-            1, math.ceil((num_experiences) / (self.policy.sequence_length))
-        ):
+        for seq_num in range(num_experiences // (self.policy.sequence_length)):
```
vincentpierre, line 90:
```suggestion
        for seq_num in range(num_experiences // self.policy.sequence_length):
```

vincentpierre (approved):

PR #5209, opened 2021-03-31
Separate grid sensor and component

dongruoping:
### Proposed change(s)

GridSensor refactor will come in three PR: 1. separate sensor and component with no behavior change (this). 2. change the design and improve interface and also fix bugs. 3. move it into main repo and update namespace. update the docs.

Main changes:
* Separate sensor and component. Add editor UI for new component.
* ~~Change variable names to better follow naming convention.~~ leave it to another PR for better diff
* Remove unused method/variables and simplify some methods.
* Only expose necessary public methods and make all the others private. Same for variables.
* Make Counting Grid Sensor become one grid depth type instead of another sensor. (More changes to this might come later)

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


com.unity.ml-agents.extensions/Editor/GridSensorComponentEditor.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            EditorGUI.BeginDisabledGroup(!EditorUtilities.CanUpdateModelProperties());
+            {
+                // These fields affect the sensor order or observation size,
+                // So can't be changed at runtime.
+                EditorGUILayout.PropertyField(so.FindProperty("m_SensorName"), true);
+
+                EditorGUILayout.LabelField("Grid Settings", EditorStyles.boldLabel);
+                EditorGUILayout.PropertyField(so.FindProperty("m_CellScaleX"), true);
+                EditorGUILayout.PropertyField(so.FindProperty("m_CellScaleZ"), true);
```
chriselion, line 28: X, Y, Z instead of X, Z, Y? I understand it's probably that way because (X, Z) are the plane that the grid is in, but it's counter-intuitive.

chriselion: Also, just for screen real estate, can we see if it's easy put the cell scale and grid num size fields on the same line? Maybe user a Vector3 for the scale?

chriselion: and Vector2Int for the GridNumSide

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorComponent.cs
```diff
@@ -0,0 +1,326 @@
...
+        {
+            get { return m_DetectableObjects; }
+            set { m_DetectableObjects = value; }
+        }
+
+        [HideInInspector, SerializeField]
+        LayerMask m_ObserveMask;
+        /// <summary>
+        /// The layer mask.
+        /// Note that changing this after the sensor is created has no effect.
```
chriselion, line 132: Is this true, or did it get copy-pasted? It seems like something it would be OK to change at runtime (although it may not give good results).

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorComponent.cs
```diff
@@ -0,0 +1,326 @@
...
+        /// </summary>
+        public SensorCompressionType CompressionType
+        {
+            get { return m_CompressionType; }
+            set { m_CompressionType = value; UpdateSensor(); }
+        }
+
+        [HideInInspector, SerializeField]
+        [Range(1, 50)]
+        [Tooltip("Number of camera frames that will be stacked before being fed to the neural network.")]
```
chriselion, line 238:
```suggestion
        [Tooltip("Number of frames of observations that will be stacked before being fed to the neural network.")]
```

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorComponent.cs
```diff
@@ -0,0 +1,326 @@
...
+                m_GridNumSideZ,
+                m_RotateWithAgent,
+                m_ChannelDepth,
+                m_DetectableObjects,
+                m_ObserveMask,
+                m_DepthType,
+                RootReference,
+                m_MaxColliderBufferSize,
+                m_InitialColliderBufferSize,
+                m_DebugColors,
```
chriselion, line 268: Can m_DebugColors and m_ShowGizmos be removed from GridSensor and kept to the component? (OK if that comes in a later PR).

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
```diff
@@ -7,447 +8,221 @@
...
-            m_ChannelHotDefaultPerceptionBuffer = new float[ObservationPerCell];
-            for (int i = 0; i < ChannelDepth.Length; i++)
-            {
-                if (ChannelDepth[i] > 1)
-                {
-                    m_ChannelHotDefaultPerceptionBuffer[ChannelOffsets[i]] = 1;
-                }
-            }
+            // The maximum number of channels in the final output must be less than 255 * 3 because the "number of PNG images" to generate must fit in one byte
+            Assert.IsTrue(m_ObservationPerCell < (255 * 3), "The maximum number of channels per cell must be less than 255 * 3");
```
chriselion, line 193: Is this still true? I think it might be from an older iteration of the code (before we supported multiple PNGs

com.unity.ml-agents.extensions/Editor/GridSensorComponentEditor.cs
dongruoping: I was going to change these name to something like length, width, height later.
But sure I can make them Vector3 and Vector2Int

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorComponent.cs
dongruoping: yes I've removed logic using m_DebugColors from the sensor, just need to remove it from constructor

The sensor needs m_ShowGizmos though to decide whether it has to fill the color indices buffer.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
dongruoping: I'm not sure where this constraint of "one byte" comes from. When GridSensor was merged it already supports multiple PNGs. 

`m_ObservationPerCell < (255 * 3)` is saying that we can't send observation that's more than 255 PNG images concatenated. Maybe it's something to do with grpc message size?

com.unity.ml-agents.extensions/Editor/GridSensorComponentEditor.cs
```diff
@@ -0,0 +1,140 @@
...
+using Unity.MLAgents.Editor;
+using Unity.MLAgents.Extensions.Sensors;
+
+namespace Unity.MLAgents.Extensions.Editor
+{
+    [CustomEditor(typeof(GridSensorComponent))]
+    [CanEditMultipleObjects]
+    internal class GridSensorComponentEditor : UnityEditor.Editor
+    {
+        const string k_SensorName = "m_SensorName";
```
chriselion, line 12: I know we don't do this elsewhere right now, but I think you can do e.g. `nameof(GridSensor.m_SensorName)` (although that might not work if it's private, so you'd have to change it to internal).

Unfortunately these string constants don't really do much, because they're not used in multiple places, and things will still break if the member gets renamed.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
chriselion: Right, but I think the comment itself is leftover from the first attempt that the PR, which was before multiple PNGs were supported.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
```diff
@@ -7,16 +8,21 @@
...
 {
+    /// <summary>
+    /// Enum describing what kind of depth type the data should be organized as
+    /// </summary>
+    public enum GridDepthType { Channel, ChannelHot, Counting };
+
     /// <summary>
     /// Grid-based sensor.
     /// </summary>
     [AddComponentMenu("ML Agents/Grid Sensor", (int)MenuGroup.Sensors)]
```
surfnerd, line 19: this should move to the component class.

dongruoping: moved it there and forgot to remove here, thanks

chriselion (approved):
Looks good! We identified some extra things offline that can be cleaned up later, but for the purposes of splitting this in half, I think this is a good start.

com.unity.ml-agents.extensions/Editor/GridSensorComponentEditor.cs
dongruoping: I ended up making GridNumSide Vector3/Vector3Int and is only configurable on x and z, in case we want 3D GridSensor one day.

com.unity.ml-agents.extensions/Editor/GridSensorComponentEditor.cs
dongruoping: changed it to use nameof()

PR #5217, opened 2021-04-01
Rename variables in Grid Sensor

dongruoping:
### Proposed change(s)

Renaming variables to better follow naming conventions.

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


com.unity.ml-agents.extensions/Runtime/Sensors/GridSensor.cs
```diff
@@ -20,174 +20,73 @@ public enum GridDepthType { Channel, ChannelHot, Counting };
...
-        /// </summary>
-        int ObservationPerCell;
-
-        /// <summary>
-        /// The offsets used to specify where within a cell's allotted data, certain observations will be inserted.
-        /// </summary>
-        int[] ChannelOffsets;
+        string m_Name;
+        Vector3 m_CellScale;
+        Vector3Int m_GridNum;
```
dongruoping, line 25: I explored grouping these parameters into struct. The constructer became cleaner but the code becomes much more verbose since these are used a lot everywhere. Also that brings another problem that in the struct some parameters can be changed at runtime while some cannot, and that will make it even more complicated.
So I decided to leave it as is.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorComponent.cs
```diff
@@ -37,23 +37,23 @@ public Vector3 CellScale
         }
 
         [HideInInspector, SerializeField]
-        internal Vector3Int m_GridNumSide = new Vector3Int(16, 1, 16);
+        internal Vector3Int m_GridNum = new Vector3Int(16, 1, 16);
```
surfnerd, line 40: Could this be called GridDimensions, or GridSize? GridNum is confusing to me.

surfnerd (approved):
Small comment about the `m_GridNum` variable name, otherwise LGTM

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorComponent.cs
dongruoping: I was thinking this as "number of grids" on each side. I'll change it to GridSize.

chriselion (approved):
Looks good, just make sure you're keeping a rough list of API changes so that we can add them to the Migration Guide

PR #5222, opened 2021-04-05
[MLA-1880] Raycast sensor interface improvements

chriselion:
### Proposed change(s)
Public-facing change: 
* Adds `RayPerceptionSensor.RayPerceptionOutput`, which contains the most recent ray hit information. This can be used to for additional game logic as needed.
* Adds new fields to `RayOutput`: `StartPositionWorld`, `EndPositionWorld`, `ScaledRayLength`, and `ScaledCastRadius`.

Internal changes:
* Get rid of DebugDisplayInfo - all of its info is now on `RayOutput`, except for the frame that the results were queried, which is now stored as `m_DebugLastFrameCount`
* Raycasts now happen in the Update() call, which always happens before the Write().

Based on feedback from several forum posts:

Forum threads:

- ["i want to do an if statement inside my agent script to say if the ray cast hits an object with the tag enemy then do something"](https://forum.unity.com/threads/getting-the-detectable-tags-from-ray-perception-sensor-script.834733/)
- ["does the ray perception 3d component get the distance to the object it impacted? I need my agents to get the distance to that object"](https://forum.unity.com/threads/ray-perception-sensor-component-3d-distance.867223/)
- ["we have to reset agent if it follows wrong raycast and crash, or we want to add reward if it follows correct raycast."](https://forum.unity.com/threads/how-can-i-use-rayperceptioncomponent3d-effectively.926426/#post-6086127)
- ["I am trying to teach an agent to find a target but once it has "seen" the target, I want to give it a reward."](https://forum.unity.com/threads/rewarding-an-agent-when-it-sees-something.955029/)
- ["I was wondering, if there is an easy way to draw the Debug gizmos of the Ray Perception Sensor ingame for demonstration purposes?"](https://forum.unity.com/threads/way-to-visualize-ray-perception-sensor.978642/)
- ["I'd like to get a value of distance between the agent and obstacles with using Ray Perception Sensor."](https://forum.unity.com/threads/how-to-get-value-of-distance-between-agent-obstacles-with-using-rayperceptionsensor3d.1043338/)
- ["I use three Ray Perception Sensor Components for measuring distances from the agent to some objects, and I'd like to get all values (HasHit, HitTaggedObject, HitTagIndex and HitFraction) from all components."](https://forum.unity.com/threads/how-to-get-values-from-multiple-ray-perception-sensor-3ds.1045960/)

### Types of change(s)
- [x] New feature
- [x] Code refactor

### Checklist

### DIFF
--- a/Project/ProjectSettings/ProjectSettings.asset
+++ b/Project/ProjectSettings/ProjectSettings.asset
@@ -125,7 +125,8 @@ PlayerSettings:
     16:9: 1
     Others: 1
   bundleVersion: 1.0
-  preloadedAssets: []
+  preloadedAssets:
+  - {fileID: 0}
   metroInputSource: 0
   wsaTransparentSwapchain: 0
   m_HolographicPauseOnTrackingLoss: 1
--- a/Project/ProjectSettings/ProjectSettings.asset
+++ b/Project/ProjectSettings/ProjectSettings.asset
@@ -125,8 +125,7 @@ PlayerSettings:
     16:9: 1
     Others: 1
   bundleVersion: 1.0
-  preloadedAssets:
-  - {fileID: 0}
+  preloadedAssets: []
   metroInputSource: 0
   wsaTransparentSwapchain: 0
   m_HolographicPauseOnTrackingLoss: 1
--- a/com.unity.ml-agents/Runtime/Sensors/RayPerceptionSensor.cs
+++ b/com.unity.ml-agents/Runtime/Sensors/RayPerceptionSensor.cs
@@ -235,37 +235,6 @@ namespace Unity.MLAgents.Sensors
     }
 
     /// <summary>
-    /// Debug information for the raycast hits. This is used by the RayPerceptionSensorComponent.
-    /// </summary>
-    internal class DebugDisplayInfo
-    {
-        public struct RayInfo
-        {
-            // public Vector3 worldStart;
-            // public Vector3 worldEnd;
-            //public float castRadius;
-            public RayPerceptionOutput.RayOutput rayOutput;
-        }
-
-        public void Reset()
-        {
-            m_Frame = Time.frameCount;
-        }
-
-        /// <summary>
-        /// "Age" of the results in number of frames. This is used to adjust the alpha when drawing.
-        /// </summary>
-        public int age
-        {
-            get { return Time.frameCount - m_Frame; }
-        }
-
-        public RayInfo[] rayInfos;
-
-        int m_Frame;
-    }
-
-    /// <summary>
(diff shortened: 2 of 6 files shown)


Project/ProjectSettings/ProjectSettings.asset
```diff
@@ -125,7 +125,8 @@ PlayerSettings:
     16:9: 1
     Others: 1
   bundleVersion: 1.0
-  preloadedAssets: []
+  preloadedAssets:
+  - {fileID: 0}
```
dongruoping, line 129: relevant?

com.unity.ml-agents/Runtime/Sensors/RayPerceptionSensorComponentBase.cs
```diff
@@ -252,16 +252,26 @@ internal void UpdateSensor()
...
+                return Time.frameCount - m_RaySensor.DebugLastFrameCount;
+            }
+
+            return 0;
+        }
+
         void OnDrawGizmosSelected()
         {
-            if (m_RaySensor?.debugDisplayInfo?.rayInfos != null)
+            if (m_RaySensor?.m_rayPerceptionOutput?.RayOutputs != null)
```
dongruoping, line 267:
```suggestion
            if (m_RaySensor?.RayPerceptionOutput?.RayOutputs != null)
```

com.unity.ml-agents/Runtime/Sensors/RayPerceptionSensor.cs
```diff
@@ -241,12 +243,16 @@ public class RayPerceptionSensor : ISensor, IBuiltInSensor
         string m_Name;
 
         RayPerceptionInput m_RayPerceptionInput;
+        internal RayPerceptionOutput m_rayPerceptionOutput;
```
dongruoping, line 246:
```suggestion
        internal RayPerceptionOutput m_RayPerceptionOutput;
```

dongruoping (approved):
Looks good to me. Verified that the gizmos work fine in editor.

com.unity.ml-agents/Runtime/Sensors/RayPerceptionSensorComponentBase.cs
chriselion: Thanks, I'll make `m_{r,R}ayPerceptionOutput` private (and fix the capitalization)

Project/ProjectSettings/ProjectSettings.asset
chriselion: `git add -u` strikes again.

PR #5227, opened 2021-04-07
[MLA-1129] Clear ActionBuffers before Heuristic calls

chriselion:
### Proposed change(s)
Reset the ActionBuffers before calls to `Agent.Heuristic`and `IHeuristicProvider.Heuristic`. Whether or not the actions carried over between steps before was unspecified, not they're always initialized to zero.

Also clean up cases in the examples where we needed to work around the old behavior.

https://jira.unity3d.com/browse/MLA-1129
https://jira.unity3d.com/browse/MLA-1123

### Types of change(s)
- [x] Code refactor
- [x] Breaking change

### Checklist


chriselion:
@surfnerd Anything that needs to change in the input system actuators? I had a quick look and it didn't look like it.

andrewcoh (approved):

surfnerd:
> @surfnerd Anything that needs to change in the input system actuators? I had a quick look and it didn't look like it.

I don't think so.  The input actuator just writes whatever input state in the heuristic method every time.  So there was no need to do any sort of `clear` in that case.

surfnerd (approved):

PR #5233, opened 2021-04-08
[MLA-1909] Match3 and Camera/RenderTexture sensor GC improvements

chriselion:
### Proposed change(s)
* Reuse `Texture2D`s to avoid allocations
* Agent now calls `Dispose()` on any `ISensor`s that implement `IDisposable`.
* Special-case RGB24 texture->ObservationWriter (which is what Camera and RenderTexture Sensors user for inference) to avoid GetPixels32(), which allocates a copy of the pixel data.

#### Match3Sensor
Before
image
After
image

#### GridWorld (Camera + RenderTexture Sensors)
Before
image
After
image

https://jira.unity3d.com/browse/MLA-1909
https://jira.unity3d.com/browse/MLA-1910
https://jira.unity3d.com/browse/MLA-208

### Types of change(s)

- [x] Bug fix
- [x] New feature
- [x] Code refactor

### Checklist

### DIFF
--- a/com.unity.ml-agents.extensions/Runtime/Match3/Match3Sensor.cs
+++ b/com.unity.ml-agents.extensions/Runtime/Match3/Match3Sensor.cs
@@ -5,7 +5,6 @@ using UnityEngine;
 
 namespace Unity.MLAgents.Extensions.Match3
 {
-
     /// <summary>
     /// Delegate that provides integer values at a given (x,y) coordinate.
     /// </summary>
@@ -167,7 +166,6 @@ namespace Unity.MLAgents.Extensions.Match3
 
 
             return offset;
-
         }
 
         /// <inheritdoc/>
@@ -243,6 +241,9 @@ namespace Unity.MLAgents.Extensions.Match3
             return BuiltInSensorType.Match3Sensor;
         }
 
+        /// <summary>
+        /// Clean up the owned Texture2D.
+        /// </summary>
         public void Dispose()
         {
             if (!ReferenceEquals(null, m_ObservationTexture))
@@ -279,7 +280,7 @@ namespace Unity.MLAgents.Extensions.Match3
             int channelOffset,
             int currentHeight,
             int currentWidth
-            )
+        )
         {
             var i = 0;
             // There's an implicit flip converting to PNG from texture, so make sure we
--- a/com.unity.ml-agents.extensions/Runtime/Match3/Match3Sensor.cs
+++ b/com.unity.ml-agents.extensions/Runtime/Match3/Match3Sensor.cs
@@ -1,7 +1,9 @@
+using System;
 using System.Collections.Generic;
 using Unity.MLAgents.Sensors;
 using UnityEngine;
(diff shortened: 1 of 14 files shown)


com.unity.ml-agents.extensions/Runtime/Match3/Match3Sensor.cs
```diff
@@ -246,6 +257,15 @@ static void DestroyTexture(Texture2D texture)
                 Object.Destroy(texture);
             }
         }
+
+        public void Dispose()
+        {
+            if (!ReferenceEquals(null, m_ObservationTexture))
+            {
+                DestroyTexture(m_ObservationTexture);
```
dongruoping, line 265: Is there a place to put sensor utilities and maybe we could move it there? I'm going to add the same thing to GridSensor and that will be the third time we define this method.

chriselion: Let's just put it in Utilities.cs

chriselion: 4th time, actually 😄 - Match3, Camera, and RenderTexture sensors already have it

chriselion: Moved.

com.unity.ml-agents/Runtime/Agent.cs
```diff
@@ -1004,6 +1006,19 @@ internal void InitializeSensors()
 #endif
         }
 
+        void CleanupSensors()
+        {
+            // Dispose all attached sensor
+            for (var i = 0; i < sensors.Count; i++)
+            {
+                var sensor = sensors[i];
+                if (sensor is IDisposable disposableSensor)
```
surfnerd, line 1015: fannnncy

com.unity.ml-agents/Runtime/Sensors/RenderTextureSensorComponent.cs
```diff
@@ -102,5 +104,17 @@ internal void UpdateSensor()
...
         }
+
+        /// <summary>
+        /// Clean up the sensor created by CreateSensors().
+        /// </summary>
+        public void Dispose()
+        {
+            if (!ReferenceEquals(null, m_Sensor))
+            {
+                m_Sensor.Dispose();
```
surfnerd, line 115: could this be a double dispose?  looks safe since the sensor checks for null.  Just curious if it's actually needed or not.

surfnerd (approved):
looks good, just not sure if the extra dispose in the components are needed anymore since the sensor disposal happens in Agent.

com.unity.ml-agents/Runtime/Sensors/RenderTextureSensorComponent.cs
chriselion: I made sure all the sensor.Dispose's are idempotent - they check for null textures first, then set the texture to null.

PR #5238, opened 2021-04-09
Separate boxoverlap out of GridSensor

dongruoping:
### Proposed change(s)

Separate BoxOverlap logic into separate class, and also make each encoding into a class derived from GridSensorBase.
The component holds a list of GridSensors, and the first one has the BoxOverlapChecker.
The component also has a dummy GridSensorBase which will not be passed to Agent, only for collecting debug gizmo data.

For each cell, the data collection process is as following:
1. BoxOverlap: Detect all colliders within the cell
2. ParseColliders: Iterate through all colliders and determine which ones will actually be used. Currently it either use the one closest to agent or use all.
3. LoadObjectData: for each parsed collider, load the observation data for that game object. It calls GetObjectData which actually fills the data buffer.

Derived GridSensors share 1 and 2, and only overwrites 3.
1 only runs once for each cell.
2 runs once for each way of parsing colliders. This allows the parsed colliders being reused by different sensors. For example, there's a one-hot sensor and a continuous value sensor both using one collider per cell. Then ParseColliders only has to compute closest collider once.
3 runs once for each sensor and loads different data for different sensor.

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [x] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorComponent.cs
```diff
@@ -202,68 +194,125 @@ public int ObservationStacks
             set { m_ObservationStacks = value; }
         }
 
+        [HideInInspector, SerializeField]
+        internal bool m_UseOneHotTag = true;
+        /// <summary>
+        /// Whether to use one-hot detected tag as observation.
+        /// Note that changing this after the sensor is created has no effect.
+        /// </summary>
+        public bool UseOneHotTag
```
dongruoping, line 203: I can make `UseOneHotTag` and `CountColliders` into a GridSensorObservationType enum if that's better.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorComponent.cs
```diff
@@ -202,68 +194,125 @@ public int ObservationStacks
...
                 gameObject,
-                m_CompressionType,
-                m_MaxColliderBufferSize,
-                m_InitialColliderBufferSize
+                m_DetectableObjects,
+                m_InitialColliderBufferSize,
+                m_MaxColliderBufferSize
             );
+            // debug data is positive int value and will trigger data validation exception if SensorCompressionType is not None.
+            m_DebugSensor = new GridSensorBase("DebugGridSensor", m_CellScale, m_GridSize, new int[] { 1 }, m_DetectableObjects, SensorCompressionType.None);
```
chriselion, line 236: Maybe lazily allocate this in OnDrawGizmos()?

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorBase.cs
```diff
@@ -0,0 +1,336 @@
...
+        Vector3Int m_GridSize;
+        int[] m_ChannelDepths;
+        string[] m_DetectableObjects;
+        SensorCompressionType m_CompressionType;
+        ObservationSpec m_ObservationSpec;
+        internal BoxOverlapChecker m_BoxOverlapChecker;
+
+        // Buffers
+        float[] m_PerceptionBuffer;
+        Color[] m_PerceptionColors;
```
chriselion, line 29: Now that you're separating out the sensor for gizmo drawing, can you move this into a subclass of GridSensorBase?

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorBase.cs
```diff
@@ -0,0 +1,336 @@
...
+        protected string[] DetectableObjects
+        {
+            get { return m_DetectableObjects; }
+        }
+
+        /// <inheritdoc/>
+        public void Reset() { }
+
+        /// <summary>
+        /// Clears the perception buffer before loading in new data. If the gridDepthType is ChannelHot, then it initializes the
```
chriselion, line 97: I think this comment ("If the gridDepthType is ChannelHot...") is out of date now.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorBase.cs
```diff
@@ -0,0 +1,336 @@
...
+        /// </example>
+        protected virtual void GetObjectData(GameObject currentColliderGo, int typeIndex, float[] dataBuffer)
+        {
+            dataBuffer[0] = typeIndex;
+        }
+
+        /// <summary>
+        /// Get the observation size for each cell. This will be the size of dataBuffer in GetObjectData().
+        /// </summary>
+        protected virtual int GetObservationSize()
```
chriselion, line 217: How about something like `GetCellObservationSize()` to make it clear that it's not the observation size for the whole sensor?

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorBase.cs
```diff
@@ -0,0 +1,336 @@
...
+        }
+
+        /// <summary>
+        /// LoadObjectData - If the GameObject matches a tag, GetObjectData is called to extract the data from the GameObject
+        /// then the data is transformed based on the GridDepthType of the gridsensor.
+        /// Further documetation on the GridDepthType can be found below
+        /// </summary>
+        /// <param name="currentColliderGo">The game object that was found colliding with a certain cell</param>
+        /// <param name="cellIndex">The index of the current cell</param>
+        protected internal void LoadObjectData(GameObject currentColliderGo, int cellIndex)
```
chriselion, line 257: Rename to `ProcessCellOverlaps()`?

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorBase.cs
```diff
@@ -0,0 +1,336 @@
...
+        {
+            Profiler.BeginSample("GridSensor.LoadObjectData");
+            for (var i = 0; i < m_DetectableObjects.Length; i++)
+            {
+                if (!ReferenceEquals(currentColliderGo, null) && currentColliderGo.CompareTag(m_DetectableObjects[i]))
+                {
+                    Array.Clear(m_CellDataBuffer, 0, m_CellDataBuffer.Length);
```
chriselion, line 264: This feels like a a good usecase for an ArraySegment that's a "view' on the part of m_CellDataBuffer that we're able to write to. Now that we're on 2019.4, we might be able to use a new enough version of C# that we can write, otherwise we could roll our own small version

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorBase.cs
dongruoping: I don't quite understand this one - m_PerceptionBuffer and m_PerceptionColors are both used in general grid sensors. The data will go from `m_PerceptionBuffer` (float) -> `m_PerceptionColors (Colors)` -> `m_PerceptionTexture.SetPixels(m_PerceptionColors)`

chriselion: My misunderstand - I thought m_PerceptionColors was only used for gizmos.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorComponent.cs
dongruoping: That will cause several issues. When there's any changes on component, the debug sensor has to be updated too. So if we turn on gizmo -> turn it off -> change some config -> and turn it back on, it will not be updated. 

Also decoupling the creation of DebugSensor and BoxOverlapChecker cause errors in editor. m_DebugSensor will attach to the existing BoxOverlapChecker at creation time, but editor refresh will keep re-creating new BoxOverlapChecker and the debug sensor won't know about it. 

I think in terms of performance it's not causing too much trouble since when showGizmo=false it doesn't trigger any unnecessary calls.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorBase.cs
```diff
@@ -0,0 +1,336 @@
...
+{
+    /// <summary>
+    /// Grid-based sensor.
+    /// </summary>
+    public class GridSensorBase : ISensor, IBuiltInSensor, IDisposable
+    {
+        string m_Name;
+        Vector3 m_CellScale;
+        Vector3Int m_GridSize;
+        int[] m_ChannelDepths;
```
chriselion, line 21: Are m_ChannelDepths still needed? I can see how a custom implementation might need them, but not the ones you have now. Seems like it would eliminate some of the confusing parts of the interface.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorComponent.cs
```diff
@@ -202,68 +194,129 @@ public int ObservationStacks
...
+                m_DetectableObjects,
+                m_InitialColliderBufferSize,
+                m_MaxColliderBufferSize
             );
 
-            if (ObservationStacks != 1)
+            if (m_UseOneHotTag)
             {
-                return new ISensor[] { new StackingSensor(m_Sensor, ObservationStacks) };
+                var sensor = new OneHotGridSensor(m_SensorName, m_CellScale, m_GridSize, m_ChannelDepths, m_DetectableObjects, m_CompressionType);
```
chriselion, line 238: note that you'll need to append something to the name either here or in the implementations, otherwise you can get two sensors with the same name.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorBase.cs
dongruoping: yes I noticed that too and I removed it

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorBase.cs
dongruoping: Renamed to `ProcessDetectedObject` since it processes one object

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorComponent.cs
dongruoping: added

com.unity.ml-agents.extensions/Editor/GridSensorComponentEditor.cs
```diff
@@ -82,14 +64,21 @@ public override void OnInspectorGUI()
             }
             EditorGUI.EndDisabledGroup();
             EditorGUILayout.PropertyField(so.FindProperty(nameof(GridSensorComponent.m_CompressionType)), true);
+            EditorGUI.BeginDisabledGroup(!EditorUtilities.CanUpdateModelProperties());
```
chriselion, line 67: nit: don't think these need to be disabled during play mode, since they don't affect observation size.

com.unity.ml-agents.extensions/Runtime/Sensors/BoxOverlapChecker.cs
```diff
@@ -0,0 +1,247 @@
...
+            GameObject closestColliderGo = null;
+            var minDistanceSquared = float.MaxValue;
+
+            for (var i = 0; i < numFound; i++)
+            {
+                var currentColliderGo = foundColliders[i].gameObject;
+
+                // Continue if the current collider go is the root reference
+                if (ReferenceEquals(currentColliderGo, m_RootReference))
+                    continue;
```
chriselion, line 187: nit:
```suggestion
                {
                    continue;
                }
```
(this is probably from the original code)

com.unity.ml-agents.extensions/Runtime/Sensors/BoxOverlapChecker.cs
```diff
@@ -0,0 +1,247 @@
...
+
+                var closestColliderPoint = foundColliders[i].ClosestPointOnBounds(cellCenter);
+                var currentDistanceSquared = (closestColliderPoint - m_RootReference.transform.position).sqrMagnitude;
+
```
chriselion, line 191: optional: I would early out of the loop iteration sooner, to avoid tag comparison if possible. E.g.

```suggestion
                if(currentDistanceSquared >= minDistanceSquared)
                {
                    continue;
                }
```
(I know, now the intent of the PR, just noticed it while reviewing)

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorComponent.cs
```diff
@@ -70,27 +76,15 @@ public bool RotateWithAgent
...
-        /// </summary>
-        public int[] ChannelDepths
-        {
-            get { return m_ChannelDepths; }
-            set { m_ChannelDepths = value; }
-        }
-
-        [HideInInspector, SerializeField]
-        internal string[] m_DetectableObjects;
+        internal string[] m_DetectableTags;
```
chriselion, line 79: Good name change!

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorComponent.cs
```diff
@@ -202,68 +184,149 @@ public int ObservationStacks
...
+                m_InitialColliderBufferSize,
+                m_MaxColliderBufferSize
             );
 
-            if (ObservationStacks != 1)
+            // debug data is positive int value and will trigger data validation exception if SensorCompressionType is not None.
+            m_DebugSensor = new GridSensorBase("DebugGridSensor", m_CellScale, m_GridSize, m_DetectableTags, SensorCompressionType.None);
+            m_BoxOverlapChecker.RegisterDebugSensor(m_DebugSensor);
+
+            if (m_UseOneHotTag)
```
chriselion, line 230: Move the creation of OneHotGridSensor and CountingGridSensor to the default impl of GetGridSensors?

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorComponent.cs
```diff
@@ -202,68 +184,149 @@ public int ObservationStacks
...
+                    }
+                    else
+                    {
+                        m_Sensors.Add(sensor);
+                    }
+                    m_BoxOverlapChecker.RegisterSensor(sensor);
+                }
+            }
+            // Only one sensor needs to reference the boxOverlapChecker, so that it gets updated exactly once
+            ((GridSensorBase)m_Sensors[0]).m_BoxOverlapChecker = m_BoxOverlapChecker;
```
chriselion, line 273: m_Sensors will be empty if both m_UseOneHotTag and m_CountColliders are false, right? I think you need to check m_Sensors.Count first.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorComponent.cs
dongruoping: I can see how this make sense, but then when GetGridSensors is overridden the bool `m_UseOneHotTag` specified in UI will have no effect which is a bit confusing.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorComponent.cs
dongruoping: Fixed it. Also added tests for this.

com.unity.ml-agents.extensions/Editor/GridSensorComponentEditor.cs
dongruoping: updated.

com.unity.ml-agents.extensions/Runtime/Sensors/GridSensorComponent.cs
chriselion: If you override GetGridSensors(), you can still call base.GetGridSensors() (or not)

dongruoping: ok, I moved it.

surfnerd (approved):

com.unity.ml-agents.extensions/Runtime/Sensors/BoxOverlapChecker.cs
```diff
@@ -0,0 +1,254 @@
...
+                if (!ReferenceEquals(currentColliderGo, m_RootReference))
+                {
+                    detectedAction.Invoke(currentColliderGo, cellIndex);
+                }
+            }
+        }
+
+        internal void RegisterSensor(GridSensorBase sensor)
+        {
+            if (sensor.ProcessAllCollidersInCell())
```
chriselion, line 239: nit: instead of `bool ProcessAllCollidersInCell()` and two separate registration methods, could you make the GridSensorBase return an enum (ProcessAllColliders, ProcessClosestCollider, DebugColliders) ? I think that would simplify this a bit.

chriselion (approved):

com.unity.ml-agents.extensions/Runtime/Sensors/BoxOverlapChecker.cs
dongruoping: I made two separate registration methods because I don't want to expose `DebugColliders` since user should never use that.
I can make enum for `ProcessAllColliders` and `ProcessClosestCollider`.

PR #5240, opened 2021-04-09
better message for new model format

chriselion:
### Proposed change(s)
More explicit message if trying to load a new Barracuda model. Since we know the next major release will break the format, we should make this message clearer and more actionable.

image


chriselion:
The yamato failures are the same changelog validation ones discussed in slack.

ervteng (approved):

chriselion:
Merging since the test failures are unrelated.

PR #5262, opened 2021-04-14
fix all PVS and doc generation warnings

chriselion:
### Proposed change(s)
Fix all the package validation errors (at least for xmldoc) and the warnings from generating the docs.

### Types of change(s)
- [x] Documentation update

### DIFF
--- a/com.unity.ml-agents/Runtime/Actuators/ActionSegment.cs
+++ b/com.unity.ml-agents/Runtime/Actuators/ActionSegment.cs
@@ -116,19 +116,29 @@ namespace Unity.MLAgents.Actuators
             return Array == null || Array.Length == 0;
         }
 
-        /// <inheritdoc/>
+        /// <summary>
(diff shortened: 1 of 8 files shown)


com.unity.ml-agents/Runtime/Actuators/ActionSegment.cs
```diff
@@ -116,19 +116,29 @@ public bool IsEmpty()
             return Array == null || Array.Length == 0;
         }
 
-        /// <inheritdoc/>
+        /// <summary>
+        /// Returns an enumerator that iterates through the ActionSegment.
```
chriselion, line 120: These are basically copies of the .net docs, e.g. https://docs.microsoft.com/en-us/dotnet/api/system.collections.ienumerable.getenumerator?view=net-5.0

dongruoping (approved):

PR #5265, opened 2021-04-15
Rider suggested cleanup, part 1 

chriselion:
### Proposed change(s)
Low-hanging fruit, mostly unused `using ...`

### Types of change(s)
- [x] Bug fix

### DIFF
--- a/Project/Assets/ML-Agents/Examples/Sorter/Scripts/NumberTile.cs
+++ b/Project/Assets/ML-Agents/Examples/Sorter/Scripts/NumberTile.cs
@@ -1,5 +1,3 @@
-using System.Collections;
-using System.Collections.Generic;
 using UnityEngine;
 
 public class NumberTile : MonoBehaviour
@@ -8,7 +6,7 @@ public class NumberTile : MonoBehaviour
     public Material DefaultMaterial;
     public Material SuccessMaterial;
 
-    private bool m_Visited = false;
+    private bool m_Visited;
     private MeshRenderer m_Renderer;
 
     public bool IsVisited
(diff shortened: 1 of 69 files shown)


Project/Assets/ML-Agents/Examples/Sorter/Scripts/NumberTile.cs
```diff
@@ -8,7 +6,7 @@ public class NumberTile : MonoBehaviour
     public Material DefaultMaterial;
     public Material SuccessMaterial;
 
-    private bool m_Visited = false;
```
chriselion, line 11: Not sure how people feel about. My instinct is to a) be explicit (from python) and b) always initialize everything (coming from C++), but Rider likes to complain about it.

com.unity.ml-agents/Runtime/MLAgentsSettingsManager.cs
```diff
@@ -1,8 +1,9 @@
 using System;
-using System.Linq;
```
chriselion, line 2: This is only needed in play mode.

Project/Assets/ML-Agents/Examples/Sorter/Scripts/NumberTile.cs
dongruoping: Yeah I think making them explicit is not a bad thing and sometimes it feels safer (I know it works that way but I like seeing it written that way).
No strong opinions anyway.

dongruoping (approved):

Project/Assets/ML-Agents/Examples/Sorter/Scripts/NumberTile.cs
surfnerd: _traumatized by c++_

surfnerd (commented):
LGTM

PR #5276, opened 2021-04-16
check for missing AbstractBoard, display warning

chriselion:
### Proposed change(s)
If the AbstractBoard is missing, sensor creation will fail. I think it makes sense to just return empty arrays for the sensors and actuators in this case.

Also added a warning to the respective editor components.

I tried, but we can't add `[RequireComponent(typeof(AbstractBoard))]`, since AbstractBoard is, well, abstract.

### Types of change(s)
- [x] Bug fix

### Checklist

### DIFF
--- a/com.unity.ml-agents/Tests/Editor/Integrations/Match3/Match3ActuatorTests.cs
+++ b/com.unity.ml-agents/Tests/Editor/Integrations/Match3/Match3ActuatorTests.cs
@@ -194,5 +194,14 @@ namespace Unity.MLAgents.Tests.Integrations.Match3
             // And they should add up to all the potential moves
             Assert.AreEqual(validIndices.Count + masks.HashSets[0].Count, board.NumMoves());
         }
+
+        [Test]
+        public void TestNoBoardReturnsEmptyActuators()
+        {
+            var gameObj = new GameObject("board");
+            var actuatorComponent = gameObj.AddComponent<Match3ActuatorComponent>();
+            var actuators = actuatorComponent.CreateActuators();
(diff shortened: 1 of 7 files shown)


com.unity.ml-agents/Tests/Editor/Integrations/Match3/Match3ActuatorTests.cs
```diff
@@ -194,5 +194,14 @@ public void TestMasking(bool fullBoard)
...
             Assert.AreEqual(validIndices.Count + masks.HashSets[0].Count, board.NumMoves());
         }
+
+        [Test]
+        public void TestNoBoardReturnsEmptyActuators()
+        {
+            var gameObj = new GameObject("board");
+            var actuatorComponent = gameObj.AddComponent<Match3ActuatorComponent>();
+            var actuators = actuatorComponent.CreateActuators();
+            Assert.AreEqual(0, actuators.Length);
```
dongruoping, line 204: might be good to also check it doesn't throw exceptions

dongruoping (approved):

com.unity.ml-agents/Tests/Editor/Integrations/Match3/Match3ActuatorTests.cs
chriselion: If it did throw exceptions, the test would fail.

dongruoping: Oh ok then it's good

PR #5290, opened 2021-04-19
Run package tests with warn-as-error

chriselion:
### Proposed change(s)
We're running `upm-ci project test` in the coverage job, so I think we can go back to just running the package tests in this job (assuming we didn't need this for the way we pack). Add `--warnings-as-errors` here to catch warnings that would appear in the package.

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/.yamato/com.unity.ml-agents-test.yml
+++ b/.yamato/com.unity.ml-agents-test.yml
@@ -83,7 +83,7 @@ test_{{ package.name }}_{{ platform.name }}_{{ editor.version }}:
         flavor: {{ platform.flavor}}
     commands:
         - npm install upm-ci-utils@stable -g --registry https://artifactory.prd.cds.internal.unity3d.com/artifactory/api/npm/upm-npm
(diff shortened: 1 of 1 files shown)


.yamato/com.unity.ml-agents-coverage.yml
```diff
@@ -33,7 +33,7 @@ test_coverage_{{ package.name }}_{{ platform.name }}_{{ editor.version }}:
             {% if package.name == "com.unity.ml-agents.extensions" %}
             pull_request.changes.any match "com.unity.ml-agents.extensions/**" OR
             {% endif %}
-            pull_request.changes.any match ".yamato/com.unity.ml-agents-test.yml")
+            pull_request.changes.any match ".yamato/com.unity.ml-agents-coverage.yml")
```
chriselion, line 36: This was triggering even though it wasn't supposed to (now it's triggering because it was modified).

surfnerd: thanks

PR #5329, opened 2021-04-29
Test 2021.2

surfnerd:
Test against 2021.2


chriselion:
I think we need to change 
https://github.com/Unity-Technologies/ml-agents/blob/4e627782cee6d2e16703806650eed44c1ff7f69b/.yamato/test_versions.metafile#L10
too (OK if that's in another PR though)

chriselion (approved):

PR #5352, opened 2021-05-10
Light proofreading.

jo3w4rd:
One additional thing I noticed is that we have  a few references to "NumCellTypes" and "NumSpecialTypes," possibly belonging to a Board implementation, but these don't seem to be actual fields or properties defined anywhere.

### DIFF
--- a/com.unity.ml-agents/Runtime/Sensors/ISensor.cs
+++ b/com.unity.ml-agents/Runtime/Sensors/ISensor.cs
@@ -92,7 +92,7 @@ namespace Unity.MLAgents.Sensors
         /// Return information on the compression type being used. If no compression is used, return
         /// <see cref="CompressionSpec.Default()"/>.
         /// </summary>
-        /// <returns>The type of compression used by the sensor.</returns>
+        /// <returns>An object describing the compression used by the sensor.</returns>
         CompressionSpec GetCompressionSpec();
 
         /// <summary>


com.unity.ml-agents/Runtime/Integrations/Match3/AbstractBoard.cs
```diff
@@ -38,10 +38,10 @@ public struct BoardSize
         /// <param name="lhs"></param>
         /// <param name="rhs"></param>
         /// <returns>True if all fields are less than or equal.</returns>
-        public static bool operator <=(BoardSize lhs, BoardSize rhs)
+        public static bool operator<=(BoardSize lhs, BoardSize rhs)
         {
             return lhs.Rows <= rhs.Rows && lhs.Columns <= rhs.Columns && lhs.NumCellTypes <= rhs.NumCellTypes &&
-                   lhs.NumSpecialTypes <= rhs.NumSpecialTypes;
+                lhs.NumSpecialTypes <= rhs.NumSpecialTypes;
```
jo3w4rd, line 44: Looks like my IDE was being helpful.

chriselion: `dotnet-format` disagrees :) I re-ran it and pushed the changes back.

com.unity.ml-agents/Runtime/Sensors/ISensor.cs
```diff
@@ -92,7 +92,7 @@ public interface ISensor
         /// Return information on the compression type being used. If no compression is used, return
         /// <see cref="CompressionSpec.Default()"/>.
         /// </summary>
-        /// <returns>CompressionSpec used by the sensor.</returns>
+        /// <returns>The type of compression used by the sensor.</returns>
```
chriselion, line 95: Maybe 
```suggestion
        /// <returns>An object describing the compression used by the sensor.</returns>
```
(similar to `GetObservationSpec()`)?

PR #5370, opened 2021-05-14
[MLA-1981] Don't allow connections to newer python trainers

chriselion:
### Proposed change(s)
Version 0.26.0 (and higher) of the python trainers product files incompatible with the 1.0.x versions of the package. To prevent confusion and frustration, don't allow connections with these versions.

Note that the old logic for trainers that don't produce .nn files is still there with a warning; this is a separate hard error. I'm reluctant to make the old range a hard failure, since that would be a behavior change.

To summarize:

- 0.16.1 to 0.20.0 are supported
- 0.21.0 to 0.25.x are unsupported, but still compatible (warn, but allow training)
- 0.26.0 and higher are incompatible, no training possible

Also note that the old logic would warn on anything that wasn't in an X.Y.Z format (e.g. "1.2.3.dev4"). In this case, we try a little harder to extract the numerical part of the version (so that we can determine the version for clones of the github `main` branch), but don't block if we can't parse (to prevent a potential behavior change).

https://jira.unity3d.com/browse/MLA-1981

### Types of change(s)
- [x] Bug fix

### Checklist


Project/ProjectSettings/ProjectVersion.txt
```diff
@@ -1 +1 @@
-m_EditorVersion: 2018.4.17f1
```
chriselion, line 1: This (and the manifest) were pretty old. Just affects Project, but it's annoying to keep excluding them.

com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
```diff
@@ -99,6 +100,41 @@ public RpcCommunicator(CommunicatorInitParameters communicatorInitParameters)
         }
 
         internal static bool CheckPythonPackageVersionIsCompatible(string pythonLibraryVersion)
```
chriselion, line 102: The diff is messy; the old code was renamed to CheckPythonPackageVersionIs*Supported*.

com.unity.ml-agents/Runtime/Communicator/RpcCommunicator.cs
```diff
@@ -182,7 +218,21 @@ public UnityRLInitParameters Initialize(CommunicatorInitParameters initParameter
...
+                {
+                    Debug.LogErrorFormat(
+                        "Python package version ({0}) will produce model files that are incompatible with this " +
+                        "version of the com.unity.ml-agents Unity package. Please downgrade to a Python package " +
+                        "between {1} and {2}, or update to a new version of com.unity.ml-agents.",
+                        pythonPackageVersion,
+                        PythonTrainerVersions.s_MinSupportedVersion,
+                        PythonTrainerVersions.s_MaxSupportedVersion
+                    );
+                    throw new UnityAgentsException("Incompatible trainer version.");
```
chriselion, line 232: This gets caught and a different exception is thrown below.

com.unity.ml-agents/Runtime/Constants.cs
```diff
@@ -13,10 +13,14 @@ internal enum MenuGroup
...
     {
-        // The python package version must be >= s_MinSupportedVersion
+        // The python package version should be >= s_MinSupportedVersion
         // and <= s_MaxSupportedVersion.
         internal static Version s_MinSupportedVersion = new Version("0.16.1");
         internal static Version s_MaxSupportedVersion = new Version("0.20.0");
+
+        // Any version > to this is known to be incompatible and we will block training.
+        // Covers any patch to the release before the 2.0.0 package release.
+        internal static Version s_MaxCompatibleVersion = new Version("0.25.999");
```
chriselion, line 23: Just in case we do a 0.25.2 patch release.

surfnerd (approved):

PR #5390, opened 2021-05-25
pre-commit - fix depencency on pydoc-markdown

chriselion:
### Proposed change(s)
Change the pre-commit hook type to `python` so that it can specify it's pip dependency (so that you don't need to manually install it). I can potentially change the other scripts too, but none of them have external dependencies.

Also, the return code logic was potentially broken for multiple package directories or modules, as it would only use the most recent one.

See also: https://github.com/Unity-Technologies/ml-agents/pull/5325/

### Types of change(s)

- [x] CI

### DIFF
--- a/utils/generate_markdown_docs.py
+++ b/utils/generate_markdown_docs.py
@@ -69,8 +69,7 @@ if __name__ == "__main__":
     parser.add_argument("--package_dirs", nargs="+")
     args = parser.parse_args()
 
-    ok = False
-    return_code = 0
+    ok = True
     for package_dir in args.package_dirs:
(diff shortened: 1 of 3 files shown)


utils/generate_markdown_docs.py
```diff
@@ -96,8 +95,6 @@ def remove_trailing_whitespace(filename):
                     subprocess.check_call(subprocess_args, stdout=output_file)
                 remove_trailing_whitespace(output_file_name)
                 new_hash = hash_file(output_file_name)
-                ok = old_hash == new_hash
```
chriselion, line 99: This overwrote `ok` and `return_code` each time through the loop.

vincentpierre (approved):

PR #5411, opened 2021-06-02
[WIP] Custom raycast sensor

dongruoping:
### Proposed change(s)

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/com.unity.ml-agents/Runtime/Sensors/RayPerceptionSensor.cs
+++ b/com.unity.ml-agents/Runtime/Sensors/RayPerceptionSensor.cs
@@ -241,6 +241,7 @@ namespace Unity.MLAgents.Sensors
     public class RayPerceptionSensor : ISensor, IBuiltInSensor
     {
         float[] m_Observations;
+        float[] m_SingleRayObservations;
         ObservationSpec m_ObservationSpec;
         string m_Name;
 
@@ -267,10 +268,18 @@ namespace Unity.MLAgents.Sensors
             m_Name = name;
             m_RayPerceptionInput = rayInput;
 
-            SetNumObservations(rayInput.OutputSize());
+            SetNumObservations(GetObservationSizePerRay(), GetNumberOfRays());
 
             m_DebugLastFrameCount = Time.frameCount;
             m_RayPerceptionOutput = new RayPerceptionOutput();
+        }
+
+        /// <summary>
+        /// The ray perception input configurations.
+        /// </summary>
+        public RayPerceptionInput RayPerceptionInput
+        {
+            get { return m_RayPerceptionInput; }
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


com.unity.ml-agents/Runtime/Sensors/RayPerceptionSensor.cs
```diff
@@ -314,12 +356,14 @@ public int Write(ObservationWriter writer)
...
-                var numDetectableTags = m_RayPerceptionInput.DetectableTags.Count;
+                var rayObservartionSize = GetObservationSizePerRay();
 
                 // For each ray, write the information to the observation buffer
                 for (var rayIndex = 0; rayIndex < numRays; rayIndex++)
                 {
-                    m_RayPerceptionOutput.RayOutputs?[rayIndex].ToFloatArray(numDetectableTags, rayIndex, m_Observations);
+                    Array.Clear(m_SingleRayObservations, 0, rayObservartionSize);
+                    RayOutputToArray(m_RayPerceptionOutput.RayOutputs[rayIndex], rayIndex, m_SingleRayObservations);
+                    Array.Copy(m_SingleRayObservations, 0, m_Observations, rayObservartionSize * rayIndex, rayObservartionSize);
```
dongruoping, line 366: Changes here are debatable since it hurts performance. I introduced `m_SingleRayObservations` to avoid having to handle the buffer offset for each ray in `RayOutputToArray()` but it will create more array copy and clear operations. I couldn't find a way to create a reference to subarray to avoid copying.

One alternatives would be having `RayOutputToArray` to take all the `m_RayPerceptionOutput.RayOutputs` and the whole `m_Observations` array as input and let user handle the buffer offset themselves.

com.unity.ml-agents/Runtime/Sensors/RayPerceptionSensor.cs
```diff
@@ -279,27 +290,58 @@ public RayPerceptionOutput RayPerceptionOutput
...
+        public virtual int GetObservationSizePerRay()
+        {
+            return RayPerceptionInput.DetectableTags.Count + 2;
+        }
+
+        /// <summary>
+        /// The number of rays in the sensor.
+        /// Override this method for custom observations.
+        /// </summary>
+        public int GetNumberOfRays()
```
chriselion, line 306: Needs to be `virtual` if it's meant to be overridden.

com.unity.ml-agents/Runtime/Sensors/RayPerceptionSensor.cs
```diff
@@ -279,27 +290,58 @@ public RayPerceptionOutput RayPerceptionOutput
...
+        {
+            return RayPerceptionInput.Angles.Count;
+        }
+
+        void SetNumObservations(int observationsSize, int numRays)
```
chriselion, line 311: nit: `observationsSizePerRay`

com.unity.ml-agents/Runtime/Sensors/RayPerceptionSensor.cs
chriselion: I think instead of this, you can just call `writer.AddList` once per ray. That probably lets you get rid of m_observations entirely.

chriselion: * and pass an offset to `writer.AddList`

com.unity.ml-agents/Runtime/Sensors/RayPerceptionSensor.cs
```diff
@@ -279,27 +290,58 @@ public RayPerceptionOutput RayPerceptionOutput
...
         }
 
         internal void SetRayPerceptionInput(RayPerceptionInput rayInput)
         {
             // Note that change the number of rays or tags doesn't directly call this,
             // but changing them and then changing another field will.
-            if (m_RayPerceptionInput.OutputSize() != rayInput.OutputSize())
+            var oldObservationSize = GetObservationSizePerRay();
+            m_RayPerceptionInput = rayInput;
+            if (GetObservationSizePerRay() != oldObservationSize)
```
chriselion, line 324: This change misses the case where the observation size per ray stays the same, but the number of rays changes.

com.unity.ml-agents/Runtime/Sensors/RayPerceptionSensor.cs
dongruoping: yeah you're right, didn't notice that

com.unity.ml-agents/Runtime/Sensors/RayPerceptionSensor.cs
dongruoping: I don't think this one can be overridden since we handle the angle calculations internally. I added this as more like a helper method.

com.unity.ml-agents/Runtime/Sensors/RayPerceptionSensor.cs
dongruoping: fixed

com.unity.ml-agents/Runtime/Sensors/RayPerceptionSensor.cs
chriselion: Makes sense, but in that case, you need to remove "Override this method..." from the docstring :)

dongruoping: got copy-pasted. now removed

PR #5439, opened 2021-06-23
[DRAFT][DON'T MERGE] Enable learning from multiple priors

ervteng:
### Proposed change(s)

This PR adds a feature that was inspired by the approach used by DeepMind in [this paper](https://arxiv.org/abs/2105.12196). By specifying (i.e. giving the directory to that run id) multiple different policies as "priors" in the YAML file, the code will load those policies and use them as regularization priors for the learning policy. This has proven to be effective in Dodgeball, where training with two priors (shooting and flag-getting) leads to a skilled policy _in roughly 1/3 the time_. See ELO plot below. 

image

This PR also contains a version of WallJump that can be broken down into subtasks. 

### TODO
Ideally in order for this to be a feature, we'd want too add these components:

- Add to PPO and SAC (currently only in POCA)
- Allow checkpoints with different network architectures (e.g. different num_layers) to be handled properly (I don't see how we could change the obs space and action space, though). 
- Solve the entropy issue (entropy seems to increase when more than one prior is specified, probably b/c it's trying to learn a multimodal policy. 
- Documentation and tests

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/Project/Assets/ML-Agents/Examples/WallJump/Prefabs/WallJumpArea.prefab
+++ b/Project/Assets/ML-Agents/Examples/WallJump/Prefabs/WallJumpArea.prefab
@@ -43,9 +43,10 @@ Camera:
   m_ClearFlags: 2
   m_BackGroundColor: {r: 0.46666667, g: 0.5647059, b: 0.60784316, a: 1}
   m_projectionMatrixMode: 1
+  m_GateFitMode: 2
+  m_FOVAxisMode: 0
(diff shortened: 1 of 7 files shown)


montplaisir:
Since this PR is incomplete and outdated, it will be closed.
The idea is interesting, and would be worth investigating from the current codebase.

PR #5467, opened 2021-07-22
Fix the torch device option in argparser

dongruoping:
### Proposed change(s)

I fixed the attribute name part in #5433 but it broke the part when assigning the value from argparser to TorchSettings.

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


ervteng (approved):
I think we need to add this one to the changelog.md

PR #5512, opened 2021-08-26
Harden user PII protection logic and extend TrainingAnalytics to expose detailed configuration parameters.

sini:
### Proposed change(s)

- Replace the current privacy hashing method with a more cryptographically secure method.
- Extend the TrainingAnalytics trainer metrics to expose base configuration so hyperparameter usage can be evaluated.
- Provide functions for properly scrubbing configuration data to ensure that no potentially user identifying data is exposed.

Jira: https://jira.unity3d.com/browse/MLA-2179

### Types of change(s)

- [ ] Bug fix
- [X] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### Other comments
Feature functionality is covered by existing test coverage. I can add additional test coverage for the PII scrubbing private methods.


ml-agents/mlagents/training_analytics_side_channel.py
```diff
@@ -21,23 +25,92 @@ class TrainingAnalyticsSideChannel(SideChannel):
...
         raise UnityCommunicationException(
             "The TrainingAnalyticsSideChannel received a message from Unity, "
             + "this should not have happened."
         )
 
+    @staticmethod
+    def __sanitize_run_options(config: RunOptions) -> Dict[str, Any]:
+        res = copy.deepcopy(config.as_dict())
+
+        def hash(value: str) -> str:
```
miguelalonsojr, line 54: I understand why you choose to wrap TrainingAnalyticsSideChannel.__hash in an inner function, however, I would suggest avoiding inner functions unless it's for a decorator. Instead, since you need another static method from the class, I suggest making __sanitize_run_options a class method instead of a static method, and use cls.__hash instead of having to make the inner function.

miguelalonsojr: Here's the snippet:

`@classmethod
    def __sanitize_run_options(cls, config: RunOptions) -> Dict[str, Any]:
        res = copy.deepcopy(config.as_dict())

        # Filter potentially PII behavior names
        if "behaviors" in res and res["behaviors"]:
            res["behaviors"] = {cls.__hash(k): v for (k, v) in res["behaviors"].items()}
            for (k, v) in res["behaviors"].items():
                if "init_path" in v and v["init_path"] is not None:
                    hashed_path = cls.__hash(k)(v["init_path"])
                    res["behaviors"][k]["init_path"] = hashed_path

        # Filter potentially PII curriculum and behavior names from Checkpoint Settings
        if "environment_parameters" in res and res["environment_parameters"]:
            res["environment_parameters"] = {
                cls.__hash(k)(k): v for (k, v) in res["environment_parameters"].items()
            }
            for (curriculumName, curriculum) in res["environment_parameters"].items():
                updated_lessons = []
                for lesson in curriculum["curriculum"]:
                    new_lesson = copy.deepcopy(lesson)
                    if lesson.has_keys("name"):
                        new_lesson["name"] = cls.__hash(k)(lesson["name"])
                    if lesson.has_keys("completion_criteria"):
                        new_lesson["completion_criteria"]["behavior"] = cls.__hash(k)(
                            new_lesson["completion_criteria"]["behavior"]
                        )
                    updated_lessons.append(new_lesson)
                res["environment_parameters"][curriculumName][
                    "curriculum"
                ] = updated_lessons

        # Filter potentially PII filenames from Checkpoint Settings
        if "checkpoint_settings" in res and res["checkpoint_settings"] is not None:
            if (
                "initialize_from" in res["checkpoint_settings"]
                and res["checkpoint_settings"]["initialize_from"] is not None
            ):
                res["checkpoint_settings"]["initialize_from"] = hash(
                    res["checkpoint_settings"]["initialize_from"]
                )
            if (
                "results_dir" in res["checkpoint_settings"]
                and res["checkpoint_settings"]["results_dir"] is not None
            ):
                res["checkpoint_settings"]["results_dir"] = hash(
                    res["checkpoint_settings"]["results_dir"]
                )

        return res`

miguelalonsojr (changes requested):
Overall, looks good. Please address comments and also add unit tests on the C# and python side to cover these additions. The TrainingAnalyticsSideChannel class does not have any tests that I could find. I'll do a final integration test after you make the updates.

ml-agents/mlagents/training_analytics_side_channel.py
```diff
@@ -21,23 +25,92 @@ class TrainingAnalyticsSideChannel(SideChannel):
...
         # UUID('b664a4a9-d86f-5a5f-95cb-e8353a7e8356')
         super().__init__(uuid.UUID("b664a4a9-d86f-5a5f-95cb-e8353a7e8356"))
         self.run_options: Optional[RunOptions] = None
 
+    @staticmethod
+    def __hash(key: str, data: str) -> str:
+        res = hmac.new(
+            key.encode("utf-8"), data.encode("utf-8"), hashlib.sha256
+        ).hexdigest()
+        print(res)
```
miguelalonsojr, line 41: As per our convo, please remove the print() statement.

com.unity.ml-agents/Runtime/Analytics/AnalyticsUtils.cs
```diff
@@ -3,17 +3,41 @@
...
+        /// <returns>A string containing the key to be used for HMAC encoding.</returns>
+        /// <param name="value"></param>
+        /// <returns>A string containing the value to be encoded.</returns>
+        public static string Hash(string key, string value)
         {
-            var behaviorNameHash = Hash128.Compute(s);
-            return behaviorNameHash.ToString();
+            string hash;
+            System.Text.UTF8Encoding encoder = new System.Text.UTF8Encoding();
+            using (System.Security.Cryptography.HMACSHA256 hmac = new System.Security.Cryptography.HMACSHA256(encoder.GetBytes(key)))
```
mahon94, line 35: can we import it on top to give visibility that we're using this for hashing?

miguelalonsojr: +1

com.unity.ml-agents/Runtime/Analytics/InferenceAnalytics.cs
```diff
@@ -156,7 +156,7 @@ IList<IActuator> actuators
             var inferenceEvent = new InferenceEvent();
 
             // Hash the behavior name so that there's no concern about PII or "secret" data being leaked.
-            inferenceEvent.BehaviorName = AnalyticsUtils.Hash(behaviorName);
+            inferenceEvent.BehaviorName = AnalyticsUtils.Hash(k_VendorKey, behaviorName);
```
mahon94, line 159: There are other hashing in this file too ( eg. L222) should we change them to hmac as well?

sini: In this instance it looks like it's being used just for object identification. I don't see an argument that the model's layer data is itself PII and so the weaker hash is likely fine here.

ml-agents/mlagents/training_analytics_side_channel.py
```diff
@@ -21,23 +25,86 @@ class TrainingAnalyticsSideChannel(SideChannel):
...
+    __vendorKey: str = "unity.ml-agents"
+
     def __init__(self) -> None:
         # >>> uuid.uuid5(uuid.NAMESPACE_URL, "com.unity.ml-agents/TrainingAnalyticsSideChannel")
         # UUID('b664a4a9-d86f-5a5f-95cb-e8353a7e8356')
         super().__init__(uuid.UUID("b664a4a9-d86f-5a5f-95cb-e8353a7e8356"))
         self.run_options: Optional[RunOptions] = None
 
+    @classmethod
+    def __hash(cls, data: str) -> str:
```
dongruoping, line 37: Is there a specific reason that you used two underscores instead of one for these functions? I know it marks the function as private but in python it's still possible to access (if you really want to) and actually will prevent us to access it if we make a subclass and want to reuse these functions. So it doesn't seem necessary to me and I don't think we use this anywhere in the python codebase.

ml-agents/mlagents/training_analytics_side_channel.py
```diff
@@ -56,9 +124,18 @@ def environment_initialized(self, run_options: RunOptions) -> None:
         env_init_msg.set_raw_bytes(any_message.SerializeToString())
         super().queue_message_to_send(env_init_msg)
 
+    @classmethod
+    def __sanitize_trainer_settings(cls, config: TrainerSettings) -> Dict[str, Any]:
```
dongruoping, line 128: demo_path should be hashed too if using behavior cloning or GAIL

com.unity.ml-agents/Runtime/Analytics/TrainingAnalytics.cs
```diff
@@ -211,10 +211,14 @@ public static void TrainingBehaviorInitialized(TrainingBehaviorInitializedEvent
...
-            // Hash the behavior name so that there's no concern about PII or "secret" data being leaked.
             tbiEvent.TrainingSessionGuid = s_TrainingSessionGuid.ToString();
-            tbiEvent.BehaviorName = AnalyticsUtils.Hash(tbiEvent.BehaviorName);
 
+            if(tbiEvent.Config.Length == 0 || tbiEvent.BehaviorName.Length != 64) {
+              // Hash the behavior name if the message version is from an older version of ml-agents that doesn't do trainer-side hashing.
+              // We'll also, for extra safety, verify that the BehaviorName is the size of the expected SHA256 hash.
+              // Context: The config field was added at the same time as trainer side hashing, so messages including it should already be hashed.
+              tbiEvent.BehaviorName = AnalyticsUtils.Hash(k_VendorKey, tbiEvent.BehaviorName);
+            }
```
dongruoping, line 221: Besides adding python unit tests, it would be nice to also update the C# training analytics test to make sure it can properly handle this

dongruoping (commented):
We should consider to find a way to encrypt all the string data entered by user since it's very likely that we add more settings containing PII in the future without updating the analytics, and the unit tests won't be able to catch that.

.pre-commit-config.yaml
```diff
@@ -10,22 +10,25 @@ repos:
             )$
 
 -   repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v0.761
+    rev: v0.910
```
sini, line 13: Had compatibility issues with Python 3.9 -- this new version seems to be working much better with improved error handling and warnings.

com.unity.ml-agents/Runtime/Analytics/TrainingAnalytics.cs
miguelalonsojr: +1

ml-agents/mlagents/training_analytics_side_channel.py
miguelalonsojr: Although in this case, I agree it might be useful to remove name mangling from the hash function and allow derived classes to use the method, it also opens the possibility for collisions between the base and extended class's hash method, if a user chooses to override it. Since this is being used for to protect PII, we want to be able to guarantee that the hashing is being done correctly and to our specs. I'm leaning more towards keeping the mangling. If we want to allow extending this class, then we should design the class to abstract the calls to the hash function away.

miguelalonsojr (approved):

mahon94 (approved):
LGTM

PR #5528, opened 2021-09-02
Update dotnet-format to address breaking changes introduced by upstream changes

sini:
### Proposed change(s)

Pre-commit is using a version of dotnet and dotnet-format and has picked up a newer version of the utility that is not compatible with .NET 3.1 or 5.x -- it installs it's packages using `dotnet tool install`. This version's default entrypoint is also incompatible with our project structure, as the style and analyzer subcommands are now run by default and do not support the --folder option.

The entrypoint can be over-ridden to specify the whitespace module, which should match previous behavior.

Internal tracking issue: https://jira.unity3d.com/browse/MLA-2198
Other projects experience similar issues: https://github.com/icosa-gallery/open-brush/pull/167
dotnet-format issues tracking this:
dotnet/format#1317
dotnet/format#1318

### Types of change(s)

- [X] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -84,9 +84,10 @@ repos:
     exclude: ".*localized.*"
 
 - repo: https://github.com/dotnet/format
-  rev: "7e343070a0355c86f72bdee226b5e19ffcbac931"  # TODO - update to a tagged version when one that includes the hook is ready.
+  rev: v5.1.225507
   hooks:
   - id: dotnet-format
+    entry: dotnet-format whitespace
     args: [--folder, --include]
 
 # "Local" hooks, see https://pre-commit.com/#repository-local-hooks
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -84,9 +84,10 @@ repos:
     exclude: ".*localized.*"
 
 - repo: https://github.com/dotnet/format
-  rev: "7e343070a0355c86f72bdee226b5e19ffcbac931"  # TODO - update to a tagged version when one that includes the hook is ready.
+  rev: v5.1.225507
   hooks:
   - id: dotnet-format
+    entry: dotnet-format whitespace
```
dongruoping, line 90: can you add a line saying we added this because style and analyzer doesn't support --folder?

dongruoping (approved):

.pre-commit-config.yaml
sini: Discussed offline -- this may or may not be true in the future and context should be in the git history so this may be unnecessary to do.

mahon94 (approved):

PR #5538, opened 2021-09-16
Feature/mla 2205 separate schedule lr beta epsilon

miguelalonsojr:
### Proposed change(s)

Added support for separate schedules for learning rate, beta, and epsilon.

### Types of change(s)

- [ ] Bug fix
- [X] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist
- [X] Added tests that prove my fix is effective or that my feature works
- [X] Updated the [changelog](https://github.com/Unity-Technologies/ml-agents/blob/main/com.unity.ml-agents/CHANGELOG.md) (if applicable)
- [X] Updated the [documentation](https://github.com/Unity-Technologies/ml-agents/tree/main/docs) (if applicable)


dongruoping (commented):
1. Please add it to changelog
2. Optional: is there any example environment that can benefit from this? It would be great if we can actually train and test the effectiveness and put that as example for users

docs/Training-Configuration-File.md
```diff
@@ -39,6 +39,8 @@ choice of the trainer (which we review on subsequent sections).
...
+| `hyperparameters -> beta_schedule` | (default = `constant` for PPO) Determines how beta changes over time. <br><br>`linear` decays beta linearly, reaching 0 at max_steps, while `constant` keeps beta constant for the entire training run.                                                                                                           |
+| `hyperparameters -> epsilon_schedule` | (default = `linear`) Determines how epsilon changes over time (PPO only). <br><br>`linear` decays epsilon linearly, reaching 0 at max_steps, while `constant` keeps the epsilon constant for the entire training run.                                                                                                           |
```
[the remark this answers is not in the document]
miguelalonsojr: Yep, no problem.

ml-agents/mlagents/trainers/settings.py
```diff
@@ -151,6 +155,7 @@ class PPOSettings(HyperparamSettings):
     lambd: float = 0.95
     num_epoch: int = 3
     learning_rate_schedule: ScheduleType = ScheduleType.LINEAR
+    epsilon_schedule = ScheduleType = ScheduleType.LINEAR
```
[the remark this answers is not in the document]
miguelalonsojr: Great catch. Moved these to PPOSettings.

docs/Training-Configuration-File.md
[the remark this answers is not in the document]
miguelalonsojr: Also added the changes to the CHANGELOG.

docs/Training-Configuration-File.md
```diff
@@ -59,6 +59,8 @@ the `trainer` setting above).
...
+| `hyperparameters -> beta_schedule` | (default = `linear`) Determines how beta changes over time. <br><br>`linear` decays beta linearly, reaching 0 at max_steps, while `constant` keeps beta constant for the entire training run.                                                                                                           |
+| `hyperparameters -> epsilon_schedule` | (default = `linear`) Determines how epsilon changes over time (PPO only). <br><br>`linear` decays epsilon linearly, reaching 0 at max_steps, while `constant` keeps the epsilon constant for the entire training run.
```
dongruoping, line 63: After second thought, I think the default value should be the same as the learning_rate_schedule. These default values are changing the behavior of old config yaml and users will have to update **all** their yaml files. Although it has not been released as 1.0 yet, this could cause a huge confusion and troubles if people didn't notice this change.

I'd propose here we set the default value as whatever the learning_rate_schedule is, and revert all the changes to the config files. And create a ticket to remind ourselves to change this if we ever do trainer major release. If you strongly believe we should keep these default values we can discuss offline and in that case you should add it to the migration guide.

com.unity.ml-agents/CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ terminated teammates. (#5441)
 episode. (#5375)
 
 #### ml-agents / ml-agents-envs / gym-unity (Python)
+- Added support for having `beta`, `epsilon`, and `learning rate` on separate schedules (affects only PPO and POCA). (#5538)
```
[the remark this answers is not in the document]
miguelalonsojr: Weird. I though I added in the right place. Fixed.

docs/Training-Configuration-File.md
miguelalonsojr: That's what I did. I set the default to linear, which is what the default LR schedule. I also went in and added it to the config files to match whatever the LR schedule was set to, to make sure that there was no change to training. Not sure what else to do.

dongruoping: oh I saw you set something to constant originally.

But if a user has a config file that specified `learning_rate_schedule: constant`, they will get different behavior using the old trainer and new trainer right?

andrewcoh: I'm not sure there's a great way to use the value of one config for others if not specified using attrs i.e. default to lr schedule if beta/epsilon are not scheduled. It's possible that this can be done in a validator/structure method.

Another benefit of this would be that we don't need to add it to all of the configs.

ml-agents/mlagents/trainers/settings.py
```diff
@@ -44,6 +44,19 @@ def check_and_structure(key: str, value: Any, class_type: type) -> Any:
     return cattr.structure(value, attr_fields_dict[key].type)
 
 
+def check_hyperparam_schedules(val: Dict, trainer_type: type) -> Dict:
```
mahon94, line 47:
```suggestion
def check_hyperparam_schedules(val: Dict, trainer_type: TrainerType) -> Dict:
```

docs/Training-Configuration-File.md
mahon94, line 63: It worth noting the default is `hyperparameters -> learning_rate_schedule` here

ml-agents/mlagents/trainers/tests/test_config_conversion.py
```diff
@@ -27,6 +27,8 @@
         lambd: 0.95
         learning_rate: 3.0e-4
         learning_rate_schedule: linear
+        beta_schedule: constant
+        epsilon_schedule: linear
```
mahon94, line 31: Is it possible to add some assertion at here to check the default replacement of beta_schedule and epsilon_schedule?

mahon94 (approved):

PR #5597, opened 2021-10-27
Return deterministic actions 

cmard:
### Proposed change(s)

This PR will add an ability to retrieve actions deterministically based on the input to the model. A new run-options configuration has been added as well as a new CLI flag `--deterministic`.  

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


ml-agents/mlagents/trainers/torch/distributions.py
```diff
@@ -113,6 +123,9 @@ def __init__(self, logits):
     def sample(self):
         return torch.multinomial(self.probs, 1)
 
+    def deterministic_sample(self):
+        return torch.argmax(self.probs).reshape((1, 1))
```
mahon94, line 127:
```suggestion
        return torch.argmax(self.probs, dim=1, keepdim=True)
```

PR #5599, opened 2021-10-29
Package changes to support deterministic inference

mahon94:
### Proposed change(s)

New models in ml-agents are exported with 2 new action tensors to support deterministic action selection during inference with barracuda. This option is only available in editor mode (with no python training attached) using `Stochastic Inference` flag. 
Screen Shot 2021-11-03 at 11 51 26 AM

Screen Shot 2021-11-04 at 8 23 42 AM

MLA-2263

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


miguelalonsojr (changes requested):
High level review before digging into the code:

1. You mentioned this is an editor only option. Can we make this a runtime option as well? My thinking here is that folks may want to package up the policy and set it to deterministic mode and ship it that way.
2. Let's be consistent between the python and Unity side. The default is assumed to be Stochastic, so you must add a "deterministic" flag to either the CLI or config. Let's do the same in Behavior Parameters, i.e. the checkbox should be labeled "Deterministic Policy" and have it unchecked. by default.

mahon94:
1.  The flag only works within unity editor or unity standalone after build. If it's used from python `mlagents-learn`, `--deterministic` should be used.
2.  2.Done, reverted the flag to deterministic inference.

Thanks

com.unity.ml-agents/Tests/Editor/Inference/ModelRunnerTest.cs
```diff
@@ -138,5 +177,46 @@ public void TestRunModel()
...
+            info1.episodeId = 1;
+            modelRunner.PutObservations(info1, obs);
+            modelRunner.DecideBatch();
+            var stochAction1 = (float[])modelRunner.GetAction(1).ContinuousActions.Array.Clone();
+
+            modelRunner.PutObservations(info1, obs);
+            modelRunner.DecideBatch();
+            var stochAction2 = (float[])modelRunner.GetAction(1).ContinuousActions.Array.Clone();
+            // Stochastic action selection should output randomly different action values with same obs
+            Assert.IsFalse(Enumerable.SequenceEqual(stochAction1, stochAction2, new FloatThresholdComparer(0.001f)));
```
miguelalonsojr, line 204: I think it would be a good idea to refactor this test into two tests, one for deterministic actions and one for stochastic actions.

miguelalonsojr (changes requested):

cmard (approved):

miguelalonsojr (approved):
LGTM.

PR #5619, opened 2021-11-18
Deterministic actions python training

mahon94:
### Proposed change(s)

- [Inference] support deterministic action selection during inference with barracuda. This option is only available in editor mode (with no python training attached) using `Stochastic Inference` flag. (#5599)
- [Inference] Serialize two output tensors to onnx for deterministic actions (#5593):
   -  "deterministic\_continuous\_actions"
   -   "deterministic\_discrete\_actions"

- [Training] Add an ability to retrieve actions deterministically based on the input to the model. A new run-options configuration has been added as well as a new CLI flag `--deterministic`.  (#5597)

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [x] Documentation update
- [ ] Other (please describe)

### Checklist
- [x] Added tests that prove my fix is effective or that my feature works
- [x] Updated the [changelog](https://github.com/Unity-Technologies/ml-agents/blob/main/com.unity.ml-agents/CHANGELOG.md) (if applicable)
- [x] Updated the [documentation](https://github.com/Unity-Technologies/ml-agents/tree/main/docs) (if applicable)
- [ ] Updated the [migration guide](https://github.com/Unity-Technologies/ml-agents/blob/main/docs/Migrating.md) (if applicable)

PR #5622, opened 2021-11-18
Revert "Deterministic actions python training"

mahon94:
Reverts #5619 Due to extra unrelated prints


miguelalonsojr (approved):
Stamping.

PR #5628, opened 2021-12-02
Fix analytics error when compiling XBOX and PS5

hvpeteet:
### Proposed change(s)

Bugfix: Turns off analytics for platforms that don't support them. To use analytics safely you must wrap usage with `#if ENABLE_CLOUD_SERVICES_ANALYTICS
` 
See https://docs.unity3d.com/ScriptReference/Analytics.Analytics.html

Internal JIRA: https://jira.unity3d.com/browse/MLA-2151

### Types of change(s)

- [x] Bug fix

### Testing

Errors went away when compiling for XBox in windows VM

ALSO: 2020.3 is failing because of an existing issue https://jira.unity3d.com/browse/MLA-2311

### DIFF
--- a/com.unity.ml-agents/CHANGELOG.md
+++ b/com.unity.ml-agents/CHANGELOG.md
@@ -39,6 +39,7 @@ and this project adheres to
 ### Bug Fixes
 - Fixed a bug where the critics were not being normalized during training. (#5595)
 - Fixed the bug where curriculum learning would crash because of the incorrect run_options parsing. (#5586)
+- Fixed a bug where ml-agents code wouldn't compile on platforms that didn't support analytics (PS4/5, XBoxOne) (#5628)
 
 #### com.unity.ml-agents / com.unity.ml-agents.extensions (C#)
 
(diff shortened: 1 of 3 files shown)


maryamhonari (approved):
LGTM. Would you add a line to the change log? thanks!

hvpeteet:
> LGTM. Would you add a line to the change log? thanks!

Done, thanks for the catch.

jrupert-unity (approved):
LGTM in terms of addressing the issue.  I'm wondering if we need all three of UNITY_EDITOR , MLA_UNITY_ANALYTICS_MODULE and ENABLE_CLOUD_SERVICES_ANALYTICS in these checks  and if they could be made more consistent.  This is something that could be looked at separately if there's a question about it.

hvpeteet:
> I'm wondering if we need all three of UNITY_EDITOR , MLA_UNITY_ANALYTICS_MODULE and ENABLE_CLOUD_SERVICES_ANALYTICS in these checks and if they could be made more consistent. 

👍   was mainly just trying to get a fix in before the release. I feel like this could be pruned down and we could commit to just using [conditional attributes ](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.conditionalattribute) (already used in this code alongside the preprocessor directives), but I wasn't sure of all the edge cases with analytics so I played it safe and just added onto the existing directives.

PR #5770, opened 2022-07-27
Include extensions in Doxygen docs

TV4Fun:
Classes in the `com.unity.ml-agents.extensions` package might get more
use if people knew they existed. Add them to docs here.


montplaisir:
Thanks for improving our docs! 
`com.unity.ml-agents.extensions` package was merged into `com.unity.ml-agents` in PR #6227.

PR #5922, opened 2023-05-16
Stop pinning numpy to a specific version

hyerra:
### Proposed change(s)
I think this might have been a typo as numpy was specified twice. However, I removed the second line that pins numpy to a specific version. Requiring this exact version of numpy breaks compatibility with pytorch versions > 2.0.0.

N/A

### Types of change(s)
Bug Fix

### Checklist
- [x] Updated the documentation (if applicable)
- [x] Updated the migration guide (if applicable)

### Other comments
N/A

### DIFF
--- a/ml-agents-envs/setup.py
+++ b/ml-agents-envs/setup.py
@@ -57,7 +57,6 @@ setup(
         "pyyaml>=3.1.0",
         "gym>=0.21.0",
         "pettingzoo==1.15.0",
-        "numpy==1.21.2",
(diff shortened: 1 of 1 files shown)


hyerra:
(cc: @miguelalonsojr) I was hoping you could take a look; it’s a really small fix that’s pretty important.

miguelalonsojr:
We'll be working on an update to pytorch 2.x separately. For now, it should remain pinned. Closing.

PR #5939, opened 2023-06-20
Added DecisionStep parameter to the decision requester

mitsunami:
### Proposed change(s)

I propose to introduce a new parameter, named DecisionStep, into the DecisionRequester module. This addition will allow the staggering of execution timing for each agent. This would be useful for developers when using multi-agents, leading to more stable performance. Otherwise, you would experience significant processing time spikes every DecisionPeriod step.

The default value of DecisionStep is set to zero, which means it will behave exactly the same as the default DecisionRequester when not adjusted.

This is a proposal from our experience developing a multi-agent game, we believe it would be beneficial to allow manual setting of DecisionStep for each agent rather than automatic/random adjustment because it allows to specify which agents to run in batch as efficiently as possible.

This feature is a generalized implementation of the one used in the demo we presented and showcased at GDC2023. An overview of the presentation and demo can be found at the following link:
- https://schedule.gdconf.com/session/how-to-implement-multi-agent-machine-learning-scenarios-in-mobile-gaming-presented-by-arm/894142
- https://community.arm.com/arm-community-blogs/b/ai-and-ml-blog/posts/p1-multi-agent-reinforcement-learning

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/com.unity.ml-agents/CHANGELOG.md
+++ b/com.unity.ml-agents/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to
 
 ### Minor Changes
 #### com.unity.ml-agents / com.unity.ml-agents.extensions (C#)
+- Added DecisionStep parameter to DecisionRequester (#)
+  - This will allow the staggering of execution timing when using multi-agents, leading to more stable performance.
 
 #### ml-agents / ml-agents-envs
 - Added training config feature to evenly distribute checkpoints throughout training. (#5842)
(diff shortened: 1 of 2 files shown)


miguelalonsojr (approved):
LGTM.

PR #6062, opened 2024-03-04
Upgrade ONNX to 1.15.0

alex-mccarthy-unity:
### Proposed change(s)

This fixes builds of ONNX on OS X while installing ml-agents.

OS X builds use Xcode by default, and the Xcode compiler defaults to using C++98 mode for C++ ( https://stackoverflow.com/a/21349148 ). This causes errors building protocol buffer libraries, which need to be compiled with support for C++14 or newer ( https://github.com/protocolbuffers/protobuf/issues/12393#issuecomment-1508983035 ).

[This ONNX commit](https://github.com/onnx/onnx/commit/a979e756133167e0bdf5af23441cca4d189e805c) changes its compilation to use C++14 mode: releases that include this commit (1.15.0 or newer) build with Xcode by default.

ONNX 1.15.0 uses a newer protocol buffer library, so allow newer versions here too.
### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Other (please describe) dependency upgrade

### Checklist

### DIFF
--- a/com.unity.ml-agents/CHANGELOG.md
+++ b/com.unity.ml-agents/CHANGELOG.md
@@ -27,6 +27,8 @@ and this project adheres to
 
 #### ml-agents / ml-agents-envs
 - Bumped numpy version to >=1.21.2,<1.24.0 (#5997)
+- Bumped onnx version to 1.15.0 (#6062)
+- Bumped protobuf version to >=3.6,<21 (#6062)
 
 ## [3.0.0-exp.1] - 2023-10-09
 ### Major Changes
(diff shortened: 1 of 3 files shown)


Hunter-Unity (commented):
Confirmed install works on M1 mac in combo with "conda install grpcio" PR. installed ML-A and was able to train and resume training just fine

miguelalonsojr (approved):
LGTM.

PR #6094, opened 2024-03-31
add support for mac m1/m2 chips mps gpu

zyunlongz:
need upgrade onnx and protobuf, add setup default device for torch, so user can use command params "--torch-device mps:0" to launch gpu training on mac

### Proposed change(s)

add support for mac m1/m2 chips mps gpu
user can use command params
"--torch-device mps:0" to launch gpu training on mac


montplaisir:
Your changes were covered by PR #6303.

PR #6097, opened 2024-04-18
Upgraded to Sentis v1.4.0-pre.3

miguelalonsojr:
### Proposed change(s)

Sentis v1.4 upgrade.

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [X] Other (please describe)

Dependency update.

### Checklist
- [X] Added tests that prove my fix is effective or that my feature works
- [X] Updated the changelog (if applicable)
- [X] Updated the documentation (if applicable)


com.unity.ml-agents/package.json
```diff
@@ -5,7 +5,7 @@
   "unity": "2023.2",
```
mplantady, line 5: Do you need to bump the version? And update the changelog?

com.unity.ml-agents/Runtime/Inference/ApplierImpl.cs
```diff
@@ -165,10 +170,6 @@ internal void ComputeCdf(TensorProxy logProbs, int batch, int channelOffset, int
         {
             // Find the class maximum
             var maxProb = float.NegativeInfinity;
-            if (logProbs.Device == DeviceType.GPU)
-            {
-                logProbs.data.MakeReadable();
```
mplantady, line 170: Should you add a CompleteOperationsAndDownload() in this method?

com.unity.ml-agents/Runtime/Inference/TensorProxy.cs
```diff
@@ -69,13 +69,21 @@ public long Channels
...
         {
-            data?.Dispose();
+            Dispose();
+        }
+
+        void Dispose()
+        {
+            if (data.dataOnBackend.backendType != BackendType.CPU)
+            {
+                data?.Dispose();
```
mplantady, line 79: Why is it not needed to dispose when using CPU?

com.unity.ml-agents/Runtime/Inference/ModelRunner.cs
```diff
@@ -76,9 +74,10 @@ internal class ModelRunner
                 // D.logEnabled = m_Verbose;
 
                 sentisModel = ModelLoader.Load(model);
+                sentisModelInfo = new SentisModelInfo(sentisModel, deterministicInference);
```
mplantady, line 77: you could do
`using var new SentisModelInfo(`
To avoid having to call the dispose manually at the end (and which may not be called if an exception is thrown)

mplantady (approved):
That looks good to me!

com.unity.ml-agents/package.json
miguelalonsojr: Changelog yes, version should be ok I think. Did 2023.3 come out already?

com.unity.ml-agents/Runtime/Inference/ApplierImpl.cs
miguelalonsojr: Yes. Good catch. This is for legacy support, but yeah, should be CompleteOperationsAndDownload for non CPU.

com.unity.ml-agents/Runtime/Inference/TensorProxy.cs
miguelalonsojr: There's no need to copy to the CPU mem. It's already there and accessible. This really only applies, at least for us, when using GPU backends.

com.unity.ml-agents/Runtime/Inference/ModelRunner.cs
miguelalonsojr: ah. good point. will change.

Issue #6008, opened 2023-11-10
ERROR: Failed building wheel for numpy when installing ml-agents==1.0.0


ohernpaul:
In the guide and documentation, it states that 1.0.0 Requires-Python >=3.10.1,<=3.10.12. Given any Anaconda environment between this range, I am failing to build wheel for numpy. The version that ml-agents 1.0 is trying to use is numpy 1.13.3 and fails with this error. 

>   × Building wheel for numpy (pyproject.toml) did not run successfully.
>   │ exit code: 1
>   ╰─> [287 lines of output]
>       setup.py:63: RuntimeWarning: NumPy 1.21.2 may not yet support Python 3.10.

The steps I am taking are:

1. Create a new conda environment within the given python version range defined above.
2. Navigate to the directory containing the ml-agents and ml-agents-envs folders.
3. Calling pip install ./ml-agents
4. Failure message

I have even tried installing numpy 1.21.2 which is the version for python 3.10 and it still fails.

To make sure my Anaconda setup is correct, I went through a re-install of my previous environment for ML-Agents==0.28.0 and that worked perfectly fine the first time using python 3.7.16.

```
(mlagents_python_3_10_12) D:\MProjects\AIsland\ml-agents-master\ml-agents>pip install ./
Processing d:\mprojects\aisland\ml-agents-master\ml-agents
  Preparing metadata (setup.py) ... done
Collecting grpcio<=1.48.2,>=1.11.0 (from mlagents==1.0.0)
  Using cached grpcio-1.48.2-cp310-cp310-win_amd64.whl (3.6 MB)
Collecting h5py>=2.9.0 (from mlagents==1.0.0)
  Using cached h5py-3.10.0-cp310-cp310-win_amd64.whl.metadata (2.5 kB)
Collecting mlagents_envs==1.0.0 (from mlagents==1.0.0)
  Using cached mlagents_envs-1.0.0-py3-none-any.whl.metadata (2.4 kB)
Requirement already satisfied: numpy<2.0,>=1.13.3 in c:\programdata\anaconda3\envs\mlagents_python_3_10_12\lib\site-packages (from mlagents==1.0.0) (1.22.1)
Collecting Pillow>=4.2.1 (from mlagents==1.0.0)
  Using cached Pillow-10.1.0-cp310-cp310-win_amd64.whl.metadata (9.6 kB)
Collecting protobuf<3.20,>=3.6 (from mlagents==1.0.0)
[...]

PR #6111, opened 2024-05-08
Update Installation.md  because the branch-21 release numpy setup fil…

NicolasKuske:
…e is not compatible with python 3.10.12

Update Installation.md  because the branch-21 release numpy setup file is not compatible with python 3.10.12

### Proposed change(s)

I merged the sections on git copy branch-21 release and development repository

https://github.com/Unity-Technologies/ml-agents/issues/6008

### Types of change(s)

- [x] Bug fix

### Other comments

Tried it myself, this is the only way your install doc does not lead to automatic error

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -72,20 +72,15 @@ of our tutorials / guides assume you have access to our example environments).
 the repository if you would like to explore more examples.
 
 ```sh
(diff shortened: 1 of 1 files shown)


miguelalonsojr:
Thanks for the contribution. Will update the Installation doc to note that release branches may become stale and to use bleeding edge `develop` branch in those instances.

PR #6136, opened 2024-08-18
Update unity_pettingzoo_base_env.py

AmineAndam04:
### Proposed change(s)

in the the ```_update_action_spaces``` method of the UnityPettingzooBaseEnv class, you force the actions to be of int32 and in the [-1,1] range (see the code below). This constraint doesn't apply to all scenarios

```python
act_spec.continuous_size > 0:
                    c_space = spaces.Box(
                        -1, 1, (act_spec.continuous_size,), dtype=np.int32
                    )
```

It makes more sense to use the same logic as in ```_update_observation_spaces```.

```python
obs_spaces = tuple(
                    spaces.Box(
                        low=-np.float32(np.inf),
                        high=np.float32(np.inf),
                        shape=spec.shape,
                        dtype=np.float32,
                    )
```
We can generalize to all scenarios by using:
```
if act_spec.continuous_size > 0:
                    c_space = spaces.Box(
                        -np.float32(np.inf), np.float32(np.inf), (act_spec.continuous_size,), dtype=np.float32
                    )
```

### Types of change(s)

- [x ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/ml-agents-envs/mlagents_envs/envs/unity_pettingzoo_base_env.py
+++ b/ml-agents-envs/mlagents_envs/envs/unity_pettingzoo_base_env.py
@@ -168,7 +168,7 @@ class UnityPettingzooBaseEnv:
             if action.continuous is not None:
                 self._current_action[current_behavior].continuous[
                     current_index
-                ] = action.continuous[0]
+                ] = action.continuous
             if action.discrete is not None:
                 self._current_action[current_behavior].discrete[
                     current_index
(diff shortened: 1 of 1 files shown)


alexander-zap:
+1, ran into the same problem. This is especially problematic when only using the mlagents-envs package for environment interfacing, but using own training algorithms (e.g., StableBaselines3).

montplaisir:
Your changes were covered by PR #6309. You are credited as co-author.
FYI We kept the continuous action bounds at [-1, 1] rather than [-inf, inf] as per the ML-Agents convention.

Issue #5890, opened 2023-03-29
Feature request: upgrade to gymnasium/update pettingzoo version


elliottower:
**Is your feature request related to a problem? Please describe.**

Hi, would it be possible add compatibility with [Gymnasium](https://github.com/Farama-Foundation/Gymnasium) and newer versions of [PettingZoo](https://github.com/Farama-Foundation/PettingZoo)? 

Gymnasium and PettingZoo are compatible with current RL training libraries ([rllib](https://github.com/ray-project/ray/tree/master/rllib), [tianshou](https://github.com/thu-ml/tianshou) and [CleanRL](https://github.com/vwxyzjn/cleanrl) have already migrated, and [stable-baselines3](https://github.com/DLR-RM/stable-baselines3/blob/e5deeed16efb57c34ccdcb14692439154d970527/docs/guide/install.rst#bleeding-edge-version) will soon) as well as other tools such as Comet and WandB.

For information about upgrading and compatibility, see [migration guide](https://gymnasium.farama.org/content/migration-guide/) and [gym compatibility](https://gymnasium.farama.org/content/gym_compatibility/). The main difference is the API has switched to returning `truncated` and `terminated`, rather than `done`, in order to give more information and mitigate edge case issues. 

We would like to list this package in the third party environments lists ([PettingZoo](https://pettingzoo.farama.org/environments/third_party_envs/), [Gymnasium](https://gymnasium.farama.org/environments/third_party_environments/)), but are removing all libraries which do not plan to migrate to Gymnasium/recent versions of PettingZoo.

**Describe the solution you'd like**

This package would be updated to use gymnasium rather than gym, and use at least PettingZoo v1.20 (uses termination/truncation and gymnnasium rather than gym).

**Describe alternatives you've considered**

**Additional context**

elliottower:
Hi, just bumping this as I would love to be able to use gymnasium/pettingzoo with unity ml-agents. I would be happy to contribute any help which is needed, just wanted to see if this is on the radar and how difficult it would be. 

We are cutting a new pettingzoo release shortly with some important fixes (no major breaking changes) and I was thinking after that is out it would be great to try and update this project to be compatible. Gymnasium had a new release V28 recently and I would be happy to contribute to make that compatible as well.

miguelalonsojr:
Thanks for the request. We're planning these updates for our next release.

elliottower:
> Thanks for the request. We're planning these updates for our next release.

[...]

PR #6211, opened 2025-06-13
Pettingzoo wrapper changed to gymnasium interface

alexander-zap:
### Proposed change(s)

I adjusted the Unity Pettingzoo Wrapper to support the gymnasium interface (gymnasium>=0.25).

This includes:

- returning `terminated, truncated` on step
- returning `info` on reset
- accept `seed` and `options` on reset
- not accept `mode` on render

Additionall, I fixed interface problems with pettingzoo implementation:

- all agents return an observations synchronously (instead of having steps where only one agent returns an observation)
- in case the agent sends a TerminationStep *and* a DecisionStep after, the reward from the termination step should be taken (but for observations it's the other way around because of SB3 convention)
- `agents` attribute should store agents even if they have *just* finished (they should be removed the step after)
- fixed a bug where a continuous action vector was assigned len(action) * action[0]
- fixed double removal of an agent from _live_agents for ParallelEnv

Also I bumped the dependencies for pettingzoo to 1.22.0 (because this is there Gymnasium support started on their side).

This also could be helpful for: https://github.com/Unity-Technologies/ml-agents/issues/5890

### Types of change(s)

- [x] Bug fix
- [x] New feature
- [x] Code refactor
- [x] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist
- [x] Updated the documentation (if applicable)


alexander-zap:
Should also solve #6214

montplaisir:
Your changes were covered by PR #6309 - Migrate gym to gymnasium.
You are credited as co-author.

PR #6221, opened 2025-07-16
Update Analytics to use new API

maryamziaa:
### Proposed change(s)

The old analytics API has been deprecated. This PR updates the analytics implementation to use the new API.

### Types of change(s)

- [ ] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [x] Other (please describe): Upgraded to new Analytics API

### Checklist

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -12,6 +12,8 @@ on:
 jobs:
   pre-commit:
     runs-on: ubuntu-24.04
+    env:
+        DOTNET_NOLOGO: 1
     steps:
     - uses: actions/checkout@v4
     - uses: actions/setup-python@v5
@@ -26,6 +28,8 @@ jobs:
     - uses: actions/setup-dotnet@v4
       with:
         dotnet-version: '8.0.202'
+    -   name: Clean dotnet shared memory
+        run: sudo rm -rf /tmp/.dotnet/shm
     - name: Install manual dependencies
       run: |
         python -m pip install pre-commit
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -25,7 +25,7 @@ jobs:
         bundler-cache: true # runs 'bundle install' and caches installed gems automatically
     - uses: actions/setup-dotnet@v4
       with:
-        dotnet-version: '9.0.7'
+        dotnet-version: '9.0.301'
     - name: Install manual dependencies
       run: |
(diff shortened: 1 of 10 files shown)


.github/workflows/pre-commit.yml
```diff
@@ -26,6 +28,8 @@ jobs:
     - uses: actions/setup-dotnet@v4
       with:
         dotnet-version: '8.0.202'
+    -   name: Clean dotnet shared memory
+        run: sudo rm -rf /tmp/.dotnet/shm
```
maryamziaa, line 32: any objections to do a cleanup first? Anything else is using /tmp/.dotnet/shm?

com.unity.ml-agents/Runtime/Analytics/TrainingAnalytics.cs
```diff
@@ -90,7 +51,7 @@ internal static bool EnableAnalytics()
                 s_TrainingSessionGuid = Guid.NewGuid();
             }
 
-            return s_EventsRegistered;
+            return true;
```
maryamziaa, line 54: EnableAnalytics() always returns true if the conditions apply, i.e. UNITY_EDITOR && MLA_UNITY_ANALYTICS_MODULE && ENABLE_CLOUD_SERVICES_ANALYTICS. Is that OK? We don't RegisterEvent anymore in the new API.

com.unity.ml-agents/Runtime/Analytics/InferenceAnalytics.cs
```diff
@@ -54,25 +36,17 @@ internal class InferenceAnalytics
...
-            {
-                s_EventRegistered = true;
-            }
-            if (s_EventRegistered && s_SentModels == null)
+            if (s_SentModels == null)
             {
                 s_SentModels = new HashSet<ModelAsset>();
             }
 
+            return true;
```
maryamziaa, line 45: EnableAnalytics() always returns true if the conditions apply, i.e. UNITY_EDITOR && MLA_UNITY_ANALYTICS_MODULE && ENABLE_CLOUD_SERVICES_ANALYTICS. Is that OK? We don't RegisterEvent anymore in the new API.

antoine-agthe-unity (approved):
I invite you to give more explanation about why `IAnalytics` and `IData` types are implemented that way as it might look exotic, who an external point of view.

alexander-suvorov (approved):

PR #6233, opened 2025-08-03
A new training mode called "both" has been added. This allows simultaneous training with both PPO and SAC, each on a separate instance. For example, if you have 4 environments, the even-numbered ones will use PPO, and the odd-numbered ones will use SAC. Therefore, it's necessary to specify even values for num_envs.

VictorBarbosa:
### Proposed change(s)

### Types of change(s)

- [ ] Bug fix
- [x] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist


VictorBarbosa:
We often find ourselves unsure whether to use PPO or SAC, and sometimes we only realize we should’ve chosen the other one after spending a long time training. So why not train both at the same time?

I’ve added a feature that allows simultaneous training with both PPO and SAC. Each will run on its own separate instance. For this to work properly, it's important to set num_envs to a value divisible by 2.

Below, you can see an example of how the YAML configuration looks for this training setup.
 
image

PR #6243, opened 2025-09-02
Fix: data on both cuda:0 and cuda:1 when using --torch-device cuda:1

robinDLFM:
### Proposed change(s)
Solves:
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:1 and cuda:0! (when checking argument for argument mat1 in method wrapper_CUDA_addmm)
when using --torch-device cuda:1

- [x] Bug fix
- [x] Code refactor


montplaisir:
Your changes were covered by PR #6303.

PR #6274, opened 2026-01-08
Crash on quit runmode

Codengineer:
Dispose was not called in Academy when editor was quitting resulting in a crash in the Job system when CPU Workers were disposed in finalizers called from background threads (they need to be called on the Main thread)

By calling Dispose before tearing down the job system, CPU Workers are disposed properly the same way they are when you leave play mode.

### Proposed change(s)

Added a Dispose action in Academy when Editor Application is quitting

### Types of change(s)

- [x] Bug fix
- [ ] New feature
- [ ] Code refactor
- [ ] Breaking change
- [ ] Documentation update
- [ ] Other (please describe)

### Checklist

### DIFF
--- a/com.unity.ml-agents/CHANGELOG.md
+++ b/com.unity.ml-agents/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to
 - Upgraded to Inference Engine 2.4.1 (#6269)
 - Fixed tensor indexing to use correct CHW layout (#6239)
 - Updated the installation doc (#6242)
+- Fixed Unity Editor crashing when quitting in play mode (#6267)
 
 #### ml-agents / ml-agents-envs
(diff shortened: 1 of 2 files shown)


com.unity.ml-agents/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to
 - Upgraded to Inference Engine 2.4.1 (#6269)
 - Fixed tensor indexing to use correct CHW layout (#6239)
 - Updated the installation doc (#6242)
+- Fixed Unity Editor crashing when quitting in play mode (#6267)
```
maryamziaa: Please change the PR number.

Codengineer: Done

maryamziaa (approved):

